Issue #11, opened 2019-11-07
Jupyter Notebook File Failing


BarisSari:
When the jupyter notebook file tried to run (**_/examples/notebooks/data_store_sqlite.ipynb_**) , it throws the following error:
```
OperationalError: (sqlite3.OperationalError) no such table: Datafiles
[SQL: SELECT "Datafiles".datafile_id AS "Datafiles_datafile_id", "Datafiles".simulated AS "Datafiles_simulated", "Datafiles".reference AS "Datafiles_reference", "Datafiles".url AS "Datafiles_url", "Datafiles".privacy_id AS "Datafiles_privacy_id", "Datafiles".datafile_type_id AS "Datafiles_datafile_type_id" 
FROM "Datafiles" 
WHERE "Datafiles".reference = ?
 LIMIT ? OFFSET ?]
[parameters: ('../repl_files/missing_platform.rep', 1, 0)]
```
![Screenshot from 2019-11-07 13-33-51](https://user-images.githubusercontent.com/41292272/68381854-97e55280-0163-11ea-8604-24e894241c39.png)

We should examine and fix it.

IanMayo:
Could the issue be because we still need to initialise the database?  That would explain why the `Datafiles` table isn't there.

BarisSari:
> Could the issue be because we still need to initialise the database? That would explain why the `Datafiles` table isn't there.

Yes, it was the first issue. But then, many other issues popped up.

PR #13, opened 2019-11-07
#11 Fix old method names and old attribute names

BarisSari:
Supports #11 

It fixes the old method names and old attribute names in **_pepys_import/core_**.


IanMayo:
@BarisSari - codecov is rejecting this, but at this point codecov isn't essential, it's something nice to steadily improve.

PR #20, opened 2019-11-08
Issue# 8: Welcome banner

tahirs95:


pepys_import/core/store/data_store.py
```diff
@@ -12,7 +12,8 @@
 from .db_base import base_postgres, base_sqlite
 from .db_status import TableTypes
 from pepys_import.core.formats import unit_registry
-
+from pepys_import import __version__
+from pepys_import.utils.branding_util import show_welcome_banner, show_software_meta_info
```
BarisSari: This might violate the black standard. Can you follow pre-commit instructions on README.md? So, it can automatically reformat the code

pepys_import/utils/branding_util.py
```diff
@@ -0,0 +1,16 @@
+from pyfiglet import Figlet
```
BarisSari: I think we don't have this library in the _requirements.txt_ . Can you add it?

tahirs95: Sure, I will add it.

tahirs95:
Welcome Banner Screen.

welcome_banner

PR #37, opened 2019-11-11
Feature/4 import strategy

IanMayo:
Experiment with parsing folders of data


IanMayo:
@BarisSari - I've uncommented the block where we push actual measurements to the data store.

(This class doesn't match our intended API, it's just to prove the data flow, and to create a SQLite datafile that's populated with real data)

Would you mind stepping back in to work on the SQLAlchemy code in this branch?  I'm getting "not bound to a session" errors for `test_sample_parser.py`

BarisSari:
> @BarisSari - I've uncommented the block where we push actual measurements to the data store.
> 
> (This class doesn't match our intended API, it's just to prove the data flow, and to create a SQLite datafile that's populated with real data)
> 
> Would you mind stepping back in to work on the SQLAlchemy code in this branch? I'm getting "not bound to a session" errors for `test_sample_parser.py`

Okay, I'm working on it :+1:

BarisSari:
I think you can merge the branch now. One wrong method (search_nationality instead of search_privacy) and cached Privacy & Sensor objects were causing the problem. I think it's not possible to use cached objects in our API for most of the cases.

pepys_import/file/replay_parser.py
```diff
@@ -169,15 +169,14 @@ def process(self, data_store, path, file_contents, data_file_id):
 
                 # and finally store it
                 with data_store.session_scope() as session:
-                    datafile = session.search_datafile_by_id(data_file_id)
-                    platform = session.add_to_platforms_from_rep(
+                    datafile = data_store.search_datafile_by_id(data_file_id)
```
IanMayo: thanks for these fixes @BarisSari 

For the earlier one, I hope that Unit Testing will prevent problems like those arising.

But, could you explain what I was doing wrong here?  Was I looking at the wrong object?

The problem wasn't a compiler error, it was a SQLAlchemy error.

BarisSari: As I said before, data_store.session_scope() yields objects itself(data_store). So, you needed to call data_store.search_XXX or data_store.add_XXX methods. However, there were other errors causing SQLAlchemy session error. For example, if you look at _add_to_platforms_from_rep_ it returns the cached object, however, it's not in the same session with the current and therefore it gives an error. Therefore, I commented out that lines

PR #43, opened 2019-11-12
Feature #3 - Introduce Spatial Objects

### LINKED ISSUE #3: Introduce Spatial objects
We need to introduce spatial objects to both the PostGIS and SQLite database implementations.

This will give us PostGIS and Spatialite. 

Once we've got good spatial referencing in, we can start work on spatial indexes in the database, to improve the performance of spatial search.

BarisSari:
Fixes #3


BarisSari:
I think it's more or less ready to review :) But before that, I'd like to say a few words about the PR:

- Location changed from String to Geometry(Point) type. I had trouble when I tried Geography Point on Spatialite. Therefore, I kept it as a Geometry type.
- **postgres_db.py** didn't have foreign keys and UUID was giving an error for the Entry table. It showed me that we needed to test populate methods for PostgreSQL/PostGIS as well. I configured the file and populated sample data without a problem at the end. But we may add tests for each populate method.
- Running two PostgreSQL/PostGIS tests caused an error of "Schema already exists!". Therefore, I added drop schema to tearDown methods of PostgreSQL/PostGIS tests.

tests/test_spatial_data.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+                .first()
+            )
+
+            # Check location point's type and WKTE value
+            self.assertFalse(isinstance(first_state.location, str))
+            self.assertTrue(isinstance(first_state.location, WKBElement))
+            self.assertEqual(
+                first_state.location.data,
+                "0101000020FFFFFFFF00000000000047400000000000004040",
```
IanMayo: @BarisSari - can you investigate how to view the contents of this object?  We _should_ verify it's (46,32)

BarisSari: Sure, I'll search for it tomorrow

tests/test_spatial_data.py
```diff
@@ -0,0 +1,111 @@
...
+        pass
+
+    def test_location(self):
+        with self.store.session_scope() as session:
+            # Filter state object by spatial location
+            first_state = (
+                self.store.session.query(self.store.db_classes.State)
+                .filter(
+                    self.store.db_classes.State.location.ST_Contains(
+                        WKBElement("POINT(46.000 32.000)")
```
IanMayo: @BarisSari - I changed the `46` above to `48`, and the test still passed. I think the logic may not work in the way we expect it to.  It's worth us learning about it.  

Could you introduce a test where we search for a point we know isn't present, and verify that `None` (or similar) is returned?

.travis.yml
```diff
@@ -9,11 +9,13 @@ addons:
...
 services:
   - postgresql
+before_install:
+  - sudo apt-get install -y sqlite3
+  - sudo apt install -y libgdal-dev libproj-dev libproj-dev
+  - sudo apt install -y libsqlite3-dev spatialite-bin libsqlite3-mod-spatialite
 install:
   - pip install -r requirements.txt
   - pip install codecov
-before_script:
```
IanMayo: @BarisSari - don't we need to install postgis any more?

pepys_import/utils/geoalchemy_utils.py
```diff
@@ -0,0 +1,6 @@
+EXTENSION_PATH = "/usr/lib/x86_64-linux-gnu/mod_spatialite.so"
+
```
IanMayo: @BarisSari - could you insert an O/S specific flag please?

On my MacOS the path is `"/usr/local/lib/mod_spatialite.dylib"`

tests/test_spatial_data.py
BarisSari: I added tests. The problem was filtering. There are two options in [Geo Alchemy tutorials](https://geoalchemy-2.readthedocs.io/en/latest/spatialite_tutorial.html). I played with them more and realized that I confused [WKBE and WKTE Elements](https://geoalchemy-2.readthedocs.io/en/latest/elements.html#geoalchemy2.elements.WKTElement) and therefore it didn't filter State objects. So, I changed tests according to that. It is not possible now to return more than one object (because one() used instead of first() after filtering). And The returned object is correctly filtered now. I added tests for non-existing points as well.

https://github.com/debrief/pepys-import/pull/43/commits/bc4e9b5534ca022f3b8130c9bb1f03aba1aa46be

.travis.yml
BarisSari: We are installing the package on line 9:  **postgresql-10-postgis-2.5** . However, we need to create extension for each new database. Therefore, I moved the code in initialise()

pepys_import/utils/geoalchemy_utils.py
BarisSari: Added.

https://github.com/debrief/pepys-import/pull/43/commits/15de49b6c54c568e99d0787e5ecba54da815d225

PR #46, opened 2020-02-03
30 update api

IanMayo:
This PR is to track progress on updating the Pepys API to match the design document.


IanMayo:
@BarisSari  - I don't think we should have `new` in the package name for the refactored code:
image

Also, if you were using `new` to separate it from the legacy version, I'm happy for you to delete the legacy version.

tests/test_unknown_engine.py
```diff
@@ -0,0 +1,50 @@
...
+    @unittest.expectedFailure
+    def test_unknown_engine_for_spatialite(self):
+        """Test whether DataStore raises exception when unknown database name entered"""
+        store = DataStore("", "", "", 0, ":memory:", db_type="sqlite")
+        BaseSpatiaLite.metadata.create_all(store.engine)
+        store.engine = None
+        # it must raise exception during initialise()
+        store.initialise()
+
+    @unittest.expectedFailure
```
IanMayo: Hmm, @BarisSari  - is it possible for this test to be more specific in how the failure is caught?  If PostGIS isn't running, I guess this test will fail - but that's not because we used an invalid database name.

BarisSari:
> @BarisSari - I don't think we should have `new` in the package name for the refactored code:
> image
> 
> Also, if you were using `new` to separate it from the legacy version, I'm happy for you to delete the legacy version.

I deleted **_new_** from the names:
https://github.com/debrief/pepys-import/pull/46/commits/8ae673dabac3dd0a87ea2e391b0886e4350cce55

tests/test_unknown_engine.py
BarisSari: Actually, this test is wrong right now. I'm sorry that I forgot to change it. I created it to hit OperationalError exception during the creation of tables. I'm going to change it or remove it. It doesn't test anything at all right now.

BarisSari:
Fixes #46

PR #67, opened 2020-02-17
[WIP] Added support for Windows incl installation instructions

robintw:
Work in progress to add support for Windows 

- Deal with Windows as a platform in geoalchemy_utils
 - Added installation instructions for Windows


IanMayo:
@robintw - I don't know if I've expressed this elsewhere or not, but I think there would be logic in installing the dependencies in a `libs` sub-folder. That will keep them easy to deploy.  We _could_ even create an initialise script that makes the relevant PATH changes, relevant to the deployed folder.

robintw:
I think this is ready to be merged now - I've tested from a bare VM and all of these steps lead to a working install ready for development.

README_Windows.md
```diff
@@ -0,0 +1,47 @@
...
+
+10. Go through the installation wizard making sure that you choose to load the 'Stack Builder' application when you exit the wizard.
+
+11. Use the StackBuilder application to install PostGIS, follow the wizard through to the end.
+
+12. Add the Postgres bin directory to the `%PATH%` - eg. `C:\Program Files\PostgreSQL\12\bin\` - but make sure it comes _after_ the `mod_spatialite` folder (hint: using [Rapid Environment Editor](https://www.rapidee.com/en/about) makes it easy to re-arrange entries in the PATH)
+
+12. Clone the [pepys-import](https://github.com/debrief/pepys-import) repository into a folder _inside_ the `pepys` folder
+
+13. Navigate to the `pepys` folder and run `python -m venv venv` to create a virtual environment in a folder called `venv`
```
IanMayo: @robintw - should this be the `pepys-import` folder? I guess it must be, since that's where the `requirements.txt` file is that you refer to a little further down.

robintw: Ah actually it should be the `pepys` folder and I should get the user to `cd` into the `pepys-import` folder later down - I found from experimentation that it helped keep things separate if the virtual environment folder was outside of the `pepys-import` code folder, particularly as we have to recreate it when we install the deployment anyway.
I'll do a small PR to fix this.

PR #69, opened 2020-02-17
Include path for spatialite DLL

### LINKED ISSUE #68: Spatialite DLL not registered for MS-Windows
When we run spatialite tests on MS-Windows, they fail with this error
```
  File "C:\git\pepys-import\pepys_import\utils\geoalchemy_utils.py", line 15, in load_spatialite
    connection.load_extension(EXTENSION_PATH)
TypeError: argument 1 must be str, not None
```

I now recognise that the error is because the `EXTENSION_PATH` isn't being provided, we only do it for Linux & Macos.

IanMayo:
Fixes #68 

Note: this fix works on the assumption that the spatialite DLL folder has been added to the MS Windows Path, as recommended in #67


robintw:
This is actually included in #67 - but we can include it here instead if you'd prefer.

IanMayo:
Sorry @robintw - I did check in the PR, too.  Must have missed it. Closing this one.

PR #73, opened 2020-02-17
49 update resolver

### LINKED ISSUE #31: Adopt CLI library
We currently have a hand-made command-line interaction module, in `command_line_input.py`

This has been acceptable so far, but now we require two refinements:
* the ability to have some responses permanently assigned to a key (such as `.` for Cancel)
* the ability to inject some key responses that would allow us to cover command-line interactions from unit tests

We should switch to the QPrompt library to support the above: https://qprompt.readthedocs.io/en/latest/

This library will also bring these benefits:
* fuzzy search for matching string. This will be good for platforms, sensors and datafiles
* change title of command prompt during process
* busy cursor for lengthy operations (such as the final batch commit of measurements)

### LINKED ISSUE #49: Command Line Resolver
Analysts will use the Command Line Resolver to interactively fill in missing fields.

The pepys-import command-line mockup included some "tricks" to save time, such as copying data from another platform.
 
I had a meeting with the clients today, but we agreed it wasn't necessary. A lot of the data (own platforms) will be pre-generated.  The import parsers will also be intelligently insert other data.

So, the command-line resolver will just work through the missing data.  But, where-ever possible we'll use the `fzf` search capability to make it quicker for the user to select existing values.

Aah, I think we do still need some logic for this, here are my first thoughts:
https://docs.google.com/spreadsheets/d/1vE1nP_ax89kqAjhYCD_1oM6u6vnzKr2v0Ie46MRJuDU/edit?usp=sharing

### LINKED ISSUE #52: NMEA Parser should get platform from user
The NMEA parser currently uses a hard-coded platform:
image

Instead, it should get the platform from the user.  We'll do that by using the resolver to get a platform name.  So, the resolver class(es) need to be able to get a platform name to use, not just to help in searching for one.

I've updated the resolver logic to support this: https://docs.google.com/spreadsheets/d/1vE1nP_ax89kqAjhYCD_1oM6u6vnzKr2v0Ie46MRJuDU/edit?usp=sharing

We should defer this issue until we have refactored the command-line-resolver class.

### LINKED ISSUE #71: Sensors are per-platform
In the sensors table, we will be storing the sensors for specific platforms.

So, there won't just be a single "GPS" sensor.  There will be a GPS sensor for each platform that is carrying a GPS sensor (or even multiple instances per platform).

That is the role of the `host` field in the Sensors table.

This is a current code snippet for getting a sensor.
```
                sensor = platform.get_sensor(
                    session=data_store.session,
                    all_sensors=all_sensors,
                    sensor_name="E-Trac",
                    sensor_type="GPS",
                    privacy="TEST",
                )
```
Note that `all_sensors` includes all sensors in the database.  Instead of this, within `platform.get_sensor` - the all_sensors should be the list of sensors for this platform.

BarisSari:
This branch aims to fix #49, fix #31, fix #52 and fix #71.


pepys_import/core/store/data_store.py
```diff
@@ -590,8 +589,43 @@ def search_table_type(self, table_type_id):
...
+            .first()
+        )
+        if datafile:
+            return datafile
+
+        synonym = (
+            self.session.query(self.db_classes.Synonym)
+            .filter(
+                self.db_classes.Synonym.synonym == datafile_name,
+                self.db_classes.Synonym.table == "Datafiles",
```
IanMayo: @BarisSari  - I don't know if this doesn't matter in Python, but this is an example of a _Magic Number_. (well, String)  Normal programming practice is to avoid this, and to only declare it as a constant, in one place.  Actually, we may already have it declared somewhere as a constant.  When it's a constant, we know we only have to change it in one place, for it to be updated everywhere.

pepys_import/core/store/data_store.py
```diff
@@ -590,8 +589,43 @@ def search_table_type(self, table_type_id):
...
+        """
+        datafile = (
+            self.session.query(self.db_classes.Datafile)
+            .filter(self.db_classes.Datafile.reference == datafile_name)
+            .first()
+        )
+        if datafile:
+            return datafile
+
+        synonym = (
```
IanMayo: Hmm, since we'll be looking for synonyms for multiple table types (esp Platforms & Sensors) - I wonder if it's worth refactoring this next block into a reusable function?

pepys_import/core/store/data_store.py
```diff
@@ -590,8 +589,43 @@ def search_table_type(self, table_type_id):
...
+                self.session.query(self.db_classes.Datafile)
+                .filter(self.db_classes.Datafile.datafile_id == synonym.entity)
+                .first()
+            )
+            if datafile:
+                return datafile
+
+        return None
+
+    def get_datafile(self, datafile_name=None, datafile_type=None):
```
IanMayo: @BarisSari - I think `get_datafile` could do with a method comment, particularly since we need to explain the difference between `get` and `find` methods for `_datafile`.

IanMayo:
Yes, another huge PR.  I've added a couple of comments, but want to have a play from the command line before I merge it.  It's late now, I'll be on it in the morning.

pepys_import/core/store/data_store.py
BarisSari: There is no `const` or any similar keywords in Python. However, you're right that we should avoid this. Therefore, I did the following changes:
- Created a new file under **_pepys_import/core/store_**, named **_constants.py_**
- Used **final** hint to tell developers that it shouldn't be reassigned even though it doesn't prevent change.
- Changed all strings passed to **add_to_synonyms** method with the values in **_constants.py_**. 
- Changed all table names in DB Classes with the values in **_constants.py_**.

https://github.com/debrief/pepys-import/pull/73/commits/b98898d73f947a7c5af5fefa0a21c7f3a44bc4fd

pepys_import/core/store/data_store.py
BarisSari: Yes, I think it is worth because it already duplicates codes. I created a new method named **synonym_search** for DataStore:

https://github.com/debrief/pepys-import/pull/73/commits/972c963728d1d0f967d751729ae83eee2dbd54d9

pepys_import/core/store/data_store.py
BarisSari: I wrote comments for it. If it is still not clear, I can add more:

https://github.com/debrief/pepys-import/pull/73/commits/552af3a5ec0e9fb2860626e6168ececac7c3e5bd

BarisSari:
> Yes, another huge PR. I've added a couple of comments, but want to have a play from the command line before I merge it. It's late now, I'll be on it in the morning.

Okay, I'm sure there will be more feedback once you play with it.

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -1,264 +1,701 @@
...
+        print("Ok, adding new platform.")
+
+        platform_name = prompt("Please enter a name: ", default=platform_name)
+        trigraph = prompt(
+            "Please enter trigraph (optional): ", default=platform_name[:3]
+        )
+        quadgraph = prompt(
+            "Please enter quadgraph (optional): ", default=platform_name[:4]
+        )
+        pennat_number = prompt("Please enter pennat number (optional): ", default="")
```
IanMayo: minor typo @BarisSari  - it's `pennant`, not `pennat`

https://en.wikipedia.org/wiki/Pennant_number

BarisSari: Corrected:

https://github.com/debrief/pepys-import/pull/73/commits/d4b20701eb383aaebfd6eab5d83cfa70c7259f06

IanMayo:
There's a lot of work in this PR :-)

I've found a hole in the logic.

1. Run test script provided in Slack (`Platform:AAAA`). 
2. Select `Search for existing platform`
3. Press "." (to cancel search)
4. App thinks we want to create platform titled ".".

When "." is pressed, we should return to the previous menu level (1,2,.)

IanMayo:
I think I've found another issue.  I've switched to a file-based database "test.db".  I went through the process to add `AAAA`, but it didn't show up in the test.db.

I first diagnosed this by running the test script twice - on the second time it still said it couldn't find the platform.   I also used DB Browser for SQLite to look inside the `.db.` file.

Aah, maybe the test script isn't pushing the datafile to the database.  I _think_ it should.

pepys_import/core/store/sqlite_db.py
```diff
@@ -78,25 +78,51 @@ class HostedBy(BaseSpatiaLite):
...
     table_type_id = 2
 
     sensor_id = Column(Integer, primary_key=True)
     name = Column(String(150), nullable=False)
     sensor_type_id = Column(Integer, nullable=False)
-    platform_id = Column(Integer, nullable=False)
+    host = Column(Integer, nullable=False)
     created_date = Column(DateTime, default=datetime.utcnow)
 
+    @classmethod
```
IanMayo: @BarisSari - `find_sensor` and it's equivalent in `postgres_db.py` file look the same. Shouldn't there be a single instance of the method?

pepys_import/resolvers/command_line_input.py
```diff
@@ -1,26 +1,39 @@
...
-        input_text = heading + "\n"
-        for idx, choice in enumerate(choices, 1):
-            if isinstance(choice, list):
-                choice_string = choice[0]
-                map_choice = True
-            else:
-                choice_string = choice
-            input_text += "   " + str(idx) + ") " + choice_string + "\n"
-        choice = input(input_text)
+from prompt_toolkit import prompt
```
IanMayo: Hmm, maybe we can save this for a future issue, but...

This class was an implementation of a command-line input.  Now though, it's a wrapper for `prompt_toolkit`.

Do we need to wrap `prompt_toolkit`?  I _Guess_ we can make more advanced usage of `prompt_toolkit `if we're accessing it directly from `command_line_resolver`.

IanMayo:
Ok, I've been through it @BarisSari - just a couple of questions.

pepys_import/core/store/sqlite_db.py
BarisSari: @IanMayo Since `get_sensor()` has to be a method of `Platform` class, I didn't want to put `find_sensor` to `data_store.py`. It could be confusing to have `find_sensor` method in data_store.py and `get_sensor` method in `postgres_db or sqlite_db`. Or I didn't want to implement it for only `sqlite or postgresdb` and call it from another because of the same reason. However, you're right that it's equivalent. If it doesn't seem right, we can find another solution.

pepys_import/resolvers/command_line_input.py
BarisSari: Yes, it is a wrapper now. I tried to create the same menus with small differences (cancelling texts, title, etc.) using `create_menu` function and I think it's useful now because it prevents many duplications. However, the name of the module is confusing. 

If you would like to, I can move the function from `command_line_input` to `command_line_resolver`. Or do you have something else on your mind?

pepys_import/resolvers/command_line_input.py
```diff
@@ -1,26 +1,39 @@
...
-            continue
 
-        if choice_value < 1 or choice_value > len(choices):
-            print(choice + " was not one of the options, please try again")
-        else:
-            if not map_choice:
-                return choice_value
-            else:
-                return choices[choice_value - 1][1]
+def is_valid(option):
```
IanMayo: I guess`is_valid` could move to the `command_line_resolver` class, since it's only called from there.

IanMayo:
@BarisSari - since merging other branches, you've now got some conflicts with `develop`.  In most cases it _looks like_ your changes simply have to overwrite those in `develop`

BarisSari:
> @BarisSari - since merging other branches, you've now got some conflicts with `develop`. In most cases it _looks like_ your changes simply have to overwrite those in `develop`

I'm resolving the conflicts now. They do not have to overwrite the changes in the develop branch. Because using `constants.py`,  which means variable names or assignments caused the conflicts.

PR #76, opened 2020-02-18
Update requirements

robintw:
There seemed to be some overlap between `requirements.txt` and `requirements_dev.txt` and the basic reqs file was bringing in some dependencies that are only needed for development.

This PR rationalises both files and splits things sensibly between them. This also significantly reduces the volume of files that need to be installed on the client's machines.

Notes: this depends on #67 - so that needs merging first.


robintw:
Oh I forgot to say, this also updates to the latest (master) version of `testing.postgresql` as that is required for Windows support.

requirements.txt
```diff
@@ -1,12 +1,7 @@
 SQLAlchemy==1.3.10
 geoalchemy2
 psycopg2
-nose2==0.9.1
-testing.postgresql
-black
+git+https://github.com/tk0miya/testing.postgresql.git#egg=testing.postgresql
```
IanMayo: @robintw - if this is for running the unit tests, I guess it should be in `requirements_dev.txt`, right?

robintw: Oh yes, good point! I'll fix that.

PR #89, opened 2020-02-20
Add deployment scripts

robintw:
- Added batch files to set PATH etc correctly on Windows so that scripts can be run with the included standalone Python distribution

Note: running these batch files requires a directory set up correctly with the various dependencies. I am working on a script to automatically download and set up these dependencies, but this is not finished yet (see outline in `\bin\_create_deployment.bat`.


bin/_create_deployment.bat
```diff
@@ -0,0 +1,7 @@
+REM Download and extract Python
+REM Download and extract libs
+REM  sqlite
+REM  mod_spatialite
+REM Copy new pth file
```
IanMayo: @robintw  - in the longer term we'll also have to distribute enough Postgis dependencies to  make a client connection to a networked PG database.  I _briefly_ thought  the *requirements.txt* would have enough, but it appears not.  I really hope we don't have to deploy the 100+Mb full PostGres component each time.

robintw: Hmm, I thought the dependencies installed from the requirements.txt should be enough - I thought it was just the psycopg2 library that was needed. It could be that psycopg2-binary is needed instead - I'll make a note to investigate next week

robintw: After a quick 5 sec look - the https://pypi.org/project/psycopg2-binary/ page says "You can also obtain a stand-alone package, not requiring a compiler or external libraries, by installing the psycopg2-binary package from PyPI" - so that should solve it. Unfortunately there's no easy way to install new packages into the deployed standalone Python installation (well there is, but it's messy) so I'll do a new release next week.

IanMayo:
@robintw - this should probably be obvious, but it wasn't to me.

I wanted to run the _import.py_ script from the project root.  I could use the machine-installed Python via `python -m pepys_import.import` - but that wouldn't have contained the dependencies. . I couldn't think how to  access the embedded python interpreter.

This will probably be a common question, so could you give an example of how to call the embedded python interpreter, please?

robintw:
@IanMayo - there should be a file called run_python.bat in the bin directory. That will run python with whatever arguments you give it - so you should be able to just replace python in your command with path/to/run_python.bat. Similarly, running without arguments will give you an interactive interpreter.

Would it be better if it was renamed to pepys_python.bat?

(On my phone at the moment and not officially working so won't be able to reply quickly)

robintw:
I think this is ready to be merged now @IanMayo. I'll do a separate PR shortly to start sorting out the README files so we end up with one user README and one dev README - with the User one explaining how to install a deployed zip file etc. (Because this is sorting out READMEs generally, I think it belongs in a separate PR).

I know none of us are very experienced with powershell, but I think I've commented the files well enough for most of us to understand them.

robintw:
Actually, please don't merge yet - I've found a couple of bugs that need fixing. I'll need Windows access to test them, so may not be able to fix until Wed evening or Thurs.

robintw:
This is now ready for review and merging - I've tested the new shortcut creation scripts and they work properly.

PR #90, opened 2020-02-21
Hotfix/incorporate pepys icon

IanMayo:
The MS-Windows shortcuts in the repo are hard-coded to a folder on my dev machine, so they can't actually be deployed.  I believe we need to create them once Pepys has been deployed.

Oh, and I've also introduced the Pepys icon file - that we can use for the shortcuts


IanMayo:
@robintw - in your Windows integration work, you may find the `.ico` file in this PR useful.  The MS-Windows shortcut editor only allows `.ico` files to be used to set the icon for a shortcut. But, if we're using Python to do it, and you need another format, please just ask me.

PR #98, opened 2020-02-25
Enhance importers

### LINKED ISSUE #74: File parsers: we need to defer "load" until the end
We currently have two file parsers, `nmea` and `replay`.  I'm currently working on `e-trac`.

The parsers follow this logic:
* loop through lines in file
    * collate data from one line, or multiple lines
    * check we have the data we need, then:
        * open a database connection
        * get the datafile
        * get the platform
        * get the sensor
        * create the new state object
        * configure the state fields
        * if the datafile validates:
            * submit the state

But, then we're opening/closing the database connection lots of times (potentially once per line).  We're also finding the datafile multiple times, even though we know we're just processing the single datafile.  And lastly, we're submitting those states one at at time.

I _think_ a better logic would be:
* open a database connection
* get the datafile
* loop through the lines:
    * collate the data
    * when we have our data:
        * get the platform
        * get the sensor
        * create the state (from the datafile)
        * configure the state fields
* check the datafile validates:
    * submit the datafile (well, all measurements in the datafile)

BarisSari:
This is an enhancement for importers. Fixes #74.

New logic for importers is as follows:
- Open a database connection for a file processor
- Create/Find datafile using get_datafile
- Loop through the lines:
      1. Process line
      2. Create/Find Platform using get_platform
      3. Create/Find Sensor using get_sensor
      4. Create State intermediary object
      5. Compute&Add new fields to the State object
- Once all lines are processed, validate datafile and commit all intermediary objects.


pepys_import/core/store/postgres_db.py
```diff
@@ -253,6 +253,10 @@ class Participant(BasePostGIS):
 
 
 class Datafile(BasePostGIS):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._measurement_files = []
```
IanMayo: @BarisSari - I don't think the entities in this array are `files` - I think they're just `measurements` (state, contact, comment).

BarisSari: Yes, you're right. Variable renamed:

https://github.com/debrief/pepys-import/pull/98/commits/814708e9e209aa01ddf534e0873fafd107867be5

IanMayo:
One item of feedback.  Several parsers may act on one file.  So, we may have one that just extracts comments - any piece of text beginning with a date.  But, another one may also be able to extract Lat/Longs from that text, to produce a series of positions (state).

Or, parsers may extract content from different parts of a larger composite datafile.

In your code, the datafile gets `commited` at the end of the parser.

If another parser then adds more state measurements, the list will grow, and we'll end up submitting the first set _and_ the second set.

I'm a fan of pulling functionality up into the `file_processor` class. So, I'd be quite happy to call `validate()` and `submit()` after the file_processor has run all of the parsers.

The analyst can then inspect errors/problems caused by all of the parsers for this file.

BarisSari:
> One item of feedback. Several parsers may act on one file. So, we may have one that just extracts comments - any piece of text beginning with a date. But, another one may also be able to extract Lat/Longs from that text, to produce a series of positions (state).
> 
> Or, parsers may extract content from different parts of a larger composite datafile.
> 
> In your code, the datafile gets `commited` at the end of the parser.
> 
> If another parser then adds more state measurements, the list will grow, and we'll end up submitting the first set _and_ the second set.
> 
> I'm a fan of pulling functionality up into the `file_processor` class. So, I'd be quite happy to call `validate()` and `submit()` after the file_processor has run all of the parsers.
> 
> The analyst can then inspect errors/problems caused by all of the parsers for this file.

Yes, I think it is a better approach to commit measurements after the run of all parsers. I moved the functionality to `file_processor`:

https://github.com/debrief/pepys-import/pull/98/commits/df2f8cc3d9077213a88de299f39328f091ed74b1

PR #104, opened 2020-02-26
Add parser for GPX files

### LINKED ISSUE #101: 2.2.1 GPX Importer
The clients regularly receive data in GPX format, including from:
* GPS Wristwatches
* GPS Trackers temporarily attached to smaller ships
* Some types of fixed-wing aircraft

I believe it's possible to represent tracks in GPX in a range of structure, so we'll have to start off by supporting our sample data-files.  I'll do a quick look for sample data on the Internet, and include some links in this issue.

robintw:
Adds parser for GPX files, including tests.

Fixes #101


robintw:
@IanMayo: a first version of the GPX parser is now working in this PR.

There are a few outstanding issues/questions:

1. There doesn't seem to be anywhere in the State table to record elevation, but GPX files sometimes record elevation. Do we want to change the database so we can store this?

2. Heading and course are different (something I learned today!), but is it correct that both should be in the 0 to 360 range? At the moment I'm using the `convert_heading` util function for checking the course values, which is a bit unclean - can we rename that to `convert_heading_or_course`?

3. How do we best go about testing the import/parsing? The REP file parser does this quite nicely with a intermediate object that parses a line and gives the relevant values - so everything can be checked. I don't think we can easily do that here with the XML parser, unless I refactor everything a lot. At the moment the tests for the GPX importer are based on the E-Trac importer, and just check that the relevant number of entries are created in the database, and not that the entries actually _contain_ the correct values.

4. Similarly to the above, how should we test that certain files produce errors? At the moment, errors are just printed out to the console - so there's no way of actually testing that the correct error was produced from a 'known bad' file.

IanMayo:
1. It's in the `Location` element.  We're currently only passing 2 fields to it (lat/long degrees) - we'll switch to passing 3rd field: elevation in metres.
2. Yes - though in many cases the analysts themselves don't consider the difference.  The function should be `convert_absolute_angle`.  It's a absolute angle because some measurements deliberately use `relative angle` - which is +/- 180 degs.
3. I guess you could include a block of XML in the source file, run it through the parsers, and check the relevant values were now present in the in-memory database.   Well, actually you could check examine the contents of entries in the `data_store.__measurements` array.
4. Yes, this logic still needs to be considered.  I'll have to sit down & think about it next week.

robintw:
This is now ready to be merged.

Once we've implemented the new ways of dealing with parsing errors on import, then we'll need to come back and change this - but we'll need to do that for all the parsers so that'll be a separate PR anyway.

pepys_import/file/gpx_importer.py
```diff
@@ -0,0 +1,134 @@
...
+                    print(
+                        f"Line {tpt.sourceline}. Error: <trkpt> element must have child <time> element"
+                    )
+                    continue
+
+                speed_str = self.get_child_text_if_exists(tpt, "{*}speed")
+                course_str = self.get_child_text_if_exists(tpt, "{*}course")
+                elevation_str = self.get_child_text_if_exists(tpt, "{*}ele")
+
+                print(
```
IanMayo: @robintw - we don't need these printed, do we?
(though obvs they'll be useful in my reviewing the PR :-) ).

IanMayo: I'll fix it.

robintw: Ah I've just fixed it - we got there at the same time!

robintw: (and yes, I should set up some sort of git hook to check for print statements when I push a final commit!)

PR #108, opened 2020-02-27
Feature/import tests

### LINKED ISSUE #93: 3.1.1 Identify initial set of import tests
Pepys will automated testing of imported data.

In the longer term these tests will deliver value in verifying the content of poor quality datafiles (typically human generated). But, in the 

To do this, we have to first identify a set of tests to apply. This issue will act as a temporary placeholder for these tests.

## None:
For very high volume data-sets that consistently contain zero defects, QA testing can be omitted. 

## Basic:
* lat >= -90 && <= +90
* long >= -180 && <= +180
* course/heading >=0 && < 360 (data stored in degs)

## Enhanced:
Note: while they are significantly more expensive (in processing time) these enhanced tests actually serve to verify the lat/long, in addition to the course/speed/heading.
* course/heading - loosely match delta between last two locations (+/- 90 degs)
* speed - loosely matches delta between last two locations (+/- speed * 10).

### LINKED ISSUE #94: 3.1.2 Implement initial import tests
We're developing a set of import tests in #93 

We need to both implement the tests, and devise an extensible framework for running the tests.

This implementation is detailed here: https://docs.google.com/document/d/1Wjf2XUOMXBsPavzaH9yHf50h4kLSSNXmi5AXxRD9Tr4/edit#heading=h.cfx6v5jgl5es

The strategic approach taken is also being considered here: https://docs.google.com/drawings/d/1JHkjhMw4YRgET8F4u3BSeo-tc4n8kPQ0YnuGEH2rTRM/edit

Ian would like the parsers to be as _dumb_ as possible, keeping as much functionality as possible in our tested, supported processing library (and similarly minimising the volume of code that the data scientists maintain).

So, for this reason Ian would like the test-suite to be run in `file_processor.py`.  An issue with this, is that different parsers ask for different levels of testing.  So, if multiple parsers handle a file, we may be running different levels of testing on the data that's sitting in "temporary storage" before being pushed to the database.

It _looks like_ the relevant place to do the testing is when a parser returns control to file_processor. Then file_processor could run the relevant level of testing against the records read in by that parser.

This would mean we have to do something like:
1.  capture `measurements.size()` before running a parser, then run the tests from that row to the end of measurements.
2. after running tests, submit those measurements to the database - which empties the `measurements` array
3. store the measurements for each parser into a hashmap indexed by the parser, then run all tests at the end, before submitting.

_Maybe_ the strategy should be led by what happens if there's a failure.  We can't submit just the records from one parser, since if the tests for a subsequent parser fail, the analyst will fix the file and resubmit it, and the records from the first parser would all be submitted to the database again.  Hmm, so I think we should run all of the parsers, and all of the tests before we fail - then we can give the analyst a full to-do list of what needs fixing.

BarisSari:
This branch aims to fix #93, and fix #94.

In order to examine the errors/extract logs, the following code snippet might be run:

```python
from pepys_import.file.file_processor import FileProcessor
from pepys_import.file.get_importers import get_importers

processor = FileProcessor()
importers = get_importers()
processor.register_importers(importers)
processor.process(DATA_PATH, None, True)
```
**Note:** Your DATA_PATH should be `YOUR_REPO_PATH/pepys_import/tests/sample_data`

**Note-2:** Right now, errors are saved as a dictionary. For example: 

```json
{"E-Trac Importer - Parsing error on e_trac_bad.txt": "Error on line 42. Date format '  2019/08' should be 10 figure data"}
``` 
I print them using `json.dump`. However, I think we can find a better way to save/export these errors. 

On the other hand, extract information has the following format: `44 measurement objects parsed by E-Trac Importer.` This also has room for improvement.

**Note-3:** If tests are run locally using `coverage3 run -m pytest tests`, it will create logs/errors for all importer tests. You can find them below `tests/sample_data/output`


pepys_import/core/formats/rep_line.py
```diff
@@ -83,14 +83,14 @@ def record(self, importer, record_type, measurement_object) -> None:
...
-    def parse(self):
+    def parse(self, error, error_type):
         tokens = self.tokens()
 
         if len(tokens) < 15:
-            print(
-                "Error on line {} not enough tokens: {}".format(
-                    self.line_num, self.line
-                )
+            error.append(
```
IanMayo: `error` should be `errors`, please.  It's `errors` in the calling class:
```
if not rep_line.parse(self.errors, self.message):
```

pepys_import/file/e_trac_importer.py
```diff
@@ -1,15 +1,22 @@
...
+
 from .importer import Importer
 from datetime import datetime
 from pepys_import.core.formats import unit_registry
 from pepys_import.utils.unit_utils import convert_absolute_angle, convert_speed
+from pepys_import.core.validators import constants
 
 
 class ETracImporter(Importer):
     name = "E-Trac Format Importer"
```
IanMayo: @BarisSari - we create 3 class level variables here. We always want them, for all importers/parsers.  Isn't it _tidier_ to have an `init()` method in `Importer`, and for e_trac_importer to call that in it's `init()` method?

Again, that will prevent a _lazy_ parser maintainer from not bothering to set the values.

IanMayo:
@BarisSari - I look forward to "driving" this PR.  Could you include in the initial description one or more commands to "drive" a bulk import, so I can look at how the errors are handled/presented, and see what success/failure of import looks like?

IanMayo:
Aah, @BarisSari - I see you've got failing unit tests. They're useful for something ;-)

pepys_import/core/formats/rep_line.py
BarisSari: Renamed:
https://github.com/debrief/pepys-import/pull/108/commits/363190f7324a0c27cb4c339fd717bdc115cb7819

pepys_import/file/e_trac_importer.py
BarisSari: Done. I kept default values for importers, yet when a new importer is implemented, it is going to be necessary to pass these 3 variables to `Importer` for `init()`:
https://github.com/debrief/pepys-import/pull/108/commits/5cd71a61ccd86c742cca0831da15fa12882bd951

BarisSari:
> @BarisSari - I look forward to "driving" this PR. Could you include in the initial description one or more commands to "drive" a bulk import, so I can look at how the errors are handled/presented, and see what success/failure of import looks like?

@IanMayo - I edited the description. You can find a code snippet and a few notes. According to your feedback, we can finalize this PR and merge it.

IanMayo:
@BarisSari - if you think this is done, you should remove the `WIP` in the title.

Note: for future "in progress" PRs, you can create a PR in `draft form`, which can't be merged until you say it's complete:
https://github.blog/2019-02-14-introducing-draft-pull-requests/

BarisSari:
> @BarisSari - if you think this is done, you should remove the `WIP` in the title.
> 
> Note: for future "in progress" PRs, you can create a PR in `draft form`, which can't be merged until you say it's complete:
> https://github.blog/2019-02-14-introducing-draft-pull-requests/

I removed `WIP`. Thanks for the link, I'll use `draft form` from now on. I didn't know it.

PR #110, opened 2020-02-28
Remove Entry and TableType tables, related methods, and tests

BarisSari:
This PR removes not used tables, _Entry and TableType_, from the project.


IanMayo:
@BarisSari - do you also need to remove `TableTypes.sql` ?

BarisSari:
> @BarisSari - do you also need to remove `TableTypes.sql` ?

Yes, I think we don't need TableTypes.sql . I'll remove it now

PR #112, opened 2020-03-01
Hotfix - Fix cleanup for test_importers

robintw:
A few of the tests in `test_importers.py` were leaving a few sqlite db files (`descending.db` and `single_path.db`) in the root directory of the project after running.
This caused problems for running the tests twice in a row, as they would give errors saying that the spatial reference systems table had already been created, as the file already existed.

These were possibly originally caused by pytest treating current directories slightly differently to unittest.

Fix just ensures that files are deleted from current directory, which is where the SQLAlchemy SQLite backend creates them.


tests/test_importers.py
```diff
@@ -18,11 +19,12 @@ def setUp(self) -> None:
...
 
     def tearDown(self) -> None:
-        single_level_file = os.path.join(FILE_PATH, "single_level.db")
+        single_level_file = os.path.join(CURRENT_DIR, "single_level.db")
         if os.path.exists(single_level_file):
             os.remove(single_level_file)
 
-        descending_file = os.path.join(FILE_PATH, "descending.db")
+        descending_file = os.path.join(CURRENT_DIR, "descending.db")
+        print(descending_file)
```
IanMayo: @robintw - we can probably delete that debug line.

robintw: Fixed now - thanks.

PR #113, opened 2020-03-01
Hotfix - Fix minor issues with setup.py file, specifically Python versions

robintw:
We aren't actually using the setup.py file much at the moment, but I thought it was best to fix a few issues before they cause us problems in the future.

Basically just fixed Python versions required, as we use f-strings in the code we require python >= 3.6.


IanMayo:
Great, thanks @robintw .

PR #114, opened 2020-03-01
WIP Switch to 3D points so we can store elevation/depth

robintw:
Currently a WIP.

All tests pass.

Questions:

1. What do we do when we don't have an elevation value? We can't give some points an elevation and not others: all now have to have an X, Y _and_ Z value. Currently I'm just setting it to zero when we don't know an elevation - do we want to carry on with this going forward? In situations where we do know, I'm using either the actual elevation value from the data (eg. GPX files), or `-1 * depth` if we have a depth value (eg. REP and E-Trac).

2. The Z value doesn't have an associated unit that is fixed in the database (unlike the X and Y values which are in a particular co-ordinate system). Do we want to store this as metres? If so, do we need to do some conversions (eg. is the depth value in the REP and E-Trac files stored as metres or feet?)


IanMayo:
Nice work Robin.

1. Missing Elevation. Hmm, that's a pain - thanks for researching it.  I just investigated if we could use NaN (not a number), but the Stack Overflow page I found said that PostGIS didn't support that.  Given this, I think the tidier solution is to have `Elevation` as an optional field in the `States` table (and in other tables that include `Location`).  In our schema, we can always have `Elevation (opt)` immediately after `Location` - to aid the user.

2. As detailed above, let's move it to a separate field. We're using SI as units within the database, so the `Elevation` will be in metres.

robintw:
Based on Ian's comments above, we're going to do a separate elevation field rather than using 3D points - so I will close this PR and open a new one to add a new field to the DB.

PR #116, opened 2020-03-02
Added elevation fields alongside all location fields in the Database

robintw:
Added elevation fields to database alongside all location fields.

This involved

 - Modifying the SQLAlchemy definitions of the tables in sqlite_db.py and postgres_db.py
 - Modifying the functions that read/write states in data_store.py
 - Modifying the data that is imported for testing purposes so it includes the new elevation field
 - Modifying the importers so that they store the elevation data that they parse


robintw:
This is ready to merge now @IanMayo.

PR #137, opened 2020-03-04
Integrate Extraction Highlighter

robintw:
Integrate the Extraction Highlighter code with the main pepys-import codebase. Highlighting now works properly, and all importers have been updated to record which tokens they're extraction, so a highlighted report can be produced.

This has also involved various other changes to the codebase, including changing some out-dated tests so that we're actually testing things in the way that we use them (particularly the NMEA and REP importers, which were being tested in a strange way).


robintw:
An example of a file highlighted by multiple parsers (REP file importer and REP comment importer) is attached (Github wouldn't let me attach a HTML file, so it's just an image)
Screen Shot 2020-03-09 at 10 49 06

robintw:
Right, this is now done and ready to be merged.

.gitignore
```diff
@@ -115,4 +115,7 @@ venv.bak/
 /lib
 /7zip
 *.zip
-*.lnk
\ No newline at end of file
+*.lnk
+
+# Log files produced by import code
```
IanMayo: Hmm, could this cause a problem for our trial data? the `NMEA` data comes with a `.log` suffix.

PR #139, opened 2020-03-04
Added path arguement in admin shell.

tahirs95:
Command to open the admin app, and to load the sample CSV files.
`python pepys_admin/admin_cli.py --path pepys_import/database/default_data`


IanMayo:
Aah, @tahirs95 , in other Pepys command line APIs we use a space between the label and the value.  So instead of ` --path=pepys_import/database/default_data` we would have ` --path pepys_import/database/default_data`

I don't know which is the _better_ way of doing it, but I'd like your case to follow the pattern we're using elsewhere, please.

tahirs95:
> Aah, @tahirs95 , in other Pepys command line APIs we use a space between the label and the value. So instead of ` --path=pepys_import/database/default_data` we would have ` --path pepys_import/database/default_data`
> 
> I don't know which is the _better_ way of doing it, but I'd like your case to follow the pattern we're using elsewhere, please.

@IanMayo You can use any of the ways. I don't need to change the code. Both cases will work.

PR #152, opened 2020-03-05
Apply Black and isort, solve Flake8 warnings

### LINKED ISSUE #144: Sort & Refactor Imports
During the development of the project, we are adding new imports when they are necessary. As far as I observed, we don't think about the order of imports or types of imports(absolute, relative, etc.). According to PEP-8, absolute imports are recommended if your package layout is not too complex:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path)

It also [mentions](https://www.python.org/dev/peps/pep-0008/#imports) the order of imports. We can use `isort` library to refactor imports. I was thinking to include it to pre-commit but it sometimes causes conflicts with `black`. Therefore, we can use it from time to time to have better imports.

This is a low priority task that aims to enhance our imports.

### LINKED ISSUE #145: Change Flake8 & Black Character Length
Maybe I couldn't find it, but I don't see the flake8 config in the repository anymore. If we are going to use the 100 character limit, we should declare it in a config file and adapt our black commands according to that.

This is a low priority task that aims to enhance code reading and review.

BarisSari:
Fixes #144.
Fixes #145.

Since `pyproject.toml` is added to the project, we don't need to pass any argument to `black`. It reads the line-length and Python version from the file.

### DIFF
--- a/pepys_admin/admin_cli.py
+++ b/pepys_admin/admin_cli.py
@@ -1,14 +1,16 @@
 # TODO: we have to keep these statements on top to load pepys_import.
 # We will see better approach to access modules inside the other module.
+import argparse  # noqa: E402
+import cmd  # noqa: E402
+import os  # noqa: E402
 import sys
+
+from iterfzf import iterfzf  # noqa: E402
+
+from pepys_import.core.store.data_store import DataStore  # noqa: E402
 
 sys.path.append(".")
 
-import argparse  # noqa: E402
-import cmd  # noqa: E402
-from iterfzf import iterfzf  # noqa: E402
-import os  # noqa: E402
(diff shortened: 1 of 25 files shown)


IanMayo:
@BarisSari - we've got two large PRs open.   I _Think_ we will help ourselves if we defer this until they're both merged.  

I _suggest_ we cancel this PR. Then as soon as those large PRs are merged, we create a new PR with this PR's `pyproject.toml` - and then let the other tidying happen.

BarisSari:
Okay, @IanMayo . I'm closing the PR for now.

Issue #123, opened 2020-03-02
2.2.4 Archive Input File


IanMayo:
Pepys-import users have several requirements for file processing:
1.  They can repeatedly call `import` on a folder, but without the folder contents getting loaded repeatedly
2. They require a non-editable `master` copy of input files, in case any audit of the source of some pepys data is required
3. They wish to be able to produce a permanent URL for the master input file, to allow future analysis tools to directly link to the source of some measured data

The above requirements are being met through moving input datafiles to a safe location (see archive location environmental variable in #119 ).  Python's `os.chown` can be used to make the file read-only for all users.

The archive target will be structured by date.  Each `import session` will put its results into a folder with the path generated like this:
`YYYY \ MM \ DD \ HH \ MM \ SS`

If imports are being processed very quickly, and a folder already exists with the current timestamp, then successive folders will be created with `_SSS` (millisecond) suffix.  So, in one folder we may have 4 subfolders:
```
32\
32_124\
32_674\
32_912\
```

PR #153, opened 2020-03-05
Introduce Replay comment parser

IanMayo:
Introduce new parser, for comments within `.rep` files, partially so @robintw has instance where multiple parsers run over the same file.

It's also a useful precedent for @BarisSari - since it introduces use of a new table type.

@BarisSari - you're aiming for high test coverage. But, this exposed a problem that I don't think we were previously aware of. I don't think we tried to `process` a single file that failed - so we needed to write to an output file. For single files, we _weren't_ generating the output folder, so the error reporting failed.

I'd appreciate you having a look at my modifications to `file_processor.py`.


pepys_import/file/file_processor.py
```diff
@@ -175,7 +182,7 @@ def process_file(self, file, current_path, data_store, processed_ctr):
 
             # If all tests pass for all parsers, commit datafile
             # get current time without milliseconds
-            timestamp = str(datetime.utcnow())[:-7]
+            timestamp = sanitize_filename(str(datetime.utcnow())[:-7])
```
BarisSari: Why do we need sanitize_filename?

pepys_import/file/replay_comment_importer.py
```diff
@@ -0,0 +1,105 @@
...
+        self.errors = list()
+        basename = os.path.basename(path)
+        print(f"Rep Comment parser working on '{basename}'")
+        error_type = self.short_name + f" - Parsing error on '{basename}'"
+        datafile.measurements[self.short_name] = list()
+        for line_number, line in enumerate(file_contents, 1):
+            if line.startswith(";"):
+                if line.startswith(";NARRATIVE:"):
+                    # ok for for it
+                    tokens = line.split(" ")
```
BarisSari: This might cause some problems if there are redundant whitespaces. For **split()** method, it is not the same with split() and split(" "). I remember that I had problems when I used split(" ") in REP Importer. Please check the following link: https://docs.python.org/3.7/library/stdtypes.html#str.split

> If sep is not specified or is None, a different splitting algorithm is applied: runs of consecutive whitespace are regarded as a single separator, and the result will contain no empty strings at the start or end if the string has leading or trailing whitespace. Consequently, splitting an empty string or a string consisting of just whitespace with a None separator returns [].

pepys_import/file/replay_comment_importer.py
```diff
@@ -0,0 +1,105 @@
...
+
+                    # separate token strings
+                    date_token = tokens[1]
+                    time_token = tokens[2]
+                    vessel_name_token = tokens[3]
+                    message_token = tokens[4]
+                    comment_type = "None"
+                elif line.startswith(";NARRATIVE2:"):
+                    # ok for for it
+                    tokens = line.split(" ")
```
BarisSari: This should change with split() as well.

pepys_import/file/file_processor.py
IanMayo: The file we generate contains characters that are illegal on MS-Windows:
```
  File "C:\Apps\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Apps\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\git\pepys-import\pepys_import\import.py", line 35, in <module>
    main(path=args.path, db=args.db)
  File "C:\git\pepys-import\pepys_import\import.py", line 22, in main
    processor.process(path, data_store, True)
  File "C:\git\pepys-import\pepys_import\file\file_processor.py", line 101, in process
    file, current_path, data_store, processed_ctr
  File "C:\git\pepys-import\pepys_import\file\file_processor.py", line 194, in process_file
    "w",
OSError: [Errno 22] Invalid argument: 'C:\\git\\pepys-import\\tests\\sample_data\\track_files\\output\\gpx_1_0_output_2020-03-06 08:57:11.log'
```

So, we need to either sanitize it, or change the underlying format of the timestamp string.

pepys_import/file/replay_comment_importer.py
IanMayo: Great - thanks. I've modified the sample data to include spaces.  It broke the test, and when I switched to `split()`, all worked fine :-)

pepys_import/file/replay_comment_importer.py
IanMayo: Yes, I missed that, just as you predicted 🤣

IanMayo:
> However, I didn't understand the problem of processing one file. 

@BarisSari  - we weren't generating the `self.output_path` for a single file:
https://github.com/debrief/pepys-import/pull/153/files#diff-8a372da3b1dd7e9eba3c99164d10756dR42

That member variable is relied upon for the output reporting.

pepys_import/file/file_processor.py
BarisSari: Okay, I understand now. I think we won't need to sanitize it after #123 is solved. Because we will use `OUTPUT_PATH \ YYYY \ MM \ DD \ HH \ MM \ SS` to put our logs and files. So, I think we won't need timestamp strings in the filenames.

pepys_import/file/replay_comment_importer.py
BarisSari: Okay, great 👍

PR #154, opened 2020-03-06
Feature/2.2.4 archive input files

### LINKED ISSUE #123: 2.2.4 Archive Input File

BarisSari:
Aims to fix #123


BarisSari:
@IanMayo, this PR is ready to review. However, before your review, can you try to parse the `sample_data` folder on MS-Windows? There shouldn't be any problems because I removed timestamps from the filenames.

pepys_import/file/file_processor.py
```diff
@@ -28,6 +38,38 @@ def process(
         :param descend_tree: Whether to recursively descend through the folder tree
         :type descend_tree: bool
         """
+        dir_path = os.path.dirname(path)
+        # create output folder if not exists
+        if not self.output_path:
+            self.output_path = os.path.join(dir_path, "output")
+            if not os.path.exists(self.output_path):
+                os.makedirs(self.output_path)
+        # create input_files folder if not exists
```
IanMayo: @BarisSari - can you explain the role of the `input_files` folder?

BarisSari: @IanMayo I thought that it would be easier to identify moved files if they are in ìnput_files`folder (_$OUTPUT_PATH/input_files_) . Otherwise, it might be difficult to find them if there are many folders created by importers. What do you think, is that make sense?

IanMayo: Sure.  Yes, I get your point about a cluttered folder.  Hmm, we have a choice: see the original files, and demote results/error logs to a child folder, or do it the other way around.  I can discuss this with the users, but we need a strategy for now.   Let's go with your strategy.

BarisSari: Okay, then. Let's keep it for now.

IanMayo:
@BarisSari  - ok, it doesn't fail on MS-Windows.

IanMayo:
@BarisSari - approved. Ok if I merge?

BarisSari:
> @BarisSari - approved. Ok if I merge?

Yes, that's okay 👍

Issue #119, opened 2020-03-02
Support for environment variables


IanMayo:
### Update
The _current_ plan is for just a single environment variable, that points to the Pepys config file.

The contents of this config file can be adjusted without needing to use sys-admin to amend the content of environmental variables.

The Environmental variable will be `PEPYS_CONFIG_FILE`.  It will point to the config file that is covered in issue #195 

----

@robintw has introduced the first environment variable.

~~I believe we have a requirement for others, too.  They should also be documented (in an .rst file so they get caught by `readthedocs`)~~

~~* `PEPYS_CONNECTION_STRING` = path to file in shared drive, containing database connection string. This can be amended to another file containing local file-based SQLite connection string.~~
~~* `PEPYS_ARCHIVE_LOCATION` = path to folder where _safe copies_ of input files, and extraction summaries are stored.~~
~~* `PEPYS_LOCAL_PARSERS` = path to folder containing local parsers, to be loaded dynamically by `file_processor.py`~~
~~* `PEPYS_LOCAL_TESTS_BASIC` = path to folder containing local QA test files to be run by parsers that need `BASIC` or `ENHANCED` testing~~
~~* `PEPYS_LOCAL_TESTS_ENHANCED` = path to folder containing local QA test files to be run by parsers that need `ENHANCED` testing~~
~~* `PEPYS_SPATIALITE_PATH`  = path to Spatialite folder, used to (optionally) override the hard-coded per-platform paths currently in use.~~

PR #172, opened 2020-03-09
Feature/local parsers tests

### LINKED ISSUE #122: 2.2.3 Local Parsers/Tests
There will be two types of parsers/QA-tests:
1. core ones, supplied with pepys-import
2. local ones, supplied/maintained by peyps-import users

The second category covers file formats that are private to the clients, or which require very frequent modification (due to the fragility of data-files that are entered by hand).

We will support local implementations through a series of environmental variables - the `PEPYS_LOCAL_xxx` vars defined in #119 .  At the relevant point in `file_processor.py` we will check for the presence of either environmental variable.  If the variable is present, then the files in that folder will be processed, and added to the core ones provided by pepys-import.

I've realised that the clients may wish to replace some of our parsers with their own.  This _could_ be supported by Pepys dynamically loading its own parsers, in addition to the local ones.  So, some _expert user_ could go into the Pepys folder and move a parser to their local folder, and apply the fix (or relaxation of some test).

BarisSari:
Fixes #122 

Note: It sticks with the previous version of issue #119. This means that it reads `PEPYS_LOCAL_PARSERS` environment variable and import modules in that folder.


pepys_import/file/file_processor.py
```diff
@@ -1,18 +1,36 @@
...
 from datetime import datetime
 from stat import S_IREAD
+from pathlib import Path
 
 from pepys_import.core.store.data_store import DataStore
 from pepys_import.core.store.table_summary import TableSummary, TableSummarySet
 from pepys_import.file.highlighter.highlighter import HighlightedFile
+from pepys_import.file.importer import Importer
+
+ROOT_PATH = Path(__file__).parent.parent.parent  # Main directory
```
IanMayo: ooh, this seems fragile.  Is there a way of doing it that doesn't rely on us being 3 folders deep?

BarisSari:
> Just the one change please @BarisSari , if it's possible that is. Do `setup.py` (or some other config file) set the home location of the project?
> 
> But, if we _have_ to rely on current implementation (`.parent.parent.parent) , I can live with it.

Okay, I'll search for it. I'll let you know when I find something and change the code.

pepys_import/file/file_processor.py
BarisSari: I created `path.py` at the root of the repository. So, we don't need to use Path() library to go to the file's parents 3 times:

https://github.com/debrief/pepys-import/pull/172/commits/5d0b3e88ac56a5691977ad2a4a3b99ae9de41c23

PR #177, opened 2020-03-09
Change how we integrate PostGIS

### LINKED ISSUE #176: Trouble extracting data from datastore
@tahirs95 is working on the management app, which includes the ability to export data.

But, when select queries are on the PG datastore which return `State` objects, this error is thrown:
```
  File "/Users/ian/git/pepys-import/venv/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 580, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedFunction) function st_asewkb(pepys.geometry) does not exist
LINE 1: ...s."States".sensor_id AS "pepys_States_sensor_id", ST_AsEWKB(...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
[SQL: SELECT pepys."States".state_id AS "pepys_States_state_id", pepys."States".time AS "pepys_States_time", pepys."States".sensor_id AS "pepys_States_sensor_id", ST_AsEWKB(pepys."States".location) AS "pepys_States_location", pepys."States".elevation AS "pepys_States_elevation", pepys."States".heading AS "pepys_States_heading", pepys."States".course AS "pepys_States_course", pepys."States".speed AS "pepys_States_speed", pepys."States".source_id AS "pepys_States_source_id", pepys."States".privacy_id AS "pepys_States_privacy_id", pepys."States".created_date AS "pepys_States_created_date" 
FROM pepys."States" 
```
 
I've done some research.  [This](https://stackoverflow.com/questions/37233116/point-type-in-sqlalchemy/37234645) solution says to provide the object type, but we already do that.

Or, [This](https://docs.ckan.org/projects/ckanext-spatial/en/latest/install.html) solution says to double-check that PostGIS is installed.  But, if we've already inserted this spatial data into the database, then the PostGIS extensions must be present.

I'm pretty sure this issue isn't related to the admin export - it does look like its in the config of SQLAlchemy or GeoAlchemy.  Here is a call that fails:
```
datastore = DataStore(
        db_username="postgres",
        db_password="postgres",
        db_host="localhost",
        db_port=5432,
        db_name="pepys",
        welcome_text="Pepys_Admin",
    )
    with datastore.session_scope():
        measurement_summary = datastore.get_status(report_measurement=True) 
```

IanMayo:
Fixes #176 

I've had an issue, where I can write data to a PostGIS database (in python), but I can't retrieve the data (in python).

With the support of @tahirs95 , and trial-and-error, we isolated the problem to the PostGIS configuration.  When I manually installed the PostGIS extensions (`CREATE EXTENSION postgis;`) it worked fine.

So, I looked at how we install PostGIS from Python, and I saw that were were inserting it into the Pepys schema (which sounded logical).  But, when I made the installation more general, it fixed the error.

I've tested this on Macos and Linux.  Aah, but I don't really know the detail of why it works.

Note: we don't have the issue for our `alwaysdata` remote hosting.  For that hosting, I had to use their tool to install PostGIS (since it required elevated privileges), so it didn't use/require our Python code to do it.

### DIFF
--- a/pepys_import/core/store/data_store.py
+++ b/pepys_import/core/store/data_store.py
@@ -129,7 +129,7 @@ class DataStore(object):
                 # Create schema pepys and extension for PostGIS first
                 query = """
                     CREATE SCHEMA IF NOT EXISTS pepys;
-                    CREATE EXTENSION IF NOT EXISTS postgis SCHEMA pepys;
+                    CREATE EXTENSION IF NOT EXISTS postgis;
                     SET search_path = pepys,public;
                 """
                 with self.engine.connect() as conn:
--- a/tests/test_data_store_initialise.py
+++ b/tests/test_data_store_initialise.py
@@ -58,16 +58,20 @@ class DataStoreInitialisePostGISTestCase(TestCase):
         table_names = inspector.get_table_names(schema="pepys")
         schema_names = inspector.get_schema_names()
 
-        # 36 tables and 1 table for spatial objects (spatial_ref_sys) must be created to default schema
-        self.assertEqual(len(table_names), 35)
+        # 34 tables must be created to default schema
+        self.assertEqual(len(table_names), 34)
         self.assertIn("Platforms", table_names)
         self.assertIn("States", table_names)
         self.assertIn("Datafiles", table_names)
         self.assertIn("Nationalities", table_names)
-        self.assertIn("spatial_ref_sys", table_names)
 
         # pepys must be created
         self.assertIn("pepys", schema_names)
+
+        # 1 table for spatial objects (spatial_ref_sys) must be created to default schema
+        table_names = inspector.get_table_names()
+        self.assertEqual(len(table_names), 1)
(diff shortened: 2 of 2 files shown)


robintw:
This makes sense, as my local install of Postgres already had PostGIS installed.

As for Windows - when installing for development, we set up PostGIS using the installation wizard that comes with the Postgres installer ('StackBuilder'), so this wouldn't apply.

IanMayo:
Note: this is breaking the test-suite. Investigating which is the `correct` implementation.

BarisSari:
> Note: this is breaking the test-suite. Investigating which is the `correct` implementation.

It breaks only one test which checks the number of tables in `pepys` schema. Since `spatial_ref_sys` is created in `public` schema, number of tables should decrement by 1:
https://github.com/debrief/pepys-import/blob/2b549b811ed53a620cf617b1d288bdb10d110028/tests/test_data_store_initialise.py#L62

Issue #170, opened 2020-03-09
Extract common parts of sqlite_db and postgres_db


robintw:
We currently have two separate files (`sqlite_db.py` and `postgres_db.py`) which define the table schema for the two backend databases that we support. They need to be separate because SQLite and Postgres don't quite support the same things (eg. data types).

However, we have a number of extra functions defined on the table classes (such as `create_state` or `submit`) and these should be extracted somehow to be common between the two files.

Not really sure how to go about this - do we need to something like multiple inheritance?

IanMayo:
I welcome input from @BarisSari on this.

BarisSari:
I think we are all aware of the issue of duplicated codes. After a quick search, it seems to me that we can use _Mixin Classes_ and solve this issue without any problems. Please check the following link:
https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/mixins.html#mixin-and-custom-base-classes

BarisSari:
Fixed in #180

PR #180, opened 2020-03-10
Extracted common parts of importer into base class

### LINKED ISSUE #168: Extract common parts of load_this_file method into superclass
All of the importers have a `load_this_file` method that starts with code like:

```
        self.errors = list()
        basename = os.path.basename(path)
        print(f"GPX parser working on {basename}")
        error_type = self.short_name + f" - Parsing error on {basename}"
        prev_location = None
        datafile.measurements[self.short_name] = list()
```

This could be extracted into the base class, and just called from there - to save work (and mistakes) for clients writing parsers.

_However_ this would mean that the `Importer` class was no longer an Abstract Base Class - are you happy for me to change this @IanMayo? Basically we'd actually be having a bit of functionality in the `Importer` class.

robintw:
This fixes #168, and should make it easier for parser writers as there is less 'boilerplate' they have to deal with.

Specifically:

 - All importers 'load_this_file' method started with the same few lines
 - These are now in the base class and are called with `super().load_this_file(data_store, path, file_object, datafile)`
 - Various variables now need 'self' in front of them: these have been changed
 - The name of the test file for parsing REP comments has now been updated to match the other tests for loading files


pepys_import/file/e_trac_importer.py
```diff
@@ -36,12 +34,7 @@ def can_load_this_file(self, file_contents):
         return True
 
     def load_this_file(self, data_store, path, file_object, datafile):
-        self.errors = list()
-        basename = os.path.basename(path)
-        print(f"E-trac parser working on {basename}")
-        error_type = self.short_name + f" - Parsing error on {basename}"
-        prev_location = dict()
-        datafile.measurements[self.short_name] = list()
+        super().load_this_file(data_store, path, file_object, datafile)
```
IanMayo: One alternate strategy here.

An issue is that the parser writer may _forget_ to call this super-method.

What if the parent class has a method, which then calls this method?

So, in `file_processor` we call something like `core_load_this_file()` - which performs those init steps, and then it calls `load_this_file()`.  That avoids the requirement for the parser writer to call the parent.

Yes - that naming could be improved ;-)

robintw: Good idea. Done. Just waiting for tests to pass and then it'll be ready to merge.

pepys_import/file/importer.py
```diff
@@ -52,6 +52,21 @@ def can_load_this_file(self, file_contents) -> bool:
         :rtype: bool
         """
 
+    def init_load_this_file(self, data_store, path, file_object, datafile):
+        """Initialises the loading of this data file
+
+        Performs the common operations that must be performed before the
+        load_this_file method is called. This *must* be called immediately
```
IanMayo: > This *must* be called immediately

We can avoid this issue by calling the "child" method ourselves, from the end of this "parent" one.

Let me try it out.

pepys_import/file/importer.py
```diff
@@ -45,19 +45,18 @@ def can_load_this_header(self, header) -> bool:
...
         """Whether this parser can handle this whole file
-        
+
         :param file_contents: Whole file contents
         :type file_contents: String
         :return: Yes/No
         :rtype: bool
         """
 
-    def init_load_this_file(self, data_store, path, file_object, datafile):
```
IanMayo: @robintw - I'm not fully happy with the name `core_load_file_file`, but I think this implementation simplifies the file_processor logic.

robintw: Yes, I couldn't work out what to call it either. Probably just stick with `core_load_this_file` until we can come up with a better name.

IanMayo: I _think_ I'm following the logic here: https://stackoverflow.com/a/34225912/92441

pepys_import/file/importer.py
```diff
@@ -52,7 +52,7 @@ def can_load_this_file(self, file_contents) -> bool:
         :rtype: bool
         """
 
-    def core_load_this_file(self, data_store, path, file_object, datafile):
+    def load_this_file(self, data_store, path, file_object, datafile):
```
IanMayo: @robintw - I've switched to this strategy - _slightly_ extending my Python skillset. 

But, I don't know if I'm correctly interpreting:
> The underscore prefix is meant as a hint to another programmer that a variable or method starting with a single underscore is intended for internal use. 

What do you think?

robintw: I like it - and wish I'd thought of it! Underscore methods in Python are used quite frequently to show that it's for internal use only - and although that's not enforced, it is well-accepted convention (and things like IPython won't show auto-completion for methods that start with an underscore). I think this is a far nicer way of doing it than my original way - thanks.

IanMayo: The, erm, tests were failing for my fix.  I think I've found the error, and am re-testing.

```
>       self._load_this_file(self, data_store, path, file_object, datafile)
E       TypeError: _load_this_file() takes 5 positional arguments but 6 were given
```
I _guess_ that another `self` was being prepended to the arguments in the call.

robintw: Yes you're right: when you call a method as `self.blah(a, b, c)` then `self` is automatically injected into the argument list and doesn't need to be specified explicitly.

BarisSari:
Fixed #170

PR #181, opened 2020-03-10
Uncomment the Enhanced Validator processing, now that bug fixed

robintw:
There was a bug in Enhanced Validator, so that was commented out in PR. That has now been fixed, so assuming all tests pass, this PR should be accepted uncomment the Enhanced Validator call.


robintw:
This isn't passing tests at the moment, which is to be expected as a different number of states are being created because a number of the REP files are failing validation and therefore not being imported.

The weird thing is that when running the test on my machine, I have 8 states created, but when it is run on Travis there are 16 states. I have no idea why this is, apart from the fact that 16 = 2 x 8...so maybe somehow Travis is doing it twice?!

tests/test_load_rep.py
```diff
@@ -42,7 +42,7 @@ def test_load_rep_data(self):
         with self.store.session_scope():
             # there must be states after the import
             states = self.store.session.query(self.store.db_classes.State).all()
-            self.assertEqual(len(states), 746)
+            self.assertEqual(len(states), 16)
```
IanMayo: this is a very large drop in the number of State measurements.  Are rejecting allow those rows?

BarisSari: When Enhanced Validator is used, only `rep_test1.rep` and `sen_ssk_freq.dsf` files are imported. And the latter doesn't have any State objects. That's why there is a very large drop and most of the rows(files) are rejected.

BarisSari:
> The weird thing is that when running the test on my machine, I have 8 states created, but when it is run on Travis there are 16 states. I have no idea why this is, apart from the fact that 16 = 2 x 8...so maybe somehow Travis is doing it twice?!

I also have this problem now.

tests/test_load_rep.py
IanMayo: I can't see the failures for `sen_tracks.rep` or `uk_track.rep`. 

Aah, it's a bug:
```
[
    {
        "REP Importer - Enhanced Validation Error": "Difference between Bearing (359.251) and Heading (22.500) is more than 90 degrees!"
    }
]
```

Those angles are only 23 degrees apart. We can merge this, I'll create an issue to cover the bug.

PR #184, opened 2020-03-10
Fixed REP comment importer to import full message not just first word

### LINKED ISSUE #178: REP Comment Parser only catching first word
In the REP format for comments, the content for the comment itself is all the tokens that appear after the track name (or narrative-entry type for `;NARRATIVE2:` entries.

But, our current token-based processing only store the first comment. That's a mistake - it should group them together to store a phrase.

_Hopefully_ the REP comment parser uses the ExtractionHighlighter, and this issue gives us a further opportunity to test-drive composite highlight tokens.

robintw:
Fixes #178 

Highlighted output now looks like the attached. Can you check this is right @IanMayo - as you're the expert on REP files. It looks right to me though.
Screen Shot 2020-03-10 at 10 02 03

### DIFF
--- a/pepys_import/file/replay_comment_importer.py
+++ b/pepys_import/file/replay_comment_importer.py
@@ -56,7 +56,7 @@ class ReplayCommentImporter(Importer):
                     date_token = tokens[1]
                     time_token = tokens[2]
                     vessel_name_token = tokens[3]
-                    message_token = tokens[4]
+                    message_tokens = tokens[4:]
                     comment_type = "None"
                 elif line.text.startswith(";NARRATIVE2:"):
                     # ok for for it
@@ -80,7 +80,7 @@ class ReplayCommentImporter(Importer):
                     comment_type_token.record(
                         self.name, "comment type", comment_type, "n/a"
                     )
-                    message_token = tokens[5]
+                    message_tokens = tokens[5:]
                 else:
                     continue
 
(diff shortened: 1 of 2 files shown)


pepys_import/file/replay_comment_importer.py
```diff
@@ -108,8 +108,10 @@ def load_this_file(self, data_store, path, file_object, datafile):
                     self.name, "timestamp", timestamp, "n/a"
                 )
 
-                message = message_token.text
-                message_token.record(self.name, "message", message, "n/a")
+                message = " ".join([t.text for t in message_tokens])
+                combine_tokens(*message_tokens).record(
```
IanMayo: Aah, this was the Python that threw me, the use of the `*` operator.

robintw: We could add a comment to explain it if you want? It basically turns a list into a set of arguments to a function. So here if there are five tokens in `message_tokens` then it passes those as five arguments to `combine_tokens`.

IanMayo:
I'll modify the test, to verify the full string has gone into the new record.

pepys_import/file/replay_comment_importer.py
IanMayo: No, it's ok @robintw .  We have to assume a basic level of Python skills.

PR #185, opened 2020-03-10
Send filename with suffix to get_datafile method

### LINKED ISSUE #157: Bug in storing Datafile reference
The reference in the DataFile is missing the file suffix.  We should have the whole filename in there.

BarisSari:
Fixes #157


IanMayo:
@BarisSari - can we cover it with a unit test, too?

We'll just need to modify an importer, so it retrieves the new datafile instance, and we check the filename is complete.

BarisSari:
> @BarisSari - can we cover it with a unit test, too?
> 
> We'll just need to modify an importer, so it retrieves the new datafile instance, and we check the filename is complete.

Sure, working on it.

PR #186, opened 2020-03-10
Restrict Importers

### LINKED ISSUE #162: Over-relaxed importer catching .DS_Store files
@robintw has indicated that Pepys is trying to process MacOS `.DS_Store` files.

We should identify the importer, and make the tests more strict.

BarisSari:
Fixes #162

### DIFF
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -216,6 +216,10 @@ class FileProcessor:
                 if not importer.can_load_this_file(file_contents):
                     good_importers.remove(importer)
 
+            # if good importers list is empty, return 0, which means the file is not processed
+            if not good_importers:
+                return 0
+
             # ok, let these importers handle the file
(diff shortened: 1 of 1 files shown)


pepys_import/file/file_processor.py
```diff
@@ -216,6 +216,10 @@ def process_file(self, file, current_path, data_store, processed_ctr):
                 if not importer.can_load_this_file(file_contents):
                     good_importers.remove(importer)
 
+            # if good importers list is empty, return 0, which means the file is not processed
```
IanMayo: Aah, @BarisSari - well spotted :-D

BarisSari: It looks like this was the only problem. Our importers are fine 🙂 NMEA and E-Trac Importers are listed as a good importer at first because they accept .txt files (NMEA also accepts .log). However, we read the first line of the files, which make them removed from the `good_importers`. However, we were not checking whether we still have a good importer or not.

PR #194, opened 2020-03-11
Switching methods on DB classes to Mixins to reduce duplication

robintw:
This moves all of the methods on database classes (from `sqlite_db.py` and `postgres_db.py`) into common mixin classes in `common_db.py` that are 'mixed-in' to the DB classes. This means we don't need to duplicate the methods for both the SQLite and Postgres DB classes.

As well as moving the code, this has involved a few small changes to the API - mostly to pass `DataStore` instances rather than `Session` instances in some places, so that we can access the DB classes as `DataStore.db_classes`.


robintw:
This has passed all tests locally. Assuming it passes on Travis then it's ready to be merged.

@IanMayo If you could get this merged before tomorrow then I can start working on the new getters/setters tomorrow, as they will depend on this.

Also, @BarisSari's PR #196 will conflict with this, so I think we should merge my PR first and then the conflicts will be relatively easy to deal with.

PR #196, opened 2020-03-11
Better Validation Errors

### LINKED ISSUE #166: Add some way for the user to find where in the input file the validator found a problem
At the moment the validators report problems, but don't say _where_ the problem was found. I'm aware that we probably don't have the line number for the location where the error was found (because we're in the validator rather than the parser, and so don't know anything about line numbers) - but could we include the timestamp, so the user can look through the file and find the right location by looking for the right timestamp value?

BarisSari:
Fixes #166

### DIFF
--- a/pepys_import/core/store/common_db.py
+++ b/pepys_import/core/store/common_db.py
@@ -107,30 +107,43 @@ class PlatformMixin:
 
 
 class DatafileMixin:
-    def create_state(self, data_store, sensor, timestamp, parser_name):
+    def create_state(self, data_store, platform, sensor, timestamp, parser_name):
         state = data_store.db_classes.State(
             sensor_id=sensor.sensor_id, time=timestamp, source_id=self.datafile_id
         )
+        state.platform_name = platform.name
+        state.sensor_name = sensor.name
         self.measurements[parser_name].append(state)
         return state
 
-    def create_contact(self, data_store, sensor, timestamp, parser_name):
+    def create_contact(self, data_store, platform, sensor, timestamp, parser_name):
         contact = data_store.db_classes.Contact(
             sensor_id=sensor.sensor_id, time=timestamp, source_id=self.datafile_id
         )
+        contact.platform_name = platform.name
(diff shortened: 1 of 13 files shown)


pepys_import/file/e_trac_importer.py
```diff
@@ -134,6 +134,9 @@ def _load_this_file(self, data_store, path, file_object, datafile):
             state.speed = speed
             speed_token.record(self.name, "speed", speed, "knots")
 
+            state.platform_name = vessel_name
```
IanMayo: I'm nervous about  the way we're storing this cached data in the state, since it relies on the parser author remembering to use the lines.  Can't we retrieve/store these values inside the `datafile.createState()` method?  Then it's under our control, and we know it will be done every time.

IanMayo: We're also relying on the parser author storing the _correct_ values in the state.  Again, if it happens in our code we can rely on it.

BarisSari: Yes, we can do that. I'm changing the code.

pepys_import/core/store/common_db.py
```diff
@@ -107,30 +107,43 @@ def get_sensor(self, data_store, sensor_name=None, sensor_type=None, privacy=Non
...
+        contact.sensor_name = sensor.name
         self.measurements[parser_name].append(contact)
         return contact
 
     def create_comment(
-        self, data_store, platform_id, timestamp, comment, comment_type, parser_name
+        self,
+        data_store,
+        platform,
+        sensor,
```
IanMayo: @BarisSari - Note: we don't store `sensor` for comments.  I guess you could use `None` or "N/A" instead.  I expect subsequent reporting logic relies on it being present.

BarisSari: Sorry that I missed out. I removed the argument and replaced it with "N/A":

https://github.com/debrief/pepys-import/pull/196/commits/51b5afd15aa81352a632f60e3300a47fcc6a7655

PR #206, opened 2020-03-12
Convert database attributes on State and Media to properties to enforce units

robintw:
Converts the attributes listed below into properties that enforce the use of units. See this wiki page for more info: https://github.com/debrief/pepys-import/wiki/SQLAlchemy-attributes-and-units.

The changes for the location property are a bit more extensive, and require the (eg) `State.location` to be set to a `Location` class instance. That class has methods that can set latitude and longitude from either decimal degrees or degrees, minutes and seconds - and they are designed so you have to be explicit. For example, you deliberately can't do something like `Location(50, 1)` as it is unclear which one is latitude and which is longitude - instead you have to use a descriptively named method like `set_latitude_decimal_degrees`. All places where the Location is set or retrieved have been updated to use this new Location class.

**Note:** This PR was originally going to convert *all* relevant database attributes to properties with units, but for ease of merging quickly (given current work challenges due to the coronavirus), this has been shrunk to just apply to State and Media to start with. A new issue has been opened for completing the work on the Contact and Activation classes (see #218).

List of attributes that have been updated:

**State:**
- [x] speed (length / time)
- [x] location
- [x] elevation (length)
- [x] heading (angle)
- [x] course (angle)

**Media:**
 - [x] location
 - [x] elevation (length)


IanMayo:
Thanks @robintw  - will be on this first thing tomorrow

pepys_import/utils/unit_utils.py
```diff
@@ -151,21 +116,43 @@ def distance_between_two_points_haversine(first_location, second_location):
...
+        valid_distance = float(distance)
+    except ValueError:
+        errors.append(
+            {
+                error_type: f"Line {line_number}. Error in distance value {distance}. "
+                f"Couldn't convert to a number"
+            }
+        )
+        return False
+    distance = (valid_distance * units).to(unit_registry.meter).magnitude
```
IanMayo: Hmm, I'd rather this returned a distance with units.  I _don't think_ we need to convert it to metres.  It can move around the codebase as distance with units (`yards`), and get passed into the data_store as that.  Inside the data_store, it can be converted to metres before going into SQL `distance.to(unit_registry.metre).magnitude`.

So, the parsers convert the value read in (distance without units) into a distance with units as soon as possible, and it's that distance with units that gets passed around.

IanMayo: Aah, hold on. The above method is only used once, and then it's result is (currently) ignored anyway.  I see we do follow the intended behaviour for `state`.

PR #212, opened 2020-03-16
Hotfixes - Fix tests for moving files, fix directory in error message for local parsers

robintw:
A few minor fixes:

1. The REP Contact parser tests moved the input files to the archive location, whereas most of the other tests have patched the move function so they don't actually do the moving. I've added this for the REP Contact parser, so files don't get moved every time the test is run.

2. It was impossible to run individual files inside `config_file_tests` because of path issues, so I added an `__init__.py` file to that folder (like all other test folders) to make it work properly with pytest.

3. There was an error in FileProcessor where it tried to use an invalid variable name in an error message - I've fixed this to point to the LOCAL_PARSERS variable.

### DIFF
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -24,7 +24,7 @@ class FileProcessor:
         if LOCAL_PARSERS:
             if not os.path.exists(LOCAL_PARSERS):
                 print(
-                    f"No such file or directory: {local_importers_path}. Only core "
+                    f"No such file or directory: {LOCAL_PARSERS}. Only core "
                     "parsers are going to work."
                 )
             else:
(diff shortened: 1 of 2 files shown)


robintw:
(Sorry, please ignore the erroneous commit and revert - I accidentally committed to the wrong branch. Someone with more git expertise than me might be able to fold those two commits into a 'nothing' commit, but I'd rather not screw it up any more while I'm tired)

BarisSari:
Hi @robintw , I also realized the first and the third issue during working on my PR. I added fixes in my branch but I think it's better to have a hotfix for them. However, I wasn't aware of the second issue. I think it's because of my testing approach. I'm running tests on my IDE and it works for single files in `config_file_tests` folder as well. Sorry that it caused a path issue. 

Note: I'll squash your last erroneous commits.

Issue #55, opened 2020-02-14
REP export capability


IanMayo:
In release 1 we aim to have the ability to export data from the database to REP format.

Release 1 will just export entries in the `States`, `Contacts` and `Comments` tables.  

We will have a Python class/method with this API:
```
export(states, contacts, comments)
```

The method will return the contents of those element lists as a string.  Note: many fields in the classes won't have a representation in REP format, so are ignored.

Note: when looping through the contents of the tables, the platform name to use will be retrieved via the `Sensor` entry in each row.  

It's _probably_ worth cacheing the platform names that are retrieved from sensor ids, since the same platform is likely to appear many times.

The following types of element will be created:
* States - these will be exported as Track Data items: https://debrief.github.io/tutorial/reference.html#replay_track_format
* Comments - these will be exported as `;Narrative:` entries: https://debrief.github.io/tutorial/reference.html#replay_core_annotations
* Contacts - these will be exported as `;Sensor:` elements: https://debrief.github.io/tutorial/reference.html#replay_core_annotations

IanMayo:
The export process is failing.

I populated a PG database using this command: `python\python.exe -m pepys_import.import --path tests\sample_data\track_files`

I then tried to export some data, using `python\python pepys_admin\admin_cli.py`

But, I got this error: 
```
Exported Datafile is: uk_track.
Traceback (most recent call last):
  File "D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\cmd.py", line 214, in onecmd
AttributeError: 'AdminShell' object has no attribute 'do_1'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\git\pepys_build\pepys-import\20200304_pepys-import\python\Lib\site-packages\sqlalchemy\engine\base.py", line 1249, in _execute_context
    cursor, statement, parameters, context
  File "C:\git\pepys_build\pepys-import\20200304_pepys-import\python\Lib\site-packages\sqlalchemy\engine\default.py", line 580, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.UndefinedFunction: function st_asewkb(pepys.geometry) does not exist
LINE 1: ...s."States".sensor_id AS "pepys_States_sensor_id", ST_AsEWKB(...
                                                             ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #213, opened 2020-03-16
Rep export capability #55

### LINKED ISSUE #210: Are we exporting depth/elevation incorrectly in REP export?
When we export REP data in https://github.com/debrief/pepys-import/blob/c45fd82c220d65cbb6a373f8adbff3d8e1b2f775/pepys_import/core/store/data_store.py#L1164 we export the absolute value of the elevation.

According to https://debrief.github.io/tutorial/reference.html#replay_track_format the Depth field in REP files is 'Depth in metres'.

Given that we store elevation, do we actually need to do `-1 * elevation` for the export?

tahirs95:
Exporting depth/elevation incorrectly in REP export is fixed.

Fixes #210


IanMayo:
@tahirs95 - you also need to include exporting Contact data in here. It's currently commented out in `data_store.py`.

You can find some contact data by trying to export `.dsf` files from the admin app.

PR #215, opened 2020-03-17
Hotfix/Expire location attribute to fetch it from the DB

BarisSari:
As @robintw mentioned on Slack, the `location` attribute was returning `String` or `WKBElement` in different test cases. It was happening because of the laziness of SQL Alchemy. State objects were inserted to the DB, however, they weren't fetched from it if you are using the same session. So, the `location` attribute was still a string. Therefore, `session.expire` added after the `session.flush`. Now, when a State object is called again in the same session, the location attribute is fetched from the DB. So, it is a `WKBElement` object.

### DIFF
--- a/pepys_import/core/store/common_db.py
+++ b/pepys_import/core/store/common_db.py
@@ -206,6 +206,7 @@ class StateMixin:
         """Submit intermediate object to the DB"""
         session.add(self)
         session.flush()
+        session.expire(self, ["location"])
 
         return self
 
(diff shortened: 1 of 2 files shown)


IanMayo:
So, @BarisSari - `location` does look like the only field in `States` that gets transformed in this way.

But, I can see we also have `location` in `Contacts` - so I guess you should make the change there, too.

robintw:
We also have location in `Media` too.

BarisSari:
Yes, that's why I added it to the `ContactMixin` class as well. You can see it in my commit. However, we don't have any methods related to `Contacts` or `Media` in `data_store.py`. When we add them, we should keep in mind to use `session.expire`.

PR #216, opened 2020-03-18
Add ability to turn off moving of input files

### LINKED ISSUE #165: Archive strategy for unit tests
Our strategy for the target environment is that as files are successfully imported, they are archived to a secure drive.

This will cause trouble for our unit tests, since we'll be _losing_ our input files.

In the target environment, an environment variables will be used to provide the location of archive folder.  

I guess we have two issues:
* output logs should go to a predictable location, that we can `.gitignore`
* we should have a Python flag to indicate that input files shouldn't be archived.  This should be set for most (but not all) unit tests.

### LINKED ISSUE #205: Add ability to turn off moving of input files
I think we discussed this on Slack, but I couldn't see any issue for it. We thought it'd be helpful to be able to turn off the moving of input files for testing purposes (both for us testing the CLI and for the parser writers when testing new parsers, so they don't have to keep moving files back all the time).

This could be an appropriate setting to put into the config file @BarisSari is developing in #202.

BarisSari:
Fixes #205 
Fixes #165

### DIFF
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -40,8 +40,9 @@ class FileProcessor:
         self.output_path = None
         self.input_files_path = None
         self.directory_path = None
-        # Check if archive location environment variable exists
+        # Check if ARCHIVE_PATH is given in the config file
         if ARCHIVE_PATH:
+            # Create the path if it doesn't exist
             if not os.path.exists(ARCHIVE_PATH):
                 os.makedirs(ARCHIVE_PATH)
             self.output_path = ARCHIVE_PATH
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -21,7 +21,7 @@ USER = getuser()
 
 
 class FileProcessor:
-    def __init__(self, filename=None, archive=True):
+    def __init__(self, filename=None, archive=False):
         self.importers = []
         # Register local importers if any exists
         if LOCAL_PARSERS:
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -21,7 +21,7 @@ USER = getuser()
 
 
 class FileProcessor:
-    def __init__(self, filename=None):
+    def __init__(self, filename=None, archive=True):
         self.importers = []
         # Register local importers if any exists
         if LOCAL_PARSERS:
@@ -45,6 +45,7 @@ class FileProcessor:
             if not os.path.exists(ARCHIVE_PATH):
                 os.makedirs(ARCHIVE_PATH)
             self.output_path = ARCHIVE_PATH
+        self.archive = archive
 
     def process(
         self, path: str, data_store: DataStore = None, descend_tree: bool = True
@@ -294,12 +295,12 @@ class FileProcessor:
                     os.path.join(self.directory_path, f"{filename}_output.log"), "w",
                 ) as f:
                     f.write("\n".join(log))
-                # move original file to output folder
-                new_path = os.path.join(self.input_files_path, basename)
-                shutil.move(full_path, new_path)
-                # make it read-only
-                os.chmod(new_path, S_IREAD)
-
+                if self.archive is True:
+                    # move original file to output folder
+                    new_path = os.path.join(self.input_files_path, basename)
+                    shutil.move(full_path, new_path)
+                    # make it read-only
+                    os.chmod(new_path, S_IREAD)
             else:
                 # write error log to the output folder
                 with open(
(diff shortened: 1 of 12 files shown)


pepys_import/import.py
```diff
@@ -25,9 +25,17 @@ def main(path=DIRECTORY_PATH, db=DEFAULT_DATABASE):
     parser = argparse.ArgumentParser()
     path_help = "The path to import data from (The default value is the directory of the script)"
     db_help = "The database to connect to: (The is to use an in-memory database)"
+    archive_help = "The flag to move files after the import (The default value is True)"
```
IanMayo: this guidance doesn't quite match the implementation.  Suggest:
`Instruction to archive (move) imported files to designated archive folder`

pepys_import/import.py
```diff
@@ -25,9 +25,17 @@ def main(path=DIRECTORY_PATH, db=DEFAULT_DATABASE):
...
     parser.add_argument(
         "--path", help=path_help, required=False, default=DIRECTORY_PATH
     )
     parser.add_argument("--db", help=db_help, required=False, default=DEFAULT_DATABASE)
+    parser.add_argument(
+        "--archive", dest="archive", help=archive_help, action="store_true"
+    )
+    parser.add_argument(
```
IanMayo: We can just have `-archive`. When present we archive, when it isn't, we don't.

pepys_import/file/file_processor.py
```diff
@@ -45,6 +45,7 @@ def __init__(self, filename=None):
             if not os.path.exists(ARCHIVE_PATH):
```
IanMayo: Now that we're using the config file, we can remove support for the ARCHIVE_PATH environment variable, please (including tests).  This will simplify the maintenance/configuration of Pepys.

pepys_import/import.py
BarisSari: Yes, you're right. I wrote it for `--archive True/False`. I'm changing it.

IanMayo:
@BarisSari - the code coverage suggests we're not checking that the `archive` happens:
https://codecov.io/gh/debrief/pepys-import/pull/216/src/pepys_import/file/file_processor.py#L300

pepys_import/import.py
BarisSari: Help text changed:
https://github.com/debrief/pepys-import/pull/216/commits/a89b6de6c3fc129078b8acbcd2610cf87becdda8

pepys_import/import.py
BarisSari: Done:
https://github.com/debrief/pepys-import/pull/216/commits/a89b6de6c3fc129078b8acbcd2610cf87becdda8

pepys_import/file/file_processor.py
BarisSari: Actually, this is a variable that comes from the config file (`config.py`). So, it is not an environment variable.

BarisSari: I think you were confused because of the comment. I'm going to change it now

BarisSari: Done:
https://github.com/debrief/pepys-import/pull/216/commits/246be9b009630fba3bccc358177d3c7480be4010

BarisSari:
> @BarisSari - the code coverage suggests we're not checking that the `archive` happens:
> https://codecov.io/gh/debrief/pepys-import/pull/216/src/pepys_import/file/file_processor.py#L300

A new test added for testing archiving files:
https://github.com/debrief/pepys-import/pull/216/commits/7cd9cbecc1ccd68e5b442e14af551f7a40788d17

IanMayo:
@BarisSari - I think the test also needs to verify that they have been removed from the original location, since I think that "Copy" would currently result in a passing test.

IanMayo:
Aah, I think we also need to test that the moved files are read-only.

BarisSari:
> @BarisSari - I think the test also needs to verify that they have been removed from the original location, since I think that "Copy" would currently result in a passing test.

> Aah, I think we also need to test that the moved files are read-only.

Okay, I'm updating the test

BarisSari:
@IanMayo , I made the changes you requested:
https://github.com/debrief/pepys-import/pull/216/commits/1e10261627244925f09de9f7517c9737919230c1

tests/test_importers.py
```diff
@@ -136,6 +140,9 @@ def test_archiving_files(self):
         assert "sen_tracks.rep" in names
         assert "uk_track.rep" in names
 
+        # Assert that there is no file in the input_files folder anymore
```
IanMayo: Hmm, I think this is checking that the target folder has been cleaned up.

I think we still need to test that the input files get moved on successful import. We're testing that they've arrived at the new location, we just aren't testing that they're no longer present in the start/original location.

BarisSari: Done:
https://github.com/debrief/pepys-import/pull/216/commits/28b30864dba0073933e64bea33110b34b0a2ff6b

PR #217, opened 2020-03-18
Identify Duplicated Files

### LINKED ISSUE #191: 2.6.1 Identifying duplicate files
There is a strong chance that the users will receive the same file from multiple senders, or in multiple media (CD & e-mail, for example).

Pepys should be able to identify the fact that a file is a duplicate of one that is already loaded in Pepys.  This can be done by cacheing some values in the `datafile` table:
* file-size: if this file matches the size of any existing one, go onto next test
* file-hash: if this file matches the hash of the above same-size one, it's a duplicate.

Note: if the files are huge (currently don't know what size represents huge), then we probably don't need to hash the whole file, _maybe_ just the first megabyte.  If we has the first megabyte, we _should_ be able to get past any introductory metadata.

Article on Python hashing here: https://www.pythoncentral.io/hashing-files-with-python/

BarisSari:
Fixes #191


tests/test_importers.py
```diff
@@ -157,6 +158,31 @@ def test_archiving_files(self):
         # Assert that there is no file in the input_files folder anymore
         assert len(os.listdir(moved_files_path)) == 0
```
IanMayo: @BarisSari - IMHO this needs a bit more testing. We could do with:
* create a copy of the file, with a different name. Shouldn't load.
* create a copy of the file, with a different name, change size (delete character). Should load.
* create a copy of the file, with different name, same size, different hash (change character). Should load

BarisSari: Okay, Ian. I'm implementing new tests

BarisSari: New tests created:
https://github.com/debrief/pepys-import/pull/217/commits/2a36f69979c71ed6ca4ba346637fd9994f9cfd64

PR #220, opened 2020-03-20
Added export all functionality for admin.

tahirs95:


pepys_admin/admin_cli.py
```diff
@@ -25,7 +26,7 @@ def create_postgres_data_store():
         db_password="postgres",
         db_host="localhost",
         db_port=5432,
```
IanMayo: could we keep the previous database name?

pepys_admin/admin_cli.py
```diff
@@ -148,6 +150,40 @@ def do_export(self, arg):
...
+                    "Please provide folder name (Press Enter for auto generated folder):"
+                )
+                if folder_name:
+                    if os.path.isdir(folder_name):
+                        print("{} already exists.\n".format(folder_name))
+                    else:
+                        os.mkdir(folder_name)
+                        break
+                else:
+                    folder_name = datetime.datetime.now().strftime(
```
IanMayo: we can't have ":" in pathnames for all operating systems. Happy with 8-digit date then 6-digit time `20200320_1855`

pepys_admin/admin_cli.py
tahirs95: Sure.

pepys_admin/admin_cli.py
tahirs95: Fixed.

PR #227, opened 2020-03-23
Introduce progress bar

BarisSari:


IanMayo:
nice lightweight solution, good work.

PR #235, opened 2020-03-23
Allow record call for highlighter to be used without units

robintw:
When we record a Quantity the units are present in the string conversion of the Quantity (eg. `"320 degree"`) so they don't need explicitly passing in the call to `record`. This PR adds the ability for the `units` parameter to be left out (defaulting to `None`) which then doesn't display it in the resulting tooltip. I've also changed all importers to remove the units parameter when a Quantity is passed.


pepys_import/file/highlighter/support/line.py
```diff
@@ -96,7 +96,10 @@ def record(self, tool: str, field: str, value: str, units: str = "n/a"):
         self.highlighted_file.fill_char_array_if_needed()
 
         tool_field = tool + "/" + field
```
IanMayo: @robintw - since a default value is provided for `units`, I guess that in the if structure below  `units` will never have a value of `None`.

(but that may be a shortcoming in my Python skills).

robintw: Good catch, that default value was meant to be removed - I'll push a new commit

robintw: Fixed in 
91eb3a8

Issue #241, opened 2020-03-23
Don't store `prev_location` in State


IanMayo:
I'd like to request a change to how we keep track of the previous location for a set of State measurements.

We currently store `prev_location` in each state as we work through them.  But this has to be done in every parser.

It would be more effective for the `common_db` `validate()` code to remember the previous state, and pass the new and previous states to `Basic` and `Enhanced` validators.

Once we have the previous state, we can pass it to `EnhancedValidator`, then on to `speed_loose_match_with_location` and it will have the current/previous locations _AND_ the current/previous timestamps, and it can calculate the speed to travel between them.

This will avoid us having to cache `prev_location` in each parser.

I've started to write the pseudo-code to perform the above in branch #hotfix/207_haversine_formula.

But, it won't work until we change how we handle `prev_location`.

BarisSari:
I think we can directly move the prev_location to the EnhancedValidator if we are going to pass all measurements to the validators (Issue #173). What do you think @IanMayo ?

IanMayo:
The issue is that we also need `prev_timestamp` in the enhanced validator.

Aaah, we could cache previous location _and_ time in the `validate()` loop, and pass them in.

It will only become clumsy if/when we require more previous attributes for quality testing.

Oh, and it means the API for all enhanced validators have to include `prev_location` and `prev_timestamp` parameters.

If they just contained `previous` and `current` state objects, it would be tidier, IMHO.

BarisSari:
Okay, I think you're right. It sounds better to send only previous and current state objects to the validator. I'll cache and assign `prev_location` and `prev_timestamp` in the `validate()` method.

PR #242, opened 2020-03-23
Hotfix/207 haversine formula

### LINKED ISSUE #207: Fix name or implementation of distance_between_two_points_haversine
The function `distance_between_two_points_haversine` in `unit_utils.py` states in its name that it calculates a distance, but actually returns a speed in metres per second.

Either the name should change, or the functionality should change, so it does what it says in the name.

From looking at the git history, I think @tahirs95 wrote this, so probably best for him to investigate.

IanMayo:
Fixes #207 and #241 

Ian started this, but he could do with @BarisSari handling the data management changes requested in #241


IanMayo:
Close. Fix included in #244

PR #244, opened 2020-03-24
207 Correct Haversine Formula and 241 Change Validators

### LINKED ISSUE #207: Fix name or implementation of distance_between_two_points_haversine

### LINKED ISSUE #241: Don't store `prev_location` in State

BarisSari:
Since they are highly related to each other, I combined both issues in this PR. 

At first, I changed `DatafileMixin` methods. `Measurement` objects are cached with the following format:

```python
measurements = {
    "Importer-1" = {
        "Platform-1" = List of objects,
        "Platform-2" = List of objects,
          ...
    },
    "Importer-2" = {
        "Platform-3" = List of objects,
        "Platform-4" = List of objects,
          ...
    }
}
```

Therefore, it is easy to keep track of previous `Measurement` objects even though there are other examples between them. After this dictionary change has been made, I changed the arguments of Enhanced Validator. It accepts `prev` and `current` measurement objects now:
https://github.com/debrief/pepys-import/blob/85ab840a056a6cbd2c94914e7f50b3d0bdde6ee7/pepys_import/core/validators/enhanced_validator.py#L14

Then, I corrected the return of the haversine function, which is the distance in meters right now: https://github.com/debrief/pepys-import/blob/85ab840a056a6cbd2c94914e7f50b3d0bdde6ee7/pepys_import/utils/unit_utils.py#L115
I implemented a basic `calculate_time` method:
https://github.com/debrief/pepys-import/blob/522496d75bd1c665c023912d4db9b481e7a784bf/pepys_import/core/validators/enhanced_validator.py#L103-L105
Finally, I changed `speed_loose_match_with_location` method:
https://github.com/debrief/pepys-import/blob/85ab840a056a6cbd2c94914e7f50b3d0bdde6ee7/pepys_import/core/validators/enhanced_validator.py#L106-L110

As a result, this PR fixes #241 , fixes #207 .


pepys_import/core/store/common_db.py
```diff
@@ -189,13 +196,18 @@ def validate(
...
         elif validation_level == validation_constants.ENHANCED_LEVEL:
-            for measurement in self.measurements[parser]:
-                BasicValidator(measurement, errors, parser)
-                for basic_validator in LOCAL_BASIC_VALIDATORS:
-                    basic_validator(measurement, errors, parser)
-                EnhancedValidator(measurement, errors, parser)
-                for enhanced_validator in LOCAL_ENHANCED_VALIDATORS:
-                    enhanced_validator(measurement, errors, parser)
+            for key, objects in self.measurements[parser].items():
+                for index, curr_object in enumerate(objects):
```
IanMayo: @BarisSari - I don't think this algorithm can handle when a datafile holds data for multiple ships.  I'm pretty sure we need a dictionary of previous objects, indexed by either platform name or sensor.

Oh, and here's a simpler logic for remembering the previous value (though it will need modifying for the multiple platforms):
```
prev_object = None
for index, curr_object in enumerate(objects):
    if index >= 1:
    [...]
    for enhanced_validator in LOCAL_ENHANCED_VALIDATORS:
        enhanced_validator(curr_object, errors, parser, prev_object)
    prev_object = curr_object
```

pepys_import/core/store/common_db.py
```diff
@@ -204,13 +216,14 @@ def commit(self, data_store, change_id):
...
         # and save its measurement objects.
         extraction_log = list()
-        for key in self.measurements.keys():
-            print(f"Submitting measurements extracted by {key}.")
-            for file in tqdm(self.measurements[key]):
-                file.submit(data_store, change_id)
-            extraction_log.append(
-                f"{len(self.measurements[key])} measurements extracted by {key}."
-            )
+        for key in self.measurements:
```
IanMayo: I know it's a `key` in computing terms,  but in business domain it's a `parser` isn't it?

BarisSari: Yes, it is. I'll change it now

pepys_import/core/validators/enhanced_validator.py
```diff
@@ -11,40 +11,37 @@
...
+        self.time = current_object.time if hasattr(current_object, "time") else None
+
+        if prev_object:
+            self.prev_location = (
+                prev_object.location if hasattr(prev_object, "location") else None
+            )
+            self.prev_time = prev_object.time if hasattr(prev_object, "time") else None
+
+            if self.location and self.prev_location:
+                self.course_heading_loose_match_with_location()
```
IanMayo: @BarisSari - does Python have any best practices on passing parameters to methods?  My `traditional` preference would be for the method fields to be passed as parameters, so the inline documentation can detail what the method expects/uses.

But, as I said - are there any pythonic traditions for this?

pepys_import/core/store/common_db.py
BarisSari: Corrected:
https://github.com/debrief/pepys-import/pull/244/commits/70a71543afae432dd84ec81316cfe7421b7c03e3

pepys_import/core/store/common_db.py
IanMayo: Hey, don't I look silly for not reading your PR description 🤣 

I presume your code handles just the problem I was discussing, right @BarisSari ?

BarisSari: That's fine 😄 yes, I think we already have example datafiles that hold data for multiple ships. The algorithm handles them without any problem

pepys_import/core/validators/enhanced_validator.py
BarisSari: Actually, I'm not sure if there are any pythonic traditions for this. I think we can turn them into static methods and send the location-time variables. In this situation, we don't need instance variables, so I can remove the self keyword from them. Does that make sense?

IanMayo: Yes, it would be my preference to make them static - it's a more functional approach, meaning:
* the only data they know about is what was passed in
* the method returns it's results, without any internal changes being made

BarisSari: Okay, I'll make them static then.

BarisSari: Both Basic and Enhanced Validators are static now (docstrings also added):
https://github.com/debrief/pepys-import/pull/244/commits/cbf0b05c606b0fde7e20cecc182bd1597a7af374
https://github.com/debrief/pepys-import/pull/244/commits/0181848009cf04bb36e19058e24fa6e30b9846e7

pepys_import/core/store/common_db.py
IanMayo: @BarisSari  - I think this is still simpler logic for remembering the previous measurement:
```
prev_object = None
for index, curr_object in enumerate(objects):
    [...]
    for enhanced_validator in LOCAL_ENHANCED_VALIDATORS:
        enhanced_validator(curr_object, errors, parser, prev_object)
    prev_object = curr_object
```

BarisSari: We should keep the measurement objects of the datafile in any case. Because we are committing them to the DB when they are validated. I changed only the dictionary (self.measurements), so that we can reach the previous objects easier. So, we don't need to keep track of the previous objects in any other list or variable. Right now, we aren't bothered with the correct assignment of `prev_object` according to its' platform because we are already keeping them according to their platform. So, it can't be wrong. Don't you think that it's more convenient?

IanMayo: I may be misinterpreting the code. Well, or not explaining myself.  We currently have this:
```
        elif validation_level == validation_constants.ENHANCED_LEVEL:
            for key, objects in self.measurements[parser].items():
                for index, curr_object in enumerate(objects):
                    prev_object = None
                    if index >= 1:
                        prev_object = objects[index - 1]
                   # 
                   # do tests
                   # 
```

I'm proposing it would be simpler like this:
```
        elif validation_level == validation_constants.ENHANCED_LEVEL:
            for key, objects in self.measurements[parser].items():
                prev_object = None
                for curr_object in enumerate(objects):
                   # 
                   # do tests
                   # 
                   prev_object = curr_object
```
In the latter version we're not using/checking the index counter, or doing array access.

BarisSari: Okay, @IanMayo . I understood what you're proposing. I modified it a bit for keeping track of objects according to their platform. There is no index counter and array access anymore.

https://github.com/debrief/pepys-import/blob/c58b2fc3ef53f685ff9acb79d21a9996c2ce1e7c/pepys_import/core/store/common_db.py#L200-L212

PR #249, opened 2020-03-25
Hotfix/245 Remove Default Values

### LINKED ISSUE #245: Should we update default config file and fallback config options to match?
While writing the docs for the configuration file, I noticed that the default config file specifies various settings (basically an in-memory SQLite database), but when the config file is loaded, any entries that are missing will be replaced with values that are different to the default config file values - they are actually values for a local Postgres server.

Would it make sense for these to match, so we always default to the values shown in the default config file?

BarisSari:
Fixes #245

### DIFF
--- a/config.py
+++ b/config.py
@@ -18,12 +18,12 @@ config.read(CONFIG_FILE_PATH)
 assert config.has_section("database"), f"'database' section couldn't find in '{CONFIG_FILE_PATH}'!"
 
 # Fetch database section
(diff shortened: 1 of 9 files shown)


IanMayo:
>  But, I don’t think that we assumed it in any tests. Anyway, I’ll also delete it

🤣

BarisSari:
> > But, I don’t think that we assumed it in any tests. Anyway, I’ll also delete it
> 
> 🤣

Apparently, we had omitted it in all Postgres related tests 😄

PR #253, opened 2020-03-26
Refactor Admin CLI - Introduce Tests (#219-#225)

### LINKED ISSUE #55: REP export capability

### LINKED ISSUE #219: Test coverage for Admin app
We don't have a lot of test coverage for the Admin app - largely because the "big push" was just to get something workable out there.

Now that we appear to have more time, we should increase the test coverage, and potentially do some refactoring.

I'm assigning this to @BarisSari so that we get some wider familiarity with the admin app.

_Maybe_ the task should start with an architectural review - to verify we're taking the correct approach.  There is probably some merit in breaking the app down into modules in cases where the UI handlers have substantial blocks of code.

### LINKED ISSUE #225: Datafile export tests should be more comprehensive
At the moment the datafile export tests (in `test_export_data.py`) check that one of the default datafiles imported with the initialisation data (`DATAFILE-1`) can be exported without any errors. This is a problem for a few reasons:

1. This datafile contains very little data - for example, no contact or comment data, as far as I can see. So these parts of the code are not exercised properly.

2. There is no checking of what the output datafile actually contains, to check the output has worked properly.

This meant that the tests didn't catch a number of errors caused by the integration of properties and units to the Contacts class, that I only caught because I was surprised there were no test errors and looked at the code myself to check.

BarisSari:
Fixes #219 .
Fixes #225 .
Fixes #55 .

In this PR the following issues are solved:
>  If the database doesn't exist, the scripts should report that in an error message to the CLI, the scripts should not fail.

>   In the clear database action, the scripts should not fail if the database is empty. So, check each table exists, before trying to clear it.

The following check is added to the related methods:
https://github.com/debrief/pepys-import/blob/0c6f3ec97800ee82d416025a9c9c0725b62794c2/pepys_import/core/store/data_store.py#L1354-L1367

----

> This datafile (_example datafiles.csv_) contains very little data - for example, no contact or comment data, as far as I can see. So these parts of the code are not exercised properly.

> There is no checking of what the output datafile actually contains, to check the output has worked properly.

I modified [test_data_store_export_datafile.py](https://github.com/debrief/pepys-import/blob/refactor_admin_cli/tests/test_data_store_export_datafile.py), so that tests import `rep_test1.rep `(It has State, Comment, and Contact objects). Then, tests check the written file at the end. I also added new tests that hit the exceptions of the caching methods.

Finally, I wrote tests for admin CLI methods in a different file: https://github.com/debrief/pepys-import/blob/refactor_admin_cli/tests/test_admin_cli.py

**Note:** I have checked mostly the prints in the admin CLI tests because we already have functionality tests of DataStore methods (`test_data_store_export_datafile.py`, `test_data_store_clear_db.py`, etc.). So, we are sure that DataStore methods are working. These admin CLI tests are only checking whether the admin CLI method worked correctly and printed something at the end of the methods. 

**Note-2:** CLI is divided into 3 different modules. Therefore, pepys_admin command has changed. In order to run, please use the following command:
- `python -m pepys_admin.cli`


pepys_admin/admin_cli.py
```diff
@@ -1,99 +1,20 @@
...
-            print(self.intro)
-        return cmd.Cmd.postcmd(self, stop, line)
-
-
 class AdminShell(cmd.Cmd):
     intro = """--- Menu ---
 (1) Export
 (2) Initialise/Clear
 (3) Status
+(9) Export All
```
IanMayo: @BarisSari - option 9 was deliberately hidden.  `Export All` isn't a usage of Pepys that we encourage, but it's something that's useful for testing.

BarisSari: Okay. It's deleted:
https://github.com/debrief/pepys-import/pull/253/commits/31d0eaa1b38daba4d6920a7557fcd0e705922cce

IanMayo:
Renaming the top level command is great, well done.

PR #257, opened 2020-03-27
Hotfix - Remove un-necessary self.errors initialisation in importers

robintw:
This has been done in the parent class for a while, but hadn't been removed from the subclasses, duplicating code and causing confusion. This PR removes the superfluous code.


IanMayo:
Have run tests locally, all pass.

Issue #54, opened 2020-02-14
More capable REP Parser


IanMayo:
The existing REP parser only handles the simplest REP formats.

This document details the format: https://debrief.github.io/tutorial/reference.html#replay_track_format

We should verify/test that the parser handles these idiosyncracies:
* [ ] Decimal degrees, with zeros for mins/seconds
* [ ] Decimal minutes, with zeroes for seconds
* [ ] `NaN` for depth
* [ ] Presence of extended symbology fields (just ignore them)

robintw:
I've been working on this in #258, and have a question:

In the REP parser we say:

```
if depth_token == "NaN":
                self.depth = 0.0
```

Do we actually want to do this? Surely if the depth is set to NaN then we want to record depth as None - ie. a missing value?

IanMayo:
Yes - you're completely right.  

But, I'm not sure whether we should `set value as None` or `not set the value`.  I wonder how the database handles it.  If we fail to set a value, I wonder if it gets initialised as zero?

Any thoughts @robintw  or @BarisSari ?

robintw:
Fixed by #258 and #261

PR #258, opened 2020-03-27
Extend REP parser tests 

robintw:
Now checks all of the situations specified in https://github.com/debrief/pepys-import/issues/54

Specific things to add to tests:
* [x] Decimal degrees, with zeros for mins/seconds
* [x] Decimal minutes, with zeroes for seconds
* [x] `NaN` for depth
* [x] Presence of extended symbology fields (just ignore them)

### DIFF
--- a/tests/test_rep_file_parse.py
+++ b/tests/test_rep_file_parse.py
@@ -2,6 +2,7 @@ import datetime
 import unittest
 from contextlib import redirect_stdout
 from io import StringIO
+from math import isnan
 
 from pepys_import.core.formats import unit_registry
 from pepys_import.core.formats.location import Location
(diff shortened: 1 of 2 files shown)


tests/test_rep_file_parse.py
```diff
@@ -162,6 +163,58 @@ def test_line_ok(self):
...
+    def test_extended_symbology(self):
+        rep_line = REPLine(
+            line_number=1,
+            line=create_test_line_object(
+                "100112\t120800\tSUBJECT\t@C[SYMBOL=torpedo,LAYER=Support]\t60\t23\t40.25\tS\t000\t01\t25.86\tE\t109.08\t6.00\t0.0\tLabel"
+            ),
+            separator="\t",
+        )
+        assert rep_line.parse(self.error, self.message)
+
```
IanMayo: @robintw  - do we need any additional asserts in here, just to check that, for example, the lat/long got loaded?

robintw: Good idea. Fixed in 7d8d684

PR #261, opened 2020-03-29
Fix processing of NaN for depth in a REP file.

robintw:
Now treats NaN as missing data, and assigns a value of None, which is picked up by the databases (both SQLite and Postgres) as a field value of NULL.


IanMayo:
There's very little code here - so I'm sure it works fine.  

But, it would be useful to include a test, just to check it doesn't break in the future, please.

robintw:
Of course, test added in 4abebb8.

IanMayo:
@robintw  - I'm getting a test failure:
```
_____________________________________________ BasicTests.test_nan_depth ______________________________________________

self = <tests.test_rep_file_parse.BasicTests testMethod=test_nan_depth>

    def test_nan_depth(self):
        rep_line = REPLine(
            line_number=1,
            line=create_test_line_object(
                "100112\t120800\tSUBJECT\tVC\t60\t23\t40.25\tS\t000\t01\t25.86\tE\t109.08\t6.00\tNaN\tLabel"
            ),
            separator="\t",
        )
        assert rep_line.parse(self.error, self.message)
    
>       assert isnan(rep_line.depth)
E       TypeError: must be real number, not NoneType

tests/test_rep_file_parse.py:206: TypeError
================================ 1 failed, 368 passed, 6 xfailed in 650.08s (0:10:50) =========
```

This is from running `pytest tests`.

PR #265, opened 2020-03-30
Simplify __init__ for importers

robintw:
This PR converts the __init__ method of importers, so that we don't pass arguments to __init__ just to super()._init__. Generally makes it simpler for parser-writers.

### DIFF
--- a/importers/replay_contact_importer.py
+++ b/importers/replay_contact_importer.py
@@ -1,5 +1,3 @@
-from tqdm import tqdm
-
 from pepys_import.core.formats import unit_registry
 from pepys_import.core.formats.location import Location
 from pepys_import.core.formats.rep_line import parse_timestamp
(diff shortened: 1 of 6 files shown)


importers/replay_contact_importer.py
```diff
@@ -14,12 +14,11 @@
...
-        self,
-        name="Replay Contact Importer",
-        validation_level=constants.ENHANCED_LEVEL,
-        short_name="REP Contact Importer",
-    ):
+    def __init__(self):
+        name = "Replay Contact Importer"
+        validation_level = constants.ENHANCED_LEVEL
+        short_name = "REP Contact Importer"
+
```
IanMayo: @robintw - aren't we still creating some objects unnecessarily?

Can't we use: 
```
super().__init__(
        name = "Replay Contact Importer"
        validation_level = constants.ENHANCED_LEVEL
        short_name = "REP Contact Importer"
)
```

We may require the `short_name` later on, for error reporting, but I guess the parent-level version is available .

robintw: Yes of course we can - I was obviously being dumb yesterday. Fixed now.

PR #266, opened 2020-03-31
Feature/Export datafile objects by platform name and date

BarisSari:


pepys_admin/admin_cli.py
```diff
@@ -12,8 +13,9 @@
 class AdminShell(cmd.Cmd):
     intro = """--- Menu ---
 (1) Export
-(2) Initialise/Clear
-(3) Status
```
IanMayo, line 16: @BarisSari - I think we've actually implemented "Export by platform and sensor" (which is quite ok).  So, Option 2 should be changed, please.

pepys_import/core/store/data_store.py
```diff
@@ -1215,33 +1223,160 @@ def get_cached_platform_name(self, sensor_id=None, platform_id=None):
...
+    def export_datafile(self, datafile_id, file_path, sensor_id=None, platform_id=None):
         """Gets states, contacts and comments of a Datafile.
 
         :param datafile_id:  ID of Datafile
         :type datafile_id: Integer or UUID
         :param file_path: Path of a file to export
         :type file_path: String
+        :param sensor_id: ID of Sensor to export a specific sensor in the datafile, default is None
+        :type sensor_id: Integer or UUID
+        :param platform_id: ID of Sensor to export a specific sensor in the datafile, default is None
```
IanMayo, line 1333: I think the parameter description is wrong, I presume it's the platform.

BarisSari: Yes, it was a copy-paste error. Corrected:
https://github.com/debrief/pepys-import/pull/266/commits/64872c9671e6f6511f50d3b66dae0e745057df05

pepys_admin/admin_cli.py
BarisSari: Changed:
https://github.com/debrief/pepys-import/pull/266/commits/e4fdc59f34cdfaa142c9dcb5bb582efeffb3e985

PR #267, opened 2020-03-31
Updated deployment script to deploy full dev environment

robintw:
The deployment script now installs from both requirements.txt and requirements_dev.txt.

Also updated zipping procedure for creating final deployment zip: this now uses 7zip to produce the zip file due to bugs found in the Create-Archive Powershell command.

### DIFF
--- a/create_deployment.ps1
+++ b/create_deployment.ps1
@@ -66,9 +66,10 @@ Remove-Item get-pip.py
 Write-Output "INFO: Cleaned up all except 7zip"
 
 # Zip up whole folder into a zip-file with the current date in the filename
+# excluding the 7zip folder
 $date_str = Get-Date -Format "yyyyMMdd"
 $output_filename = $date_str + "_pepys-import.zip"
-.\7zip\7za.exe a .\$output_filename .\*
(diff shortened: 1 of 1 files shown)


create_deployment.ps1
```diff
@@ -54,22 +54,25 @@ Lib\site-packages
...
+
+# Remove folders/files that we don't need any more
+Remove-Item .\7zip -Recurse
+Remove-Item sqlite.zip
+Remove-Item python.zip
+Remove-Item 7zip.zip
+Remove-Item *.7z
+Remove-Item get-pip.py
 
-Write-Output "INFO: Written zipped deployment file to $output_filename"
```
IanMayo: Aah, we now do the `remove-item` after the zip.  Does that mean we zip up elements that we weren't zipping up before?

robintw: Yeah, sorry I missed that - I ended up changing things around so many times that I lost track. I've changed it now so that most of the cleanup happens before the final zip, except for the 7zip folder which we still need to keep available. I then exclude that 7zip folder from the zip file we create, and then delete it after we've created the zip file.

PR #268, opened 2020-03-31
Marked all tests that require Postgres, so that they can be excluded from test runs

robintw:
- Added a `pytest.ini` configuration file to specify a 'custom mark' of `postgres`
 - Marked all tests that require postgres with `@pytest.mark.postgres`

To run all tests *except* the Postgres tests, run:

```
pytest tests/ -m "not postgres"
```

To run just the Postgres tests, run:

```
pytest tests/ -m postgres
```

### DIFF
--- a/DeveloperGuide.rst
+++ b/DeveloperGuide.rst
@@ -158,6 +158,11 @@ To run the unittests run::
 
     $ pytest tests/
 
+To run the unittests excluding the tests that require PostgreSQL (for example, if you couldn't install
+PostgreSQL earlier)::
+
+    $ pytest tests/ -m "not postgres"
+
 To view the coverage of the test suite, run::
 
     $ coverage run -m pytest tests/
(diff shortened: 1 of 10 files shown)


IanMayo:
@robintw  - I presume you can see that the tests are failing on Travis:
https://travis-ci.com/github/debrief/pepys-import/builds/157047903#L940

robintw:
Yeah it was a silly mistake, I thought I'd pushed the update but hadn't. Now pushed and should pass soon.

BarisSari:
I have a question @robintw. Isn't it possible to mark classes instead of each method? I think it could be much more easier to mark.

robintw:
Thanks @BarisSari - yes it is possible, so I've switched to that. Much cleaner - thanks.

IanMayo:
Aah @robintw - I only realised this after merging - we should document this marker _somewhere_ in the dev docs (maybe including the parser writing guidance)

robintw:
@IanMayo The docs were actually included in this, I just forgot to update the main PR text to explain that they were included. They're in the first two files that show under 'Files changed'.

PR #272, opened 2020-03-31
Added support for parsing depth fields in NMEA files

### LINKED ISSUE #117: Introduce depth/elevation data in NMEA sample file
NMEA data can include depth, formatted like this:
```
$POSL,PDS,9.2,M*0
```
We should determine:
* units
* is it depth or elevation?

Once the above are resolved, we should insert some sample depth data into the `NMEA_Trial.log`.

robintw:
- Support for new PDS message type, with lines like "$POSL,PDS,9.2,M*0"
 - Field is optional, so not included when checking that we have all fields before creating a State
 - Added tests, including that the correct time is associated with the depth data

Fixes #117

### DIFF
--- a/importers/nmea_importer.py
+++ b/importers/nmea_importer.py
@@ -5,7 +5,7 @@ from pepys_import.core.formats.location import Location
 from pepys_import.core.validators import constants
 from pepys_import.file.highlighter.support.combine import combine_tokens
 from pepys_import.file.importer import Importer
-from pepys_import.utils.unit_utils import convert_absolute_angle, convert_speed
+from pepys_import.utils.unit_utils import convert_absolute_angle, convert_distance, convert_speed
 
 
 class NMEAImporter(Importer):
@@ -25,6 +25,7 @@ class NMEAImporter(Importer):
         self.time = None
         self.heading = None
         self.speed = None
+        self.depth = None
         self.date_token = None
         self.time_token = None
         self.speed_token = None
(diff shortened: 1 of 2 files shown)


importers/nmea_importer.py
```diff
@@ -152,6 +156,14 @@ def _load_this_line(self, data_store, line_number, line, datafile, change_id):
                     state.speed = speed
                 self.speed_token.record(self.name, "speed", speed)
 
+                if self.depth is not None:
+                    depth = convert_distance(
+                        self.depth, unit_registry.metre, line_number, self.errors, self.error_type
+                    )
+                    if depth:
+                        state.elevation = -1 * depth
+                    self.depth_token.record(self.name, "depth", depth)
```
IanMayo, line 165: @robintw - I think line #165 should be indented. We only want to record it if we get a value, I guess.

robintw: Actually I think the code is correct. The `record` line is indented under the `if self.depth is not None` line - so it only runs if we've extracted a depth. The `if depth` line checks whether the depth could actually be converted to a float by `convert_distance`. If it can't be converted to a float then we don't want to submit it to the database, but I think we _do_ want to record the extraction so that the user can see in the highlighted output what chunk of the file was attempted to be extracted as the depth.

That's the logic that is used in the rest of the file anyway - I guess it really depends how we want the highlighted output to work - what do you think @IanMayo

IanMayo: Sure, I see your logic.  But, I think it could mistakenly give the impression that the field had been successfully imported.  I'm _expecting_ the `.record()` method implementations to also handle recording the extractions to the database - again this should only be for successful extractions.

So, let's only record the capture if it's successfully parsed, please.

robintw: Ok, that'll require changes to all the importers (and I've just realised some of them don't check for successful parsing before writing to the database).

Is it ok if you approve this PR first, and then I do a new PR that changes that for all the importers?

PR #273, opened 2020-03-31
Switch to command line resolver for import CLI

robintw:
The import CLI was still using the default resolver for everything. We have a working command-line resolver, so this switches the importer to use that.

There are a few things I want to ask/check though:

1. @BarisSari: When trying to import the test GPX files with the CommandLineResolver it crashed with an error. I've traced it through a bit (and in the process understood more about how the resolvers work), and it seems that something was being passed a `Nationality` object rather than a string containing the name of a Nationality, which is what it expected. It then tried to do an invalid SQL query where it tried to do `WHERE Nationality.name == <Nationality Object>` which obviously failed. I've fixed this by changing it to pass a name - which seems to be what the default resolver does too - but it'd be great if you could give this a quick sanity check to make sure I'm not doing anything silly.

2. @IanMayo: The command-line resolver is great, but it means we have to answer loads of questions when we run the import CLI, which makes developing/debugging the CLI difficult. What do you think about adding a `--default-resolver` option to the CLI to use the DefaultResolver instead, mainly for testing purposes (also for parser developers, so they can test imports without having to answer questions all the time).

### DIFF
--- a/create_deployment.ps1
+++ b/create_deployment.ps1
@@ -54,22 +54,26 @@ Lib\site-packages
 
 Write-Output "INFO: Set Python pth file"
 
-# Do a standard pip install of the requirements, not warning us that scripts will be unavailable
-.\python\python.exe -m pip install -r requirements.txt --no-warn-script-location
+# Do a standard pip install of the requirements and dev requirements, not warning us that scripts will be unavailable
+.\python\python.exe -m pip install -r requirements.txt -r requirements_dev.txt --no-warn-script-location
 
 Write-Output "INFO: Installed Python dependencies"
 
-# Remove folders/files that we don't need any more
-Remove-Item .\7zip -Recurse
 Remove-Item *.zip
 Remove-Item *.7z
 Remove-Item get-pip.py
 
-Write-Output "INFO: Finished cleanup"
+Write-Output "INFO: Cleaned up all except 7zip"
 
 # Zip up whole folder into a zip-file with the current date in the filename
+# excluding the 7zip folder
 $date_str = Get-Date -Format "yyyyMMdd"
 $output_filename = $date_str + "_pepys-import.zip"
-Compress-Archive -Path .\* -DestinationPath $output_filename
+.\7zip\7za.exe a .\$output_filename .\* -xr!7zip/
 
 Write-Output "INFO: Written zipped deployment file to $output_filename"
+
+# Remove folders/files that we don't need any more
+Remove-Item .\7zip -Recurse
+
+Write-Output "INFO: Finished cleanup"
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -20,14 +20,22 @@ Command-line options
 
 .. code-block:: none
 
-   usage: import.py [-h] [--path PATH] [--archive]
+  usage: import.py [-h] [--path PATH] [--archive] [--db DB]
+                  [--resolver RESOLVER]
 
-   optional arguments:
-   -h, --help   show this help message and exit
-   --path PATH  The path to import data from (The default value is the
-             directory of the script)
-   --archive    Instruction to archive (move) imported files to designated
-                archive folder
+  optional arguments:
+    -h, --help           show this help message and exit
+    --path PATH          The path to import data from (The default value is the
+                        directory of the script)
+    --archive            Instruction to archive (move) imported files to
+                        designated archive folder
+    --db DB              SQLite database file to use (overrides config file
+                        database settings). Use `:memory:` for temporary in-
+                        memory instance
+    --resolver RESOLVER  Resolver to use for unknown entities. Valid values:
+                        'default' (resolves using static default values),
+                        'command-line' (resolves using interactive command-line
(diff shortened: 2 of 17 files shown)


robintw:
Actually @BarisSari, the tests for the command-line resolver aren't passing now - so I think my 'fix' may have been misguided. If you're able to look into this and then share what you think the problem is then that'd be really useful.

BarisSari:
> The import CLI was still using the default resolver for everything. We have a working command-line resolver, so this switches the importer to use that.
> 
> There are a few things I want to ask/check though:
> 
>     1. @BarisSari: When trying to import the test GPX files with the CommandLineResolver it crashed with an error. I've traced it through a bit (and in the process understood more about how the resolvers work), and it seems that something was being passed a `Nationality` object rather than a string containing the name of a Nationality, which is what it expected. It then tried to do an invalid SQL query where it tried to do `WHERE Nationality.name == <Nationality Object>` which obviously failed. I've fixed this by changing it to pass a name - which seems to be what the default resolver does too - but it'd be great if you could give this a quick sanity check to make sure I'm not doing anything silly.
> 
>     2. @IanMayo: The command-line resolver is great, but it means we have to answer loads of questions when we run the import CLI, which makes developing/debugging the CLI difficult. What do you think about adding a `--default-resolver` option to the CLI to use the DefaultResolver instead, mainly for testing purposes (also for parser developers, so they can test imports without having to answer questions all the time).

First of all, I agree that there will be lots of questions to answer to import a file. Therefore, it would be better to have an option to use the default resolver.

When I refactored the Command Line Resolver, we didn't have an appropriate import CLI, and a few of the Importers were missing. However, when I compare the platform-sensor creation of GPX Importer with the Replay Importer(which we had it that time), I don't see any differences at all. They both send a String as a Nationality parameter. I think we might have some other problems.

By the way, I was thinking to solve this issue once I refactor the Command Line Resolver again. Because even though we have lots of tests for it, it has not been tried importing our track_files yet. If that's okay, I'd like to close this PR for now and open it when the second refactoring is done. Does it sound sensible @IanMayo and @robintw ?

BarisSari:
I debugged the GPX Importer now because we should solve this issue anyway. "**gpx_1_0_MultipleTracks.gpx**" was causing the error. The second platform, "**TestPlatform**", have the same `Nationality, Platform Type` and `Privacy` values. However, instead of their names, their objects (DB entities) were passing to the command line resolver. So, we have a design problem. Default Resolver and Command Line Resolver don't work the same way for now. That's why we should modify one of them. Otherwise, each change will cause the other's failure.

IanMayo:
Hello @robintw & @BarisSari :-)

> Therefore, it would be better to have an option to use the default resolver. 

Yes, we will require the defaults resolver for our test-suite.  Specifying it on the command-line (or via parameter in the unit-tests) is a good strategy.  I suspect we'll find more value in the parameter being `--resolver default` than `--default-resolver`, it's versatility in specifying other resolvers could help us in the future.

> Default Resolver and Command Line Resolver don't work the same way for now. That's why we should modify one of them.

That's ok - I was keen to re-introduce the Command-Line resolver to identify issues like this.  We have to decide which one to alter. We _could_ see which strategy requires the least change.  Or, we _could_ decide which strategy either has greater engineering beauty or is most Pythonic.

Any preferences?

> By the way, I was thinking to solve this issue once I refactor the Command Line Resolver again. 

I'm hoping to get a release out in the next couple of days. I could really do with a fix before we do any refactoring.

Can we agree on a strategy (`string` or `object`) and fix the _other_ resolver?

BarisSari:
> Hello @robintw & @BarisSari :-)
> 
> > Therefore, it would be better to have an option to use the default resolver.
> 
> Yes, we will require the defaults resolver for our test-suite. Specifying it on the command-line (or via parameter in the unit-tests) is a good strategy. I suspect we'll find more value in the parameter being `--resolver default` than `--default-resolver`, it's versatility in specifying other resolvers could help us in the future.
> 
> > Default Resolver and Command Line Resolver don't work the same way for now. That's why we should modify one of them.
> 
> That's ok - I was keen to re-introduce the Command-Line resolver to identify issues like this. We have to decide which one to alter. We _could_ see which strategy requires the least change. Or, we _could_ decide which strategy either has greater engineering beauty or is most Pythonic.
> 
> Any preferences?
> 
> > By the way, I was thinking to solve this issue once I refactor the Command Line Resolver again.
> 
> I'm hoping to get a release out in the next couple of days. I could really do with a fix before we do any refactoring.
> 
> Can we agree on a strategy (`string` or `object`) and fix the _other_ resolver?

I think it's easier to modify the default resolver to accept strings. Command line resolver's structure is much more complex and it would trigger many failures if we make a change on it.

robintw:
I think this is done now - thanks @BarisSari for your hotfix PR earlier to fix the resolvers, everything seems to be working now.

Basically all this PR does now is:

 - Add a new command line option to the import script of `--resolver`, which defaults to `command-line` but can be set to `default` instead to use the DefaultResolver.

PR #283, opened 2020-04-01
Change destination of logging and safe copy files (#187)

### LINKED ISSUE #187: Enhancement - logging & safe copy file destination
I've just had a first try at conducting import with the archive folder specified as an environment folder.

The logging files were written to a folder path constructed with the full date path. But, the successfully importer input files were just placed in a to level folder titled 'input files'.

We haven't come to a decision on how to organise the reports and input files within the folder, but they should all go into the folder with the relevant date/time path.

Talking to the user representative this afternoon, his preference was for this format:
* reports (extraction highlights, success/failure reports
* sources (copies of original source files)

So, there won't _currently_ be any files in the top-level folder, just those two sub-folders.

_Maybe_ one day we'll have an import summary that gives an overview of how things went.  If we had that, the import process could finish with "Do you wish to view the import summary (Y/n)". One day.

BarisSari:
Fixes #187

### DIFF
--- a/default_config.ini
+++ b/default_config.ini
@@ -8,7 +8,7 @@ db_type = sqlite
 [archive]
 user = pepys_admin
 password = 123456
-path = test_archive
+path = archive
 [local]
 parsers =
 basic_tests =
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -62,18 +62,13 @@ class FileProcessor:
             self.output_path = os.path.join(dir_path, "output")
             if not os.path.exists(self.output_path):
                 os.makedirs(self.output_path)
-        # create input_files folder if not exists
-        self.input_files_path = os.path.join(self.output_path, "sources")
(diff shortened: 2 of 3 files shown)


IanMayo:
@BarisSari - the output format for this still doesn't look right.

The output structure should `reports` and `sources` in the top level of the dated folder, like this:
2020/04/02/10/12/03/reports
2020/04/02/10/12/03/sources

BarisSari:
> @BarisSari - the output format for this still doesn't look right.
> 
> The output structure should `reports` and `sources` in the top level of the dated folder, like this:
> 2020/04/02/10/12/03/reports
> 2020/04/02/10/12/03/sources

Aah okay. When you said the top level, I thought the `output/` folder. I'm changing it.

BarisSari:
@IanMayo , It creates folders(reports and sources) under the dated folder now:
https://github.com/debrief/pepys-import/pull/283/commits/52a03d12b08c1f5739e3c82a9b29379d0a4c8eae

PR #291, opened 2020-04-02
Added Getting Started tutorial to docs

### LINKED ISSUE #279: Pepys Parser - Writer, getting started guidance
I'm not sure if this a "gap" or not.

Tomorrow or Friday I'm going to invite our client(s) to download Pepys and learn their way around it.

We have some technical level guidance on this, but I think we could do with some higher level guidance, maybe a walkthrough. 

But, I'm not hoping to duplicate the parser-writer guidance.  Aah, that's it.  This is guidance for Pepys users, not parser-writers. In this case they will also be parser-writers, but they've got to get past the first hurdle before they can do that.

I "guess" this could include:
* download/unpack the Pepys distribution
* creating the shortcuts
* view contents of SQLite config/ini file - not database & archive names.
* setting the Environment Variable for the .ini file.
* verify that the database isn't initially present
* Running an import cycle on a single file in `tests/sample_data/track_files` (CLI resolver)
* Check the database contents via `pepys-admin`
* export a file using `pepys-admin`
* take a copy of the track_files folder, since we're going to do a destructive operation (import/archive)
* run the import cycle, including `archive`, and `default` resolver
* examine the archive folder contents
* drill down into an import problem/error that we can fix
* fix the problem in a data-file, then re-run the import - check the new file gets imported.
* check the fixed file gets archived: both removed from source AND introduced in archive location
* demo/screenshot of opening SQLite database in SQLite Browser, show contents.  (We don't distribute this, so we don't need to direct user, we can just show screenshots)
* delete temporary track_file duplicate folder, and archive folder

robintw:
Fixes #279.

Complete first draft at the moment, happy to work on any edits suggested by @BarisSari or @IanMayo.


docs/getting_started.rst
```diff
@@ -0,0 +1,185 @@
...
+    | Platforms    |                1 | 2020-04-02 14:10:54.507565 |
+
+*Note:* The dates will be different in your example, but the number of rows should be the same.
+
+5. Check the database status using Pepys Admin
+----------------------------------------------
+
+Run *Pepys Admin* from the Start Menu (you can either navigate to the *Pepys* folder and choose
+*Pepys Admin* or just search for it and press enter).
+
```
IanMayo, line 63: @robintw - could you put in a reminder that if `Pepys Admin` isn't present, they should check they've run the correct steps from the installation guide.

docs/getting_started.rst
```diff
@@ -0,0 +1,185 @@
...
+installation folder to a safe place - as the example we're going to run will move files.
+
+Open the Windows Command Prompt (open the *Start Menu* and type :code:`cmd` and press Enter) and
+use the :code:`cd` command to navigate to the Pepys install folder. Then navigate to the :code:`bin`
+directory and run :code:`set_paths.bat`. Move back to the main Pepys install folder with :code:`cd ..`,
+and run the following:
+
+.. code-block:: none
+
+    python -m pepys_import.import --path .\tests\sample_data\track_files\rep_data\rep_test1_bad.rep --resolver default --archive
```
IanMayo, line 103: @robintw - this should be the path for the new folder. If we suggest a folder-name to use, we can use that folder in this example.

robintw: Fixed in d8fbe92

docs/getting_started.rst
robintw: Fixed in d8fbe92

robintw:
Issues fixed, and image added now.

PR #296, opened 2020-04-02
Load reference data when the DB to import is empty (#286)

### LINKED ISSUE #286: Suspected issue with empty database
I don't know if I've imagined this, but.

I'm importing `rep_Test1.rep` using the CLI importer.

It asks me for the classification on the data-file.  None are loaded.  So, I create a classification `public`

Shortly afterwards, it asks me for the classification for the platform.  Now the values offered are the ones from the reference `.csv` _plus_ the `public` one I entered above.

So, I think we're loading the reference data later in the process than we need to.

This only really relates to people _Test-driving_ Pepys, but it could hit a few people in the next few weeks.

BarisSari:
Fixes #286


pepys_import/core/store/data_store.py
```diff
@@ -1419,3 +1419,10 @@ def is_schema_created(self):
             print(f"Database tables are not found! (Hint: Did you initialise the DataStore?)")
             return False
         return True
+
+    def is_empty(self):
+        """Returns True if all tables are empty, False otherwise."""
```
IanMayo, line 1424: I think the above comment is a little ambitious :-D

We're not actually testing all tables.  Suggest something like:
`Returns True if sample table (Privacy) is empty, False otherwise`

pepys_import/database/default_data/Sensor Types.csv
```diff
@@ -1,3 +1,5 @@
 name
+GPS
+_GPS
```
IanMayo, line 3: I'm not sure about the history for us having a sensor called `_GPS`.  We shouldn't - we can just use `GPS`.

pepys_import/core/store/data_store.py
BarisSari: Yes, you're right. I was checking sample tables from measurement and metadata tables at the beginning. That's why it's written like that 🙂 I'm changing it.

pepys_import/database/default_data/Sensor Types.csv
BarisSari: We have it in NMEA and Replay Importer:
https://github.com/debrief/pepys-import/blob/2c8bcaf27ebf4ed91e2862b9609486af4d866f5f/importers/nmea_importer.py#L103

I'll change them as well.

pepys_import/core/store/data_store.py
BarisSari: Done:
https://github.com/debrief/pepys-import/pull/296/commits/ffded6ce619b0cefa1671471b641785425187d28

pepys_import/database/default_data/Sensor Types.csv
BarisSari: Done:
https://github.com/debrief/pepys-import/pull/296/commits/ffded6ce619b0cefa1671471b641785425187d28

PR #301, opened 2020-04-03
Fix deployment issue with installing distlib

robintw:
`distlib` was failing to install when the create_deployment.ps1 script was run. This was traced to an issue with installing this package which occurs only when it is installed from an embedded Python installation - as opposed to a standard Python installation.

To solve this, we have added a wheel file for distlib to the repo, and this PR updates the deployment script to install from this wheel file, and then delete it (so it doesn't get zipped up for the output).


robintw:
This PR has some incorrect reverted changes in it. Closing and opening a new one.

Issue #287, opened 2020-04-02
Error when dropping out of Import


IanMayo:
I'm playing with the CLI resolver.

I get an error sometimes when I cancel the import - but not all.

I cancelled at the point where I was setting the classification for a new platform:
```
processed_ctr = self.process_file(filename, current_path, data_store, processed_ctr)
  File "/Users/ian/git/pepys-import/pepys_import/file/file_processor.py", line 242, in process_file
    data_store, full_path, highlighted_file, datafile, change.change_id
  File "/Users/ian/git/pepys-import/pepys_import/file/importer.py", line 98, in load_this_file
    self._load_this_file(data_store, path, file_object, datafile, change_id)
  File "/Users/ian/git/pepys-import/pepys_import/file/importer.py", line 118, in _load_this_file
    self._load_this_line(data_store, line_number, line, datafile, change_id)
  File "/Users/ian/git/pepys-import/importers/replay_contact_importer.py", line 209, in _load_this_line
    privacy=privacy.name,
AttributeError: 'NoneType' object has no attribute 'name'
```

We should probably do some wider testing on how cancel performs.  Hmm, I think we will currently just drop back to the importer.  I _think_ we should probably propagate an exception back up the stack when `.) Cancel import` is pressed.

BarisSari:
Fixed in #307

PR #307, opened 2020-04-05
Hotfix/Solve the error when dropping out of import (#287)

BarisSari:
Solves #287


BarisSari:
@IanMayo , this PR is ready to review. I couldn't reproduce the error during the cancellation of Platform, Sensor or Datafile. However, I'd appreciate if you also try to produce the error before approving this PR.

Issue #278, opened 2020-04-01
Extractions should only be recorded if the parsing of the value succeeds


robintw:
At the moment, extractions are recorded from a token regardless whether the parsing of the value into a usable type (eg. parsing to a float in `convert_distance` or equivalent) actually succeeds.

We need to check whether it succeeds, and then only record if it does. Similarly, we also need to check that we're checking parsing succeeds before trying to write to the database - so we don't write invalid data.

robintw:
Fixed in #313

PR #313, opened 2020-04-06
Tidy up setting and recording of fields in importers

robintw:
This PR was originally going to be just fixing #278, but it turned out there were a number of other things that needed cleaning up in the importers.

Now we
 - Make sure we only set values in the State object if there hasn't been an error in parsing the value (eg. it isn't a valid float)
 - Check the parsing of timestamps, catching errors and reporting them as an error, rather than crashing with an exception
 - Added tests for parsing of invalid timestamps


importers/replay_comment_importer.py
```diff
@@ -92,7 +92,8 @@ def _load_this_line(self, data_store, line_number, line, datafile, change_id):
             comment_type = data_store.add_to_comment_types(comment_type, change_id)
 
             timestamp = parse_timestamp(date_token.text, time_token.text)
-            combine_tokens(date_token, time_token).record(self.name, "timestamp", timestamp)
+            if timestamp:
+                combine_tokens(date_token, time_token).record(self.name, "timestamp", timestamp)
```
BarisSari, line 96: Shouldn't we have an else statement which adds the parsing error for timestamp?

IanMayo:
@robintw  - it looks like we have some failing tests

PR #314, opened 2020-04-06
Add ambig_bearing field to Contacts table

robintw:
This PR adds the ambig_bearing field to the contacts table, with tests to check it is set correctly with units. It also adds extraction of ambig_bearing to the REP Contact Importer, and checks it is extracted properly in a test.


BarisSari:
@robintw , I think we should change `datastore.export_datafile()` method as well. Can you check the lines between 1364-1370?
```python
                ambiguous_bearing = None  # TODO: ambiguous_bearing.
                if ambiguous_bearing or contact.freq:
                    contact_rep_line.insert(0, ";SENSOR2:")


                    contact_rep_line.insert(
                        6, str(ambiguous_bearing.magnitude) if ambiguous_bearing else "NULL",
                    )
```

robintw:
Thanks @BarisSari - I missed the export. Now updated and tested.

@IanMayo I think this is ready to merge now.

robintw:
All conflicts and tests fixed now - I think the reduction in the coverage would be fixed by #329.

PR #320, opened 2020-04-07
Minor tidying to documentation

IanMayo:


robintw:
Actually, the description of what these shortcuts do needs changing too - I'll push a new commit later to update them.

robintw:
Actually the change I said I needed to make above now isn't needed @IanMayo - because you've deleted the usage section from the README anyway, so ignore my comment above.

IanMayo:
Yes, we already had a more thorough usage section in a dedicated page (same for installation)

PR #323, opened 2020-04-08
Added folder choice for exporting datafile

robintw:
Now asks the user which folder they want to export into. Uses prompt_toolkit which provides autocompletion for folder paths, starting with a default of ~ (user's home directory).

Tested and working on Windows and OS X.

### DIFF
--- a/pepys_admin/admin_cli.py
+++ b/pepys_admin/admin_cli.py
@@ -9,6 +9,7 @@ from prompt_toolkit.completion.filesystem import PathCompleter
 
 from pepys_admin.export_by_platform_cli import ExportByPlatformNameShell
 from pepys_admin.initialise_cli import InitialiseShell
+from pepys_admin.utils import get_default_export_folder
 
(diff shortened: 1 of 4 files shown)


robintw:
Yeah, the problem is that the current folder when the app is run via the Start Menu is the `bin` folder inside the Pepys installation. If we used that as the default folder then users would have to navigate out of a potentially quite deep folder hierarchy and get back to somewhere that is 'theirs' (as the installation will be on a shared drive).

We could do something like checking if the current folder is called `bin` and if so, defaulting to the user's home folder, and if not then defaulting to the current folder - but I'm not sure if that would just be a bit confusing?

IanMayo:
Checking the current folder seems an acceptable strategy.  I think that means that when run from the Start Menu it will default to the home folder, and when run from the command line it will us the current folder.

robintw:
This has now been tested on Windows and OS X and works: when you run from the Start Menu it will default to the home folder, when you run from the command-line it will default to the current folder.

PR #324, opened 2020-04-08
Add bin dir to PATH and make set_paths.bat work when called from anywhere

robintw:
create_shortcuts now adds the bin dir of the Pepys installation to the User's PATH variable (note: the User's PATH variable is different to the system PATH variable - and we can edit it without needing admin rights. Both the User and the System PATH are concatenated to get %PATH% which is actually used as the PATH.

Also, set_paths.bat didn't previously work if it was called outside of the directory that it is in - this now works when called from anywhere.

@IanMayo could you test this on one of your Windows VMs? It works fine for me, but it'd be good to test on a different Windows installation. You should be able to just update these two files in your installed version and then run create_shortcuts.bat. Once you've done that, open a normal Command Prompt and you should be able to just type any of the following and they will run successfully:

 - `pepys_import` (runs `pepys_import.bat`)
 - `pepys_admin` (runs `pepys_admin.bat`)
 - `set_paths` - should set the PATH, and then you can run `where python.exe` to check that the Pepys Python is the first in the list, and then `python -m pepys_import.import` should produce the same output as running `pepys_import` does.


robintw:
Docs are now updated for this, and `create_shortcuts` has been renamed to `install_pepys`.

I think this is ready to be merged once it's been tested by @IanMayo on his Windows VM.

docs/usage.rst
```diff
@@ -12,8 +12,8 @@ in two ways:
    file you want to import, and choose either :code:`Pepys Import` to import using the default settings
    or :code:`Pepys Import (no archive)` to import without archiving the imported files.
 
- - **Manually:** Run :code:`python -m pepys_import.import <options>` on the command-line, ensuring that the `python`
-   executable on your PATH is the one for which pepys-import has been installed.
+ - **Manually:** Run :code:`pepys_import <options>` on the command-line (this will call a batch file
```
IanMayo, line 15: Aah, @robintw - I do need to re-test this. When I typed pepys_import and tried to provide a path, I think the parameters got mangled.  Will try again.

robintw:
Your comment above is right. The `pepys_import.bat` file already provides the `--path` argument, as it is needed when calling via Send To - so running `pepys_import.bat --path blah` will give an error.

Two possible ways forward:

1. We tell people not to use the pepys_import.bat file (even though it will be on the path) and to just run `python -m pepys_import.cli` instead.

2. We create a separate `pepys_import_sendto.bat` file for use with Send To, and then have `pepys_import.bat` not pass the `--path` argument, so users can use it from the command-line.

I can see benefits to both - which would you prefer?

IanMayo:
Hello @robintw ,
I like option 2 more, please. It was quite "neat" being able to type `pepys_import --path data_folder` from the command line.

robintw:
Ok, I've implemented option 2 - so we now have separate `_sendto.bat` scripts that are run by Send To, and normal `pepys_import.bat` and `pepys_admin.bat` scripts that can be run directly from the command-line without mangling the parameters.

I've also updated things so that we check if the bin folder is already in the PATH before adding it again - so we don't end up with duplicate entries in PATH.

Issue #45, opened 2019-11-14
Bulk import performance


IanMayo:
An attached zip-file contains a high volume of data-files in REPLAY format.

We should use the sample data-set to investigate Pepys-Import performance bottlenecks, and see where performance can be improved.
[BulkAssetRuns.zip](https://github.com/debrief/pepys-import/files/3846385/BulkAssetRuns.zip)

IanMayo:
We have a one or two really large files in our sample data. I can get others.  It would be good if you could get some metrics on how long they take to import - maybe broken down by what’s happening at each stage (parse vs insert vs extraction highlighter).

Then as you address any bottlenecks we will see the performance improvement.

It's most important that we perform well in PG, but I am interested in SQLite performance too.

robintw:
Just FYI @BarisSari - you can turn off the extraction highlighter for a particular importer by calling `self.disable_recording()` in the `__init__` method of that importer. That will mean the extraction highlighter won't create its internal arrays, and it means that the `record` calls with become a no-op. Will be interesting to see how that speeds things up for large files.

BarisSari:
Thanks, @robintw . I'm trying to profile the import for using cProfile. It seems like that the highlighter takes at least 5-6% of the time. I'll import again with disabling recording as well. Because I'm sure that the highlighter calls many built-in functions as well. So, it is probably consuming more than 5-6 percent.

BarisSari:
First, I created a basic script,` local_file.py` and I profiled the following options using this script:
```python
from pepys_import.core.store.data_store import DataStore
from importers.replay_importer import ReplayImporter
from pepys_import.file.file_processor import FileProcessor

DATA_PATH = "/Users/baris/Documents/PycharmProjects/pepys-import/tests/sample_data/results2"  # bulk data folder

data_store = DataStore("", "", "", 0, "test.db", db_type="sqlite")
data_store.initialise()

processor = FileProcessor()
processor.register_importer(ReplayImporter())
processor.process(DATA_PATH, data_store, False)
```

- Run bulk-import without modifying the code
![pepys-import10](https://user-images.githubusercontent.com/41292272/78771343-b2cd2480-7998-11ea-9c40-d17446492ae2.png)

- Run bulk-import disabling the extraction highlighter
![pepys-import12_without_highlighter](https://user-images.githubusercontent.com/41292272/78771230-86b1a380-7998-11ea-8126-0d483a7ef638.png)

[...]

PR #328, opened 2020-04-10
Bulk import performance

BarisSari:
This PR has possible improvements to our import process. Yet, it might change once we decide to continue with a new approach.

### DIFF
--- a/pepys_import/core/store/common_db.py
+++ b/pepys_import/core/store/common_db.py
@@ -281,12 +281,17 @@ class DatafileMixin:
             for platform, objects in self.measurements[parser].items():
                 total_objects += len(objects)
                 print(f"Submitting measurements extracted by {parser}.")
-                # Commit files one by one
-                # for obj in tqdm(objects):
-                #     obj.submit(data_store, change_id)
-
                 # Bulk save table objects; state, etc.
                 data_store.session.bulk_save_objects(objects)
+
+                data_store.session.bulk_save_objects(
+                    [
+                        data_store.db_classes.Log(
+                            table=t.__tablename__, id=t.state_id, change_id=change_id
+                        )
+                        for t in objects
+                    ],
+                )
 
             extraction_log.append(f"{total_objects} measurements extracted by {parser}.")
         return extraction_log
--- a/pepys_import/core/store/common_db.py
+++ b/pepys_import/core/store/common_db.py
@@ -308,15 +308,6 @@ class SensorTypeMixin:
 
 
 class StateMixin:
-    def submit(self, data_store, change_id):
-        """Submit intermediate object to the DB"""
-        data_store.session.add(self)
-        data_store.session.flush()
-        data_store.session.expire(self, ["_location"])
-        # Log new State object creation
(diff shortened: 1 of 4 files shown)


robintw:
I've been looking into this today (while also working on improving the efficiency of the highlighter code). You mentioned in #45 that when using `bulk_save_objects`, we don't store Log entities - is there any reason why we can't write a separate function to deal with inserting all of the Log entries once we've done the bulk insert of the State/Contact/whatever objects? (Ideally creating them in a list and then submitting them with a `bulk_save_objects` call, or even a `bulk_insert_mappings` call. I imagine that the overhead of having to create an insert a `Log` entry for each new State is slowing everything down quite a bit.

Also, looking at https://docs.sqlalchemy.org/en/13/orm/persistence_techniques.html#bulk-operations, it seems that grouping the objects by type will further speed things up - so calling `bulk_save_objects` once for all the State objects, then once for all the Contact objects etc. Obviously to do that we'd have to loop through the measurements and extract them into different lists - so I don't know if that overhead would actually make things worse - probably worth trying though.

robintw:
Oh and I forgot to say - I _think_ that not expiring the Location field won't be too much of a problem. I actually wrote the Location class to be able to parse both the text representation and the full geographic-object representation of a Location - so it should work fine even if it's not expired (though expiring it is 'cleaner').

BarisSari:
> I've been looking into this today (while also working on improving the efficiency of the highlighter code). You mentioned in #45 that when using `bulk_save_objects`, we don't store Log entities - is there any reason why we can't write a separate function to deal with inserting all of the Log entries once we've done the bulk insert of the State/Contact/whatever objects? (Ideally creating them in a list and then submitting them with a `bulk_save_objects` call, or even a `bulk_insert_mappings` call. I imagine that the overhead of having to create an insert a `Log` entry for each new State is slowing everything down quite a bit.
> 

I'm thinking to use `bulk_insert_mappings` call for log entries. I just wanted to play with the bulk-import operations without bothering with Log entries. That's why I haven't added it yet.

> Also, looking at https://docs.sqlalchemy.org/en/13/orm/persistence_techniques.html#bulk-operations, it seems that grouping the objects by type will further speed things up - so calling `bulk_save_objects` once for all the State objects, then once for all the Contact objects etc. Obviously to do that we'd have to loop through the measurements and extract them into different lists - so I don't know if that overhead would actually make things worse - probably worth trying though.

Actually, the measurements are already categorized according to their parser. So, Replay Importer has State objects only, Replay Contact Importer has Contact objects etc. And we are saving the objects for each parsers' items using `bulk_save_objects`. I tried to collect all measurements and insert them with one command and it made the performance worse. Therefore, `data_store.session.bulk_save_objects(objects)` is inside of the for loop.

robintw:
That makes perfect sense @BarisSari

BarisSari:
After switching to bulk-import(implemented in this PR)+Robin's highlighter improvements, I imported the zipped rep files(in #45) one more time. Performance increased by 13-14%.

IanMayo:
@BarisSari - can you tell me why we've removed the mixins please?

PR #322, opened 2020-04-08
Add overall exception handlers to catch SQLAlchemy exceptions

robintw:
Adds exception handlers at a high level around both the pepys-import and pepys-admin CLIs, to catch a SQLAlchemy ProgrammingError, which occurs when the SQL doesn't make sense - eg. when fields are missing in the database.

Gives a sensible error message - but let me know if you want to reword it.

PR #329, opened 2020-04-10
Add CodeCov config file to tell it not to report coverage for the test files themselves

robintw:
The reason for this is that one of the other PRs (#322) was reporting low coverage, but that was because
there were exceptions handled in the tests that weren't being triggered (they were for if Postgres
wasn't installed) - but this counted as low coverage. We don't need to record coverage of the test code itself
so we can just tell CodeCov to ignore this for its reports.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+  - "tests/**/*"  # ignore all tests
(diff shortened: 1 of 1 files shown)


robintw:
This is now working - if you look [here](https://codecov.io/gh/debrief/pepys-import/compare/4f70a54590ca63d8e14bb9dec59e5507d52f76f2...878ce44b8df3fb19f9b9073421001db1746b99c6/changes) you'll see that all the test files are recorded as having no coverage reported.

PR #330, opened 2020-04-10
Improve highlighter performance

robintw:
This PR improves the performance of the extraction highlighter code, to enable faster import of data.

The time taken to run the `tests/highlighter_tests/test_combine.py::CombineTokenTests::test_combine_lines_on_multiple_lines_with_multiple` test has decreased by around 1/3 using these changes, and the time taken to run the bulk import test as described in #45 has decreased by around 15% - saving around 5 minutes for that bulk import.

The specific changes that have been implemented include:

 - Adding `__slots__` definitions to many of the classes that are instantiated a lot (eg. `Token`, `Char`, `SubToken`, `SingleUsage`). This is a Python implementation trick that means that arbitrary attributes can't be set on a class - only the ones defined in `__slots__` will be supported - and that means that the class can use a tuple to store its attributes rather than a dict, with an improvement in speed and memory usage.
 - Replacing lots of string concatenation with creating a list of strings and joining them - as this is known to improve performance in Python
 - Moving to f-strings where appropriate, as this is normally faster than using `+` to concatenate strings
 - Moving some lines of code outside of loops where possible
 - Using list comprehensions where possible, as they are faster and more memory efficient than for loops
 - Adding shortcut if statements to some functions to bypass the complicated processing if it is a simple case


BarisSari:
> Adding __slots__ definitions to many of the classes that are instantiated a lot (eg. Token, Char, SubToken, SingleUsage). This is a Python implementation trick that means that arbitrary attributes can't be set on a class - only the ones defined in __slots__ will be supported - and that means that the class can use a tuple to store its attributes rather than a dict, with an improvement in speed and memory usage.

@robintw , do you think that we should apply this change for the entire repository? Because 15% is a huge improvement and why not improve it more if we have already working updates. (I believe the rest of the improvements are already (more or less) applied in our repository)

robintw:
Good idea @BarisSari - but unfortunately I'm not sure there are many (any?) other places where they'd be both possible and useful. It's only really worth using __slots__ when you create a lot of objects of that class (at least hundreds, ideally thousands) and the only other objects that we create lots of are the SQLAlchemy entity classes (`State`, `Log` etc) - but according to [this](https://stackoverflow.com/questions/43049979/using-slots-with-sqlalchemy-model) it is difficult to use `__slots__` on a SQLAlchemy model, and even if you manage to use them then it doesn't have much benefit (as the base class doesn't define them).

BarisSari:
> Good idea @BarisSari - but unfortunately I'm not sure there are many (any?) other places where they'd be both possible and useful. It's only really worth using **slots** when you create a lot of objects of that class (at least hundreds, ideally thousands) and the only other objects that we create lots of are the SQLAlchemy entity classes (`State`, `Log` etc) - but according to [this](https://stackoverflow.com/questions/43049979/using-slots-with-sqlalchemy-model) it is difficult to use `__slots__` on a SQLAlchemy model, and even if you manage to use them then it doesn't have much benefit (as the base class doesn't define them).

Thanks for the explanation, @robintw .I thought to use it for the measurements but I didn't know that a workaround is necessary to use it in SQLAlchemy classes. I also can't think of any other classes which have more than hundreds of objects created. So, it seems to me that we don't need to use it anywhere else but highlighter.

PR #335, opened 2020-04-15
Add benchmarking

robintw:
Uses the `pytest-benchmark` library (remember to do `pip install -r requirements_dev.txt` before running this).

So far, there are two tests for the overall import process: one that should be fairly short, and one with Ian's long file.


robintw:
@IanMayo: I've started putting together some benchmarks in this PR. So far, there is one for a quick import of a small REP file, and one for the import of the large REP file that you sent me.

The framework we're using (`pytest-benchmark`) has various ways of making timings work well for functions that only take a relatively short time to run - so I thought I could probably come up with a few more short benchmarks for different parts of the process.

My main question to you now though, is about the threshold for failing a test if it takes too long. I've found that when running tests on my laptop, the results are very variable and sometimes the time taken can almost double (possibly due to my laptop increasing power sometimes when it detects a high workload, but not doing this all the time). On Travis, the results are far more stable - for example, one of the import tests takes around 86 seconds, and the max was 87 and min was 86. On my machine at home it took a lot longer (over two minutes) and was very variable.

Therefore, I wonder if we should only be actually checking the benchmarks against a time threshold when we're running the tests on Travis? We could still run the benchmarks and get the output (which is shown at the end of this comment) regardless where we're running, but then check the environment variables to see if we're running on Travis and then only do the check that it isn't taking too long on Travis (Travis sets particular environment variables so that scripts can do this sort of check - so the check would be robust).

What do you think?

```
------------------------------------------------------------------------------------------------ benchmark: 2 tests ------------------------------------------------------------------------------------------------
Name (time in ms)                             Min                    Max                   Mean            StdDev                 Median               IQR            Outliers     OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_single_rep_file_import_short        223.1573 (1.0)         230.2961 (1.0)         227.3126 (1.0)      2.3878 (inf)         228.0801 (1.0)      3.3759 (inf)           4;0  4.3992 (1.0)          10           1
test_single_rep_file_import_long      86,847.2029 (389.17)   86,847.2029 (377.11)   86,847.2029 (382.06)   0.0000 (1.0)      86,847.2029 (380.78)   0.0000 (1.0)           0;0  0.0115 (0.00)          1           1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

IanMayo:
Yes, it's quite reasonable to just run the tests on Travis.

robintw:
See wiki entry at https://github.com/debrief/pepys-import/wiki/Benchmarking-tests for more info on how these tests work.

Issue #200, opened 2020-03-11
Database access levels


IanMayo:
I've considered 3 Pepys database roles (note: these only apply to PostGres):
* pepys_admin: able to make schema changes (well, inject the schema)
* pepys_author: able to write data to the database
* pepys_viewer: able to retrieve data from the database

~~We should _probably_ manage these identifies in Pepys.~~ 

No, the IT Admin team can create 3 Pepys config files, and change the environmental variable for different users to point to different config files.

So, this doesn't require a code change - for information only.

PR #337, opened 2020-04-16
Database Migration (#158)

### LINKED ISSUE #158: Do we need to store the schema version in the database?
I'm creating an online PostGres hosting server, for us to use collaboratively.

I've realised that if the schema matures, our Python code may try to connect to a server with an older schema, and fail for unexpected reasons.

In the early stages, these failures should help us make the software more reliable.  But, in the longer term I think we need to understand the versioning.

Below are some thoughts on the issue that I captured last year.  But I now realise we store the schema in the SQLAlchemy code, rather than as SQL - which may prevent the below from being an option.

----

I'm no database expert - this is just a train of thought.

I think we need to _manage_ changes to the database.  

Well, right now we don't, we can just wipe it & recreate it.  

But, once the client has data in the database, that won't be an option. That is - once it's in _production_.

The client doesn't have a PostGres maintainer, or even a database expert.

I was wondering how we could support/manage the evolution of the database as requirements emerge.

Here's what I thought of.

We adopt the practice of capturing database changes on the end of the database creation script.  We use comment markers to separate versions.  
```
/* version:12 */
```

We'd parse the script file, using a regex that matches the above to split it into chunks.
We then include a "Version" table in the database. This is a tabular list of version numbers, and the date at which they were applied.

To update the database, a Python script connects to the database and then:

1. If the Version table isn't present, treat the version number as zero.
2. If there is a version table present retrieve the most recent version number.

With the version number, the composite script can be _split_ according to a regex that matches the above versioning comment. Then:

1. Parse the SQL script, applying all changes _after_ that version number. When a block of lines (that form a version) are applied, insert that version number into the versions table

So, the above process can either create a new, empty database, or non-destructively transform a database structure to a new version.

Aah, some of the above are covered here:
https://enterprisecraftsmanship.com/2015/08/10/database-versioning-best-practices/

Aah, and here's an Open Source tool that supports the migrating schemas:
http://www.liquibase.org/index.html

The above script would be within our version control system - so source changes and database schema changes are handled together.

----
We could include the above process in our library. After our tool has successfully connected to a db, it checks for the versions table.

If it's missing, the tool asks the user if the user wants the database to be created.

If it has got the version number, and can see that number is less than the current one, it asks the user if the user wants to update the database.

Aah, hmm. Applying the updates needs a lot more permissions than an account that just inserts data.

So, if the tool has established that the database needs updating, and the user wishes to run the update, but gets "access denied" - it should prompt the user to use an account/password with higher permissions.

----

BarisSari:
Fixes #158 . Detailed information can be found in the `migrations` folder: https://github.com/debrief/pepys-import/tree/introduce_alembic/migrations

**Note:** In order to use `is_schema_created()` without a DataStore object, it's moved into `datastore_utils`. Additionally, 2 functions implemented for Spatial table creation for PostgreSQL and SQLite.


BarisSari:
As you all know, I'm trying functionalities of Alembic in the last days. Unfortunately, I've faced many problems. Therefore, it hasn't successfully integrated to our repository yet. I've tried to find a solution for the following issue we discussed earlier:
-  Capturing database changes. Because the client can't wipe and recreate the database once its in production. So, we should somehow keep track of the client's DB version and the current version of the project.

I'm a bit familiar with the migration strategy because I've used it in many different Django projects. It has the following logic: Each package of the Django project has `migrations` folder. With the built-in commands, it is possible to collect the DB models automatically and create migration files. They are python modules but it's possible to convert them to SQL scripts to check. When migration files are created, another command migrates the updates to the DB and saves the insert/update to the migrations table to keep track of it. The good thing about Django's migration is it is possible to use `migration=False` flag in your models. So, if it is set, you aren't allowed to migrate your DB model changes to the DB. But it's possible to do vice versa.

We more or less need the same strategy. When we run pepys_admin or pepys_import, we can check the migrations and warn the user if its DB has missing migrations. We can also try to run the migration automatically after the user approval for the migration.

Alembic has used by many popular frameworks (https://github.com/miguelgrinberg/Flask-Migrate) 
as well. But here are the problems I've faced so far:
- It can automatically generate migration scripts just as Django, however, this auto-generation leads to two main problems. The first one is, it detects Spatial fields and add them to the migration script but it doesn't import `Geoalchemy2` library. I've manually added the import for now but I think we can include it to each migration script. The second problem is a hard one because it tries to drop Spatial tables from the DB when the automatic migration command is run again. An example is as follows:
```python
op.drop_table("idx_Media_location_parent")
op.drop_table("idx_Contacts_location")
op.drop_table("idx_Geometries_geometry_rowid")
```
I'm afraid, there isn't a decent solution yet for this problem. For example, this one suggests creating indexes manually: https://github.com/geoalchemy/geoalchemy2/issues/137
There is another solution: https://gist.github.com/utek/6163250
However, when I modify and use it, it doesn't create spatial tables when the DB is empty this time because they are excluded from Alembic. I believe that there might be another solution, which is to use SQLAlchemy's `create_all()` method for the first migration and stamp the DB with this migration at the beginning. I'm trying this approach for now. I'll update this message when it works/fails.

Another problem is that the developers might have two different databases, SQLite and Postgres.  It means that we should have two different base migration scripts. Alembic offers this: https://alembic.sqlalchemy.org/en/latest/branches.html#working-with-multiple-bases
However, autogenerate command becomes a problem again. Even though I followed the tutorials and searched for it thoroughly, I couldn't autogenerate changes for different DBs. It is possible to create manual migrations with the following commands:
          `alembic revision -m "migration-1" --head sqlite@head`

It creates a new file in the migration folder for SQLite, and you can manually enter your changes. I don't believe that it is what we want.

By the way, Alembic offers: downgrading migrations, creating SQL scripts from migration files(https://alembic.sqlalchemy.org/en/latest/offline.html#generating-sql-scripts-a-k-a-offline-mode), showing the history of migrations, showing the current head of the DB, etc. So, it has many useful functionalities. However, it won't mean anything unless I adapt it according to our needs first. 

Since one week has passed, I wanted to update all of you with my findings/problems. Any feedback&thoughts are welcome, @robintw and @IanMayo . Honestly, I'm a bit concerned because of the problems above 🙂 

**Note:** As @IanMayo suggested on Slack, it is possible to have pure SQL scripts for migrations. However, I don't think that it will be useful for the client or the developers. Because they should figure out the changes from SQL Scripts and apply it to the DB models in `sqlite_db.py `and `postgres_db.py`. Instead of this, I think we should have migrations scripts (in Python) that are created automatically and can be convertible to SQL scripts if necessary(it's what I'm trying to do by now). On the other hand, I don't think that Alembic-SQLAlchemy offers a migration flag for forbidding database changes from models.

IanMayo:
@BarisSari - are the new documents included in the online help? I've just been to the `readthedocs` preview, but can't see the new docs.

BarisSari:
> @BarisSari - are the new documents included in the online help? I've just been to the `readthedocs` preview, but can't see the new docs.
> 
> https://pepys-import--337.org.readthedocs.build/en/337/

No, I haven't included it yet. But for your and @robintw's review,  I'll put it now and share it with you.

IanMayo:
@robintw - you declared from the start that you weren't a database expert.  I'd still appreciate your trial/review of this PR, please.

pepys_import/core/store/data_store.py
```diff
@@ -6,20 +6,26 @@
...
+from paths import PEPYS_IMPORT_DIRECTORY, ROOT_DIRECTORY
 from pepys_import import __version__
 from pepys_import.core.formats import unit_registry
 from pepys_import.core.formats.location import Location
 from pepys_import.core.store import constants
 from pepys_import.resolvers.default_resolver import DefaultResolver
 from pepys_import.utils.branding_util import show_software_meta_info, show_welcome_banner
-from pepys_import.utils.data_store_utils import import_from_csv
+from pepys_import.utils.data_store_utils import (
+    create_spatialite_tables_for_postgres,
```
robintw, line 25: Minor issue, but `spatialite` is the name of the spatial extension for SQLite, and `postgis` is the name of the spatial extension for postgres. Maybe just rename these to `create_spatial_tables_for_X`?

migrations/env.py
```diff
@@ -0,0 +1,197 @@
...
+from pepys_import.utils.geoalchemy_utils import load_spatialite
+
+DIR_PATH = os.path.dirname(__file__)
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+db_type = config.attributes.get("db_type", DB_TYPE)
+
+driver = "sqlite"
```
robintw, line 29: Do we need to set the driver here, if we're then setting it in the if statements below?

pepys_import/core/store/data_store.py
BarisSari: Thanks for pointing out the typo. It's corrected:
https://github.com/debrief/pepys-import/pull/337/commits/77bfbb14b2fd4282d5498e75ff023337890ce8b2

migrations/env.py
BarisSari: No, we don't need it. It's removed:
https://github.com/debrief/pepys-import/pull/337/commits/77bfbb14b2fd4282d5498e75ff023337890ce8b2

robintw:
I've been testing running alembic in various configurations on my machine - comments and questions below:

1. When I added a new field to the State model in sqlite_db.py and ran `alembic revision`, I got a migration file containing the following:

```
def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    with op.batch_alter_table("Geometries", schema=None) as batch_op:
        batch_op.alter_column("geometry", existing_type=sa.NUMERIC(), nullable=False)

    with op.batch_alter_table("States", schema=None) as batch_op:
        batch_op.add_column(sa.Column("new_field", sa.Integer(), nullable=True))

    # ### end Alembic commands ###
```

Why is there an alteration to the geometries table here too, when I only added the `new_field` column?

2. I followed the instructions at https://pepys-import--337.org.readthedocs.build/en/337/database_migration.html to stamp my SQLite database manually, as I already had data in it. I then created the revision as shown above to add the `new_field` column to the `States` table. That was revision `53af5060adc6`. When I run `alembic upgrade 53af5060adc6 --sql` I get some strange output and an error.

The error is:

```
INFO  [alembic.runtime.migration] Running upgrade bcff0ccb4fbd -> 7df9dcbd47e7, base2
-- Running upgrade bcff0ccb4fbd -> 7df9dcbd47e7

  File "/Users/robin/anaconda3/envs/pepys/bin/alembic", line 8, in <module>
    sys.exit(main())
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/config.py", line 577, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/config.py", line 571, in main
    self.run_cmd(cfg, options)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/config.py", line 551, in run_cmd
    **dict((k, getattr(options, k, None)) for k in kwarg)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/command.py", line 298, in upgrade
    script.run_env()
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/base.py", line 489, in run_env
    util.load_python_file(self.dir, "env.py")
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/util/pyfiles.py", line 98, in load_python_file
    module = load_module_py(module_id, path)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/util/compat.py", line 184, in load_module_py
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "migrations/env.py", line 185, in <module>
    run_migrations_offline()
  File "migrations/env.py", line 124, in run_migrations_offline
    context.run_migrations()
  File "<string>", line 8, in run_migrations
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/runtime/environment.py", line 846, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/runtime/migration.py", line 520, in run_migrations
    step.migration_fn(**kw)
  File "/Users/robin/Documents/IanMayo/pepys-import/migrations/sqlite_versions/2020-04-23_7df9dcbd47e7_base2.py", line 22, in upgrade
    batch_op.alter_column("geometry", existing_type=sa.NUMERIC(), nullable=False)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/operations/base.py", line 354, in batch_alter_table
    impl.flush()
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/operations/batch.py", line 101, in flush
    **self.reflect_kwargs
  File "<string>", line 2, in __new__
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/util/deprecations.py", line 128, in warned
    return fn(*args, **kwargs)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/sql/schema.py", line 506, in __new__
    metadata._remove_table(name, schema)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/util/langhelpers.py", line 68, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 153, in reraise
    raise value
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/sql/schema.py", line 501, in __new__
    table._init(name, metadata, *args, **kw)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/sql/schema.py", line 595, in _init
    resolve_fks=resolve_fks,
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/sqlalchemy/sql/schema.py", line 613, in _autoload
    autoload_with.run_callable(
AttributeError: 'MockConnection' object has no attribute 'run_callable'
```

The output above that is the SQL for creating all the tables - Contacts, State, CommentTypes etc - everything that was already created before I ran alembic for the first time. I would have thought that would have been dealt with by stamping the version?

(More to come, but thought I'd post this now so you could have a look at it now)

robintw:
Some more thoughts from me after playing with the alembic migrations a bit:

1. Do we _really_ want to automatically run the migrations on Datastore.initialise(). This would silently change the schema of the user's database, without asking them or telling them or anything. I think at the very least we should print a prominent notice saying that we're upgrading from revision X to revision Y. Possibly we should ask the user for confirmation before we do the upgrade. Potentially if the database has a lot of data in, then the schema change could take some time (I guess? I don't really know) - which might affect other users connecting to the database.

2. Linked to the above, it seems from https://github.com/debrief/pepys-import/issues/200 that the normal users who will be running pepys_import won't have the ability to do the upgrade to the schema anyway (assuming they're using a common Postgres DB, rather than a local SQLite db) - and therefore they will just get an error when there is a schema mismatch anyway. Might it be better to add an option to pepys_admin to do the schema upgrade - then we know it is done at an appropriate time by a user with admin permissions. Presumably we'll only need to do schema upgrades when we send them a new version of Pepys - so we could add a step to the installation process to ask them to run pepys_admin and do the upgrade. Actually, just had an idea - we could do the upgrade automatically if it's a sqlite database (although maybe still ask for confirmation?) and just give an error and tell them to get an admin to do the upgrade if it's a Postgres database? Also, we should make the error message that appears if there's a problem with the migration more explicit, and say that it may be that the user doesn't have permissions to perform the update (actually, we should test this with different users in a Postgres install and check if a particular exception is raised when we don't have permissions, and see if we can catch that separately). 
    
 3. Can I just check that I've got the process that we'd be following for this right in my head (I'm relatively new to tools like alembic). I'm trying to work out how this works with deployments to the client, and where the migrations go in this process.
        a. During development, we realise we need a new field in the database (or some other schema change)
        b. We change the schema in sqlite_db.py and postgres_db.py
        c. We run `alembic revision`
        d. We check the autogenerated migration, and then commit the migration file.
        e. We do this again for the other database (Postgres or SQLite)
        f. We check and commit the other migration
        g. We release a new version of Pepys
        h. At the client's site, they install the new version of Pepys and then either let a database upgrade happen automatically (see 1 & 2 above) or do it manually via pepys_admin or the alembic command-line tool. Presumably as long as the client's database has an alembic version 'stamped' in to it, alembic will be able to work out which migrations to run to get the database up to date
        
Is that all right? Am I correct that we're not proposing to run `alembic revision` at the client's site, as we should be able to create all the revisions we need locally and then just push them out in the deployment process?
        
4. If we want access to the alembic command-line tool in the Windows deployment, then we'll need to modify the set_paths.bat script to add the Python 'scripts' directory to the PATH. This will make a lot of other Python command-line scripts available on the PATH, which hopefully won't be a problem (as they will disappear from the PATH when you close the Command Window that you ran set_paths.bat in). If we're not expecting the clients to run the alembic command directly then we could leave this, as it would be possible to run the command-line tool in an emergency - you'd just have to use the full path.

IanMayo:
@BarisSari - the tests are failing in Travis, on `test_spatial_data.py`.  They fail on that test locally when run in bulk (the whole `tests` folder), but not when that test is run on its own.

Here's the call stack: https://travis-ci.com/github/debrief/pepys-import/builds/161970156#L1024

BarisSari:
> @BarisSari - the tests are failing in Travis, on `test_spatial_data.py`. They fail on that test locally when run in bulk (the whole `tests` folder), but not when that test is run on its own.
> 
> Here's the call stack: https://travis-ci.com/github/debrief/pepys-import/builds/161970156#L1024

Yes, @IanMayo . I'm aware of this issue, I'm trying to solve it.

BarisSari:
>  I've been testing running alembic in various configurations on my machine - comments and questions below:
>     1.When I added a new field to the State model in sqlite_db.py and ran `alembic revision`, I got a migration file containing the following:
> ```
> def upgrade():
>     # ### commands auto generated by Alembic - please adjust! ###
>     with op.batch_alter_table("Geometries", schema=None) as batch_op:
>         batch_op.alter_column("geometry", existing_type=sa.NUMERIC(), nullable=False)
> 
>     with op.batch_alter_table("States", schema=None) as batch_op:
>         batch_op.add_column(sa.Column("new_field", sa.Integer(), nullable=True))
>     # ### end Alembic commands ###
> ```
> 
> Why is there an alteration to the geometries table here too, when I only added the `new_field` column?
>     2.I followed the instructions at https://pepys-import--337.org.readthedocs.build/en/337/database_migration.html to stamp my SQLite database manually, as I already had data in it. I then created the revision as shown above to add the `new_field` column to the `States` table. That was revision `53af5060adc6`. When I run `alembic upgrade 53af5060adc6 --sql` I get some strange output and an error.
> 
>  ...

Hi @robintw , I figured out the problem you have. I don’t know how but when we use `Base.metadata.create_all()` or Alembic, it creates the `geometry` field of the Geometry Table nullable for SQLite. Therefore, when Alembic tries to detect new changes, it says that this field changed and it can’t be nullable. Therefore, it adds an `alter_column` clause. I already added a migration script for this issue but I didn’t expect you to face this (I was thinking that `Base.metadata.create_al()` creates this column correctly). When I stamped the DB with the first migration and ran the DB upgrade, it worked. Therefore, I changed the Data Migration documentation. Please look at this section: https://pepys-import--337.org.readthedocs.build/en/337/database_migration.html#instructions

The second migration file which changes the `geometry` field was causing a failure when you tried to produce a SQL script (By default, if you don't give a start point, `--sql` creates SQL scripts for all migration scripts.). I explained the usage of `--sql` flag option more clearly now. You can check it here: https://pepys-import--337.org.readthedocs.build/en/337/database_migration.html#how-to-use-it-for-developers

I followed Alembic's cookbook to solve this issue: https://alembic.sqlalchemy.org/en/latest/batch.html#batch-offline-mode
> In the preceding sections, we’ve seen how much of an emphasis the “move and copy” process has on using reflection in order to know the structure of the table that is to be copied. This means that in the typical case, “online” mode, where a live database connection is present so that Operations.batch_alter_table() can reflect the table from the database, is required; the --sql flag cannot be used without extra steps.

> To support offline mode, the system must work without table reflection present, which means the full table as it intends to be created must be passed to Operations.batch_alter_table() using copy_from:
> ```python
> meta = MetaData()
> some_table = Table(
>     'some_table', meta,
>     Column('id', Integer, primary_key=True),
>     Column('bar', String(50))
> )
> 
> with op.batch_alter_table("some_table", copy_from=some_table) as batch_op:
>     batch_op.add_column(Column('foo', Integer))
>     batch_op.drop_column('bar')
> ```

BarisSari:
> Some more thoughts from me after playing with the alembic migrations a bit:
> 
>     1. Do we _really_ want to automatically run the migrations on Datastore.initialise(). This would silently change the schema of the user's database, without asking them or telling them or anything. I think at the very least we should print a prominent notice saying that we're upgrading from revision X to revision Y. Possibly we should ask the user for confirmation before we do the upgrade. Potentially if the database has a lot of data in, then the schema change could take some time (I guess? I don't really know) - which might affect other users connecting to the database.

>     2. Linked to the above, it seems from #200 that the normal users who will be running pepys_import won't have the ability to do the upgrade to the schema anyway (assuming they're using a common Postgres DB, rather than a local SQLite db) - and therefore they will just get an error when there is a schema mismatch anyway. Might it be better to add an option to pepys_admin to do the schema upgrade - then we know it is done at an appropriate time by a user with admin permissions. Presumably we'll only need to do schema upgrades when we send them a new version of Pepys - so we could add a step to the installation process to ask them to run pepys_admin and do the upgrade. Actually, just had an idea - we could do the upgrade automatically if it's a sqlite database (although maybe still ask for confirmation?) and just give an error and tell them to get an admin to do the upgrade if it's a Postgres database? Also, we should make the error message that appears if there's a problem with the migration more explicit, and say that it may be that the user doesn't have permissions to perform the update (actually, we should test this with different users in a Postgres install and check if a particular exception is raised when we don't have permissions, and see if we can catch that separately).


Actually, Alembic prints the applied migration which you can see at the beginning of pepys-import. For example:

> ```bash
>  @@@@@ @@..@@@@      | ___ \                    (_)                          | |  
>   @@@     @@@@@@     | |_/ /__ _ __  _   _ ___   _ _ __ ___  _ __   ___  _ __| |_ 
>   @@@     @@@@@@@@   |  __/ _ \ '_ \| | | / __| | | '_ ` _ \| '_ \ / _ \| '__| __|
>  @@@@.     @@@@@@@   | | |  __/ |_) | |_| \__ \ | | | | | | | |_) | (_) | |  | |_ 
>  @@ #@@ @   ...@@@@@ \_|  \___| .__/ \__, |___/ |_|_| |_| |_| .__/ \___/|_|   \__|
>  @@   &        @@@@@            | |     __/ | ______          | |                   
> Software Version :  0.0.10 
> 
> 
> Database Type :  sqlite
> Database Name :  :memory:
> Database Host :  
> -------------------------------------------------------------
> Database tables are not found! (Hint: Did you initialise the DataStore?)
> INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
> INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
> INFO  [alembic.runtime.migration] Running upgrade  -> bcff0ccb4fbd, base
> INFO  [alembic.runtime.migration] Running upgrade bcff0ccb4fbd -> 7df9dcbd47e7, base2
> ```

When something is changed in a table that has lots of data, of course, it will take some time to copy the values from the old table to the new one. However, the user will know that migration is working(it will be printed). 

I'm not quite sure what would be the best approach: changing silently each time they run `pepys_import` or `pepys_admin` or asking the migration to the user and applying it after their approval. I believe that @IanMayo knows the users better than us, maybe he might suggest according to their needs.

>     3. Can I just check that I've got the process that we'd be following for this right in my head (I'm relatively new to tools like alembic). I'm trying to work out how this works with deployments to the client, and where the migrations go in this process.
>        a. During development, we realise we need a new field in the database (or some other schema change)
>        b. We change the schema in sqlite_db.py and postgres_db.py
>        c. We run `alembic revision`
>        d. We check the autogenerated migration, and then commit the migration file.
>        e. We do this again for the other database (Postgres or SQLite)
>        f. We check and commit the other migration
>        g. We release a new version of Pepys
>        h. At the client's site, they install the new version of Pepys and then either let a database upgrade happen automatically (see 1 & 2 above) or do it manually via pepys_admin or the alembic command-line tool. Presumably as long as the client's database has an alembic version 'stamped' in to it, alembic will be able to work out which migrations to run to get the database up to date
> 
> Is that all right? Am I correct that we're not proposing to run `alembic revision` at the client's site, as we should be able to create all the revisions we need locally and then just push them out in the deployment process?

Yes, that's the process. We don't need to propose to run `alembic revision` at the client's side. It's a command for the developers. During the deployment process, the database might be upgraded automatically or manually with one basic command: `alembic upgrade head`. Actually there is one other option for the manually upgrade. Alembic has an API which you can call from your Python scripts: https://alembic.sqlalchemy.org/en/latest/api/commands.html
So, we might have a basic Python script, migrate.py, which might use `alembic.command.upgrade()`

>     1. If we want access to the alembic command-line tool in the Windows deployment, then we'll need to modify the set_paths.bat script to add the Python 'scripts' directory to the PATH. This will make a lot of other Python command-line scripts available on the PATH, which hopefully won't be a problem (as they will disappear from the PATH when you close the Command Window that you ran set_paths.bat in). If we're not expecting the clients to run the alembic command directly then we could leave this, as it would be possible to run the command-line tool in an emergency - you'd just have to use the full path.

I think we should decide this after we agree on the upgrade process.

IanMayo:
Hello guys @here

I think we've gained an understanding of what's achievable as @BarisSari has been working through this. 

I do expect the clients to eventually have a range of login accounts, with different privileges, most of which won't be able to do a database upgrade.

So, it makes sense for the upgrade to happen in PepysAdmin.

But, hmm, show PepysImport _drop-out_ if the database isn't at the right level.  I wouldn't want a user to get stuck.  Aaah, the database will only ever by at the wrong level on a new release.  So, as part of the release cycle we can get a suitably privileged user (such as the IT admins that are installing the release) to go into PepysAdmin, and do the database upgrade.  Until this is done, PepysImport won't work, nor will any PepysAdmin operations (other than 'apply upgrade').

How does this sound?

BarisSari:
> Hello guys @here
> 
> I think we've gained an understanding of what's achievable as @BarisSari has been working through this.
> 
> I do expect the clients to eventually have a range of login accounts, with different privileges, most of which won't be able to do a database upgrade.
> 
> So, it makes sense for the upgrade to happen in PepysAdmin.
> 
> But, hmm, show PepysImport _drop-out_ if the database isn't at the right level. I wouldn't want a user to get stuck. Aaah, the database will only ever by at the wrong level on a new release. So, as part of the release cycle we can get a suitably privileged user (such as the IT admins that are installing the release) to go into PepysAdmin, and do the database upgrade. Until this is done, PepysImport won't work, nor will any PepysAdmin operations (other than 'apply upgrade').
> 
> How does this sound?

Hello, @IanMayo . It sounds okay to me. I'll change the `DataStore.initialise()` and add `upgrade()` functionality to the Pepys-admin.

BarisSari:
This PR is ready to review and merge, @IanMayo .

migrations/README.rst
```diff
@@ -0,0 +1,252 @@
...
+==================
+
+Pepys-import repository uses `Alembic <https://github.com/alembic/alembic>`_ for data migration. Alembic is a database migration tool which is maintained by SQLAlchemy.
+It is possible to apply the existing migrations to the users' database via :code:`pepys_admin`.
+Don't forget to check your configuration file before running :code:`pepys_import` or :code:`pepys_admin`.
+
+Installation
+------------
+Alembic is in the requirements of the project. However, there is one post-installation step to run it without any problem. You should install the pepys-import project in an editable mode. The commands must be run are as follows:
+
```
IanMayo, line 12: @BarisSari - for which groups of users are these steps necessary?  Just devs that make database schema changes?

BarisSari: Yes, it's for the developers to run `alembic` commands

IanMayo: Ok.  In that case I think we need to "switch together" the developer guidance.  Could you come up with a tidy way of including this guidance in the `Developer Setup` page?

BarisSari: Sure @IanMayo . I moved this step to the Developer Guide's "Create Python virtual environment and install dependencies" section:

https://pepys-import--337.org.readthedocs.build/en/337/developer_guide.html#create-python-virtual-environment-and-install-dependencies

BarisSari:
> For some reason Github hadn't notified me about the comments on this PR - thanks for the detailed responses @BarisSari and sorry for not responding sooner.
> 
> Your explanations make perfect sense, and doing the migration in pepys-admin makes sense.
> 
> All the code seems ok - just one minor issue. If we run an import with a new SQLite database (eg by passing `--db ":memory:"` then we get a message `Database tables are not found! (Hint: Did you initialise the DataStore?)` printed, following by the import going ahead fine. This is coming from the `is_schema_created` function - do you think we need to alter this a bit so that we point out that the tables will automatically be created? At the moment it looks like an error to the end-user.
> 
> The problem is that the tables won't automatically be created in some situations - eg. when we call `is_schema_created` when doing a Status from pepys-admin, it won't create the tables (which is correct) - so we can't just update the message we print.
> 
> What do you think is best to do?

You're welcome Robin. You're right that it looks like that there is an error but it continues to work.  Maybe we might print in Alembic's `env.py` that the tables are going to be created. So, the output will be as follows:

```bash
(pepys-admin) 5
Database tables are not found! (Hint: Did you initialise the DataStore?)
Database tables are going to be created by Alembic.
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
INFO  [alembic.runtime.migration] Running upgrade  -> bcff0ccb4fbd, base
INFO  [alembic.runtime.migration] Running upgrade bcff0ccb4fbd -> 7df9dcbd47e7, base2
```

When `pepys_import` is run with `--db ":memory:"` option, the output will be as follows:

```bash
   @@@@ @@@@@        ______                      _                            _   
 @@@@@ @@..@@@@      | ___ \                    (_)                          | |  
  @@@     @@@@@@     | |_/ /__ _ __  _   _ ___   _ _ __ ___  _ __   ___  _ __| |_ 
  @@@     @@@@@@@@   |  __/ _ \ '_ \| | | / __| | | '_ ` _ \| '_ \ / _ \| '__| __|
 @@@@.     @@@@@@@   | | |  __/ |_) | |_| \__ \ | | | | | | | |_) | (_) | |  | |_ 
 @@ #@@ @   ...@@@@@ \_|  \___| .__/ \__, |___/ |_|_| |_| |_| .__/ \___/|_|   \__|
 @@   &        @@@@@            | |     __/ | ______          | |                   
Software Version :  0.0.10 


Database Type :  sqlite
Database Name :  :memory:
Database Host :  
-------------------------------------------------------------
Database tables are not found! (Hint: Did you initialise the DataStore?)
Database tables were created by DataStore's initialisation.
```

Do you think that it's a decent solution, @robintw ?

robintw:
Yes, that sounds good @BarisSari.

IanMayo:
@BarisSari - could of conflicts to resolve, please.

BarisSari:
> @BarisSari - could of conflicts to resolve, please.

Resolved, @IanMayo .

robintw:
Sorry for the delay in approving this. I wanted to test this on Windows, but am having issues getting alembic installed in our embedded Python environment that we use for deployment.

There are various potential ways to solve this (some nicer than others) and I'm investigating them at the moment. Hopefully I can add a fix to this PR so that everything works ok.

robintw:
I've fixed the issue now with installing alembic on the embedded Python version - see #344.

While debugging that, I've come up with a couple of minor additions that I think would be useful for this PR @BarisSari :

1. Could we have an output of the current database version when we run Status in Pepys Admin? I can see that being useful for debugging (I spoke to Ian and we decided not to put the alembic command on the PATH for deployment, so the clients wouldn't easily be able to run any alembic commands from the command-line).

2. Is it possible to add a bit more detail to the migrate option in Pepys Admin? At the moment it just prints the raw output from the alembic command - I know it's hard to modify that, but if a bit of explanation could be printed above or below that. Something like "Alembic migration command running, see output below. If no upgrade mentioned in output then no migration was needed"? Just gives the user a bit more confidence that something has happened, and everything is ok to go forward with.

BarisSari:
> I've fixed the issue now with installing alembic on the embedded Python version - see #344.
> 
> While debugging that, I've come up with a couple of minor additions that I think would be useful for this PR @BarisSari :
> 
>     1. Could we have an output of the current database version when we run Status in Pepys Admin? I can see that being useful for debugging (I spoke to Ian and we decided not to put the alembic command on the PATH for deployment, so the clients wouldn't easily be able to run any alembic commands from the command-line).

We might use `alembic.command.current() `. When it's verbose, it prints the following. I believe that it gives enough information about the version of the current database.

```bash
Current State of the Database
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
Current revision(s) for sqlite+pysqlite://:@:0/pepys_test.sqlite:
Rev: 7df9dcbd47e7 (head)
Parent: bcff0ccb4fbd
Path: /Users/baris/Documents/PycharmProjects/pepys-import/migrations/sqlite_versions/2020-04-25_7df9dcbd47e7_base2.py

    base2
    
    Revision ID: 7df9dcbd47e7
    Revises: bcff0ccb4fbd
    Create Date: 2020-04-23 12:26:51.084226

-------------------------------------------------------------
```

>     2. Is it possible to add a bit more detail to the migrate option in Pepys Admin? At the moment it just prints the raw output from the alembic command - I know it's hard to modify that, but if a bit of explanation could be printed above or below that. Something like "Alembic migration command running, see output below. If no upgrade mentioned in output then no migration was needed"? Just gives the user a bit more confidence that something has happened, and everything is ok to go forward with.

You're right that it's hard to modify the output of the alembic commands. Because they are logged and printed automatically. I added your suggestion ("Alembic migration command running, see output below.") to the above. However, I'm afraid we can't know whether the DB is upgraded or not unless we dive into the log of Alembic. I'm not sure whether it would be a good thing to parse the log and try to understand whether an upgrade is run or not. Do you think that we should print something if no migrations applied, @robintw ?

robintw:
Hi @BarisSari.

The output of `alembic.command.current()` looks fine for the status in Pepys Admin, and the print just saying that the migration is running is fine too. I'm very happy to approve this now - thanks for doing all this!

PR #340, opened 2020-04-21
Add support for archive folder being on SMB share

### LINKED ISSUE #189: Allow archive folder identity to be loaded
Since the archive folder contains the master copy of the import files, the users would like it to be protected from modification/delete.

They have created a dedicated username/account.

But we need to give Pepys access to a config file containing that username/password.

But, beyond that, the IT team would rather they weren't stored as plain text.

I welcome suggestions for how we implement this. We could have an encryption key in our app, but it's Open Source.

Hmm, maybe some obfuscation.

robintw:
Adds support for the archive folder being located on a Windows (SMB) share. See the updated `configuration.rst` file in the PR for details - but basically just set the archive path in the config file to something like `\\SERVER\share\path\to\folder`, along with the right username and password, and it should all work.

This is implemented by wrapper functions that are used whenever we write to the archive folder (principally in the `FileProcessor`, but also in the `HighlightedFile` class) - and these dispatch to two different implementations, depending whether the archive is on a SMB share or not. All the SMB functions are wrapped in `try...except` blocks to deal with authentication/network issues.

Tests are written on the assumption that the `smbprotocol` library behaves as advertised - and thus the tests only check that the right SMB functions are called. This also means we don't need a SMB server available to run the tests, which makes life a lot easier.

Fixes #189

### DIFF
--- a/pepys_import/file/file_processor.py
+++ b/pepys_import/file/file_processor.py
@@ -4,11 +4,11 @@ import os
 from datetime import datetime
 from getpass import getuser
 
-import pepys_import.file.smb_and_local_file_operations as smblocal
 from config import ARCHIVE_PATH, LOCAL_PARSERS
 from paths import IMPORTERS_DIRECTORY
 from pepys_import.core.store.data_store import DataStore
 from pepys_import.core.store.table_summary import TableSummary, TableSummarySet
+from pepys_import.file import smb_and_local_file_operations as smblocal
 from pepys_import.file.highlighter.highlighter import HighlightedFile
 from pepys_import.file.importer import Importer
 from pepys_import.utils.datafile_utils import hash_file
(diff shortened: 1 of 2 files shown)


robintw:
This is now ready for review. The reason for the low coverage for the patch is that we're not testing against a live SMB server, so we can't test that we get particular exceptions raised under specific circumstances (eg. a `SMBAuthenticationError` when we give the wrong username or password). I don't think there's much we can do about this, unless we decide to go down the route of requiring a live SMB server for the tests.

robintw:
Update before I leave:

This is an extremely old PR that is basically working, but will probably need some changes to work with the latest version of Pepys. The users said at one point that they wanted all the data to be archived to a SMB (Samba - Windows File Sharing) share using a different username and password to the one the user is actually logged in to. This is so a secure archive of files could be created. The user's priorities changed, and we didn't have chance to properly test it on-site, so it got deprioritised.

This was actually remarkably difficult to do, and required using a pure-Python SMB library rather than the in-built Windows methods. It's all been tested and benchmarks, and so with a few modifications for the latest Pepys it should 'just work'.

PR #344, opened 2020-04-30
Fixes paths in deployed Python to allow alembic and distlib install

robintw:
The embedded version of Python that we're using for deployment works differently to a standard Python installation. Normally the current directory is added to the `sys.path` variable automatically, but that isn't the case in embedded Python. This causes problems when installation various packages via pip: specifically, distlib (which we got around by including a wheel in the repo) and alembic (which we could get around a similar way, but this is a better way to solve it now we've run into this issue for multiple packages)

This PR makes some changes to get around that, and get that functionality even in the embedded Python.

1. We add a new `.pth` file to the site-packages folder which does two things: adds an empty entry to the first element of sys.path and adds a specific folder (`pip\_vendor\pip517`) to the path too. The first of these fixes things for distlib, and the second allows the alembic install to work correctly (I'd have thought the alembic install would have been fixed by  the first change too, but it isn't - so we had to add the directory manually)

2. We add `import site` to the main `.pth` file to tell it to look for other `.pth` files (we can't do the modifications we did above in the main `.pth` file as general import statements aren't allowed there).

3. We install from pip without using cached files. This takes a bit longer, but means we don't get hidden errors where things work on my computer (because of the cache that I already have from other work) but don't on someone else's.

4. We no longer install the distlib wheel file manually - and so this PR deletes it from the repository.


robintw:
I've tested this on my main Windows machine. @IanMayo do you want to test creating a deployment on your machine before we merge this Ian?

(Creating a deployment just from this PR won't actually test installing alembic, as alembic is added to the requirements.txt in its own PR. You can either merge the other PR first, or just install alembic manually once you've done the deployment with `python -m pip install alembic` (after running `set_paths.bat` so you're using the right Python). Possibly best to leave the full test of this by you until after both PRs have been merged though (but ideally before we actually want to create a deployment!)

PR #346, opened 2020-05-02
Minor fix to REP importer

robintw:
Moves unit setting for depth to REPLine class, like all other variables - this means it comes up with the correct units in the highlighted output like all the other variables.


IanMayo:
@Robin, should there be a unit test to verify that the highlighter includes the units?

The unit test would also stop this functionality 'disappearing' in some future change.

robintw:
Thanks @IanMayo - done now.

IanMayo:
@robintw - it looks like you have a failing test.

PR #349, opened 2020-05-05
Add Binder version of Getting Started docs

robintw:
Adds a Binder version of Getting Started docs.

Currently available on Binder at https://mybinder.org/v2/gh/debrief/pepys-import/binder-xterm?filepath=docs%2FPepys%20Getting%20Started%20-%20Binder.ipynb - but this link will need changing once we merge this PR as it has the branch hard-coded into it.


IanMayo:
I have a strange issue with the Binder.   The first time I ran it, it started the import process, and started to ask about the missing data.

The second time, it failed with this error:
image

What happens for you, @robintw

IanMayo:
Nope, can't reproduce it now.

robintw:
I've worked out what the issue is. I'm on childcare duty now, but will fix it later today or tomorrow.

IanMayo:
@robintw  - I think it's worth doing a file-listing of the GPX file that's being imported, with an overview along the lines of:

"we're going to be working with this sample data-file. It's in the GPX file format that is common with the handheld GPS trackers popularised by cyclists & walkers.  The file contains some introductory metadata, followed by 5 position records.  Pepys contains a series of importer modules, one of which can recognise and load `*.gpx` files"

With this introduction, the users will then recognise the significance of the `5` in the states table, and use it as feedback that the import has worked.

Aah, there's something else we can demonstrate.  

If we get them to run the import twice, Pepys will warn them that the datafile has already been processed, with the date and time - which should be just a few minutes ago.  This is because of that single row in the datafiles table.  So, let's get them to view the tables from the admin page, and indicate the single `Datafile` entry is for the file they just processed.  Then, run the import again, it will be rejected, and we can highlight it's because of that existing entry.

Further down, I think there's a stray underscore, possibly missing from the start of `defaults resolver`:
image

Lastly, right at the end I think we should down-play the advice to obtain a SQLite browser, this tutorial is just for _users_, not _maintainers_ .  I think it's better to show the screenshot, explain it's from a SQLite Browser, and offer them a link just in case they want to download the file & have a look.

IanMayo:
@robintw  - I think it's ok for the GPX importer to create the GPS sensor for the platform, without asking the user (if that's possible).
image

IanMayo:
When the duplicate file import is rejected, we should highlight that the times are in GMT.  On first impressions they may appear one or more hours out (in BST or other timezone)

robintw:
@IanMayo The only way to create the GPS sensor automatically is if we hard-code a classification for it into the importer. Is this something that can be done for the GPS sensor? (I'm guessing not).

I've added a brief note about times being in UTC/GMT.

IanMayo:
Correct, we can't assume the classification.  I'm just checking with the clients though.

I _think_ e-trac data comes in over the Internet, so we may reliably know the privacy level.

IanMayo:
No, we can't assume the privacy.  So, forget that request :-)

PR #353, opened 2020-05-07
Add viewing data option to pepys admin app

### LINKED ISSUE #352: View table contents from Admin app
At least while the volumes are low, it would be good to be able to view table contents from the Admin app.

It could be a capacity that grows, I'll capture that in a separate issue.

But, to start with, we let the user type a table name, and we return the contents of the first 10 rows, with column headers.

Let's get that done, then learn about it.

I see these problems: 
* some will show foreign keys that don't make sense
* some keys will be really long GUIDs.  Aah, but they're prob nice simple counters in SQLite.

Note: this 'feature' could prove useful for the getting started guidance in Binder.

Discussion on UI thoughts started here:
https://pepys-import.slack.com/archives/CU09ELTNK/p1588777735006700

BarisSari:
Fixes #352


IanMayo:
@BarisSari - could we have a video walkthrough if/when you have something to show?

BarisSari:
> @BarisSari - could we have a video walkthrough if/when you have something to show?

Sure, @IanMayo . When I have something to show, I'll let you know.

pepys_admin/view_data_cli.py
```diff
@@ -0,0 +1,143 @@
...
+from prompt_toolkit.lexers import PygmentsLexer
+from pygments.lexers.sql import SqlLexer
+from sqlalchemy import inspect
+from sqlalchemy.ext.associationproxy import AssociationProxy
+from sqlalchemy.orm import RelationshipProperty, class_mapper, load_only
+from tabulate import tabulate
+
+
+def bottom_toolbar():
+    return HTML('Press <b><style bg="ansired">ESC+Enter</style></b> to exit!')
```
IanMayo, line 14: @BarisSari - when do we see this toolbar get displayed?

IanMayo:
If press <esc> while it's letting me choose at table name, Pepys_Admin falls over:
```
  File "/Users/ian/git/pepys-import/pepys_admin/view_data_cli.py", line 133, in default
    self.aliases[cmd_]()
  File "/Users/ian/git/pepys-import/pepys_admin/view_data_cli.py", line 69, in do_view_table
    elif selected_table.endswith("ies"):
```

screencast

pepys_admin/view_data_cli.py
BarisSari: It should be displayed when you select `run sql` option.

BarisSari:
> If press while it's letting me choose at table name, Pepys_Admin falls over:
> 
> ```
>   File "/Users/ian/git/pepys-import/pepys_admin/view_data_cli.py", line 133, in default
>     self.aliases[cmd_]()
>   File "/Users/ian/git/pepys-import/pepys_admin/view_data_cli.py", line 69, in do_view_table
>     elif selected_table.endswith("ies"):
> ```
> 
> screencast

I couldn't produce the same error, @IanMayo. Could it be that your database version is old?

IanMayo:
@BarisSari - isn't my issue because selected_table is `None` when the user presses `esc`?

IanMayo:
It may be these old eyes, but I can't read the text in read when entering SQL:
image

(From the code I know what it is, but a user won't)

IanMayo:
We get _hard crash_ if I type bad SQL:
`select * from contax`

We should report database error & leave user at `View table/Run SQL/Exit`.

BarisSari:
> I've gone through this PR to help me understand more of the details of SQLAlchemy, and have a couple of queries/notes:
> 
>     1. Are you using the `deferred` option of SQLAlchemy to mark columns that aren't to be displayed when viewing data? If so, could you put a couple of comments in to explain this somewhere (eg. a note in the viewing function to explain that it will only return non-deferred columns, plus a note where the deferred columns are first declared) Also, will this have any effect on other usages of these columns - or should they just populate themselves when needed? I see you've undeferred them in a few places - eg the table summary function - could you make a note in a comment about why they need to be undeferred there?
> 
>     2. I see you've changed the alembic version which is inserted into the table when creating the schema. Presumably I should have updated that when I did my 'add uniqueness constraints' PR - is that right? Could you add to the development instructions in the documentation on the migrations to point out that we need to do this - I can see myself forgetting in the future.

Hi, @robintw . Thanks for the review. 

1. I'll add comments related to deferr-undefer but I'd like to explain it here as well. The thing is, when you defer loading a column, it is not fetched from the DB when you don't add `undefer` option. So, if we would like to catch any attribute mismatch in a table, we do want to fetch all attributes. Therefore, I added the `undefer` option to `TableSummary` which is called for each table in `data_store.get_status()`. Alternatively, `load_only` might be used to specify columns: https://docs.sqlalchemy.org/en/13/orm/loading_columns.html#load-only-and-wildcard-options

2. I didn't change the version for SQLite but I did for PostgreSQL. Because the created database using `data_store.initialise` (which uses `Base.metadata.create_all`) doesn't fit with the database created by migration scripts(`base2` script should be run by Alembic) for SQLite. 
On the other hand, PostgreSQL doesn't have this problem. That's why I thought that I can directly stamp the initialised database with the latest migration. I'm glad that you asked for it because I think we need a strategy to follow. It's a bit concerning me because it might break some stuff if we apply current migrations(which have many uniqueness and foreign key constraints) after data import. What do you think, @robintw ?

BarisSari:
> @BarisSari - isn't my issue because selected_table is `None` when the user presses `esc`?
> It may be these old eyes, but I can't read the text in read when entering SQL:
> We get hard crash if I type bad SQL:
> select * from contax
>
> We should report database error & leave user at View table/Run SQL/Exit.

I'm working on them.

BarisSari:
> It may be these old eyes, but I can't read the text in read when entering SQL:
> image
> 
> (From the code I know what it is, but a user won't)

I thought that syntax highlighting could be a good feature. Apparently it is not. Would you like me to switch it to plain text?

robintw:
> Hi, @robintw . Thanks for the review.
> 
> 1. I'll add comments related to deferr-undefer but I'd like to explain it here as well. The thing is, when you defer loading a column, it is not fetched from the DB when you don't add `undefer` option. So, if we would like to catch any attribute mismatch in a table, we do want to fetch all attributes. Therefore, I added the `undefer` option to `TableSummary` which is called for each table in `data_store.get_status()`. Alternatively, `load_only` might be used to specify columns: https://docs.sqlalchemy.org/en/13/orm/loading_columns.html#load-only-and-wildcard-options

That makes sense and I'm fine with that, once a few comments have been added to explain it.

> 2. I didn't change the version for SQLite but I did for PostgreSQL. Because the created database using `data_store.initialise` (which uses `Base.metadata.create_all`) doesn't fit with the database created by migration scripts(`base2` script should be run by Alembic) for SQLite.
>    On the other hand, PostgreSQL doesn't have this problem. That's why I thought that I can directly stamp the initialised database with the latest migration. I'm glad that you asked for it because I think we need a strategy to follow. It's a bit concerning me because it might break some stuff if we apply current migrations(which have many uniqueness and foreign key constraints) after data import. What do you think, @robintw ?

Hmm yes, I can see the problem here. If we're starting to enforce foreign keys and uniqueness then we could have the migration fail. I think there's not really any way around this: luckily we've added these constraints before the client has a particularly large database - so there shouldn't be anything that fails these constraints, or if there is then it can be modified fairly easily. So, I think we can get away with it at the moment (@IanMayo - do you agree?), but we may not be able to get away with it if we do similar sorts of constraint changes in future.

I'm a little confused about why we can't stamp the initialised db with the latest migration version for SQLite - can you remind me why that is? What is in the base2 migration that isn't done by the initialise script? I can see the problem if we were to initialise a new SQLite db when running the import CLI, then it creates it but doesn't do the full alembic migration, and then we try and do the migration afterwards and find that the import (incorrectly) created some things that don't fulfil the constraints.

Should we move away from the initialise command using SQLAlchemy to create the tables, and do everything that involves creating tables through alembic? Or have we thought of that before and decided not to do it?

BarisSari:
> I'm a little confused about why we can't stamp the initialised db with the latest migration version for SQLite - can you remind me why that is? What is in the base2 migration that isn't done by the initialise script? I can see the problem if we were to initialise a new SQLite db when running the import CLI, then it creates it but doesn't do the full alembic migration, and then we try and do the migration afterwards and find that the import (incorrectly) created some things that don't fulfil the constraints.
> 
> Should we move away from the initialise command using SQLAlchemy to create the tables, and do everything that involves creating tables through alembic? Or have we thought of that before and decided not to do it?

When you try to migrate your new changes, it was producing extra `alter_column()` for `Geometry` table's geometry field. But I tested it again, and it didn't produce anything. I think we're okay to stamp with the latest migration for SQLite as well.

We considered to use running alembic during the `data_store.initialisation` but it could lead to migrate existing databases silently. Therefore, we moved it to Pepys-admin.

I'll update developer instructions of data migration.

PR #354, opened 2020-05-07
Enforce uniqueness of all reference table `name` columns

robintw:
Reference tables like `Privacy` and `SensorType` shouldn't have multiple entries with the same name. If we enforce this, then it doesn't matter if we search by name rather than ID in the code.

This PR adds a `unique=True` argument to each of the name columns in these tables, which will enforce uniqueness.


BarisSari:
@robintw , I think there should be migrations for this change. I'm pretty sure that migration for PostgreSQL will work without any problem. However, our current `declarative_base()` for SQLite makes a problem because of naming conventions (I faced this problem in my current PR). It would probably give the following error if you tried to migrate your changes for SQLite: `ValueError: Constraint must have a name`. If that's okay for you, I might create migration scripts in my PR which has a solution for this error.

IanMayo:
Making "real" use of our support for migrations sounds like a great opportunity to learn more lessons.

BarisSari:
> Making "real" use of our support for migrations sounds like a great opportunity to learn more lessons.

Yes, Ian. Actually, my view data PR is getting bigger. Therefore, I'll open a hotfix PR which will solve this issue. I'll explain why it's happening and what is our solution in detail.

IanMayo:
Deferring acceptance of this issue, until @BarisSari  has submitted his hotfix.

IanMayo:
@BarisSari - do the migration scripts look how you expect them to?

robintw:
Hi @BarisSari. I've just created the migration script for Postgres, and am getting lines like this in the upgrade function:

```
def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    with op.batch_alter_table("ClassificationTypes", schema="pepys") as batch_op:
        batch_op.create_unique_constraint(None, ["class_type"])

    with op.batch_alter_table("CommentTypes", schema="pepys") as batch_op:
        batch_op.create_unique_constraint(None, ["name"])
```

and this in the downgrade function:

```
def downgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    with op.batch_alter_table("Users", schema="pepys") as batch_op:
        batch_op.drop_constraint(None, type_="unique")

    with op.batch_alter_table("UnitTypes", schema="pepys") as batch_op:
        batch_op.drop_constraint(None, type_="unique")
```

I noticed that it seemed to be setting the name of the constraint to be None, and thought it might not work when we try and downgrade - and I was right, it doesn't.

I think we may need the same sort of naming conventions change as we did for SQLite in #360 but for Postgres instead. What do you think?

BarisSari:
> Hi @BarisSari. I've just created the migration script for Postgres, and am getting lines like this in the upgrade function:
> 
> ```
> def upgrade():
>     # ### commands auto generated by Alembic - please adjust! ###
>     with op.batch_alter_table("ClassificationTypes", schema="pepys") as batch_op:
>         batch_op.create_unique_constraint(None, ["class_type"])
> 
>     with op.batch_alter_table("CommentTypes", schema="pepys") as batch_op:
>         batch_op.create_unique_constraint(None, ["name"])
> ```
> 
> and this in the downgrade function:
> 
> ```
> def downgrade():
>     # ### commands auto generated by Alembic - please adjust! ###
>     with op.batch_alter_table("Users", schema="pepys") as batch_op:
>         batch_op.drop_constraint(None, type_="unique")
> 
>     with op.batch_alter_table("UnitTypes", schema="pepys") as batch_op:
>         batch_op.drop_constraint(None, type_="unique")
> ```
> 
> I noticed that it seemed to be setting the name of the constraint to be None, and thought it might not work when we try and downgrade - and I was right, it doesn't.
> 
> I think we may need the same sort of naming conventions change as we did for SQLite in #360 but for Postgres instead. What do you think?

Hi @robintw. Yes, you're right. We need the same change for PostgreSQL as well. I was expecting that Alembic is smart enough to understand names for PostgreSQL. Apparently, it doesn't. I'm opening another hotfix PR.

PR #360, opened 2020-05-11
Hotfix - Pass Naming Conventions for SQLite

BarisSari:
As it's mentioned in #354, SQLite creates constraints without a name if the naming convention is not specified. In this PR, `naming_conventions` added to `declarative_base()` of SQLite. When Alembic runs and there are new constraints in the models, it passes `None` to the name field of `alembic.drop_constraint()` and `alembic.create_constraint()` methods. However, `drop_constraint() `can't work without a constraint name. It should be passed via `Base` model or `alembic.batch_alter_table()`

Example upgrade and downgrade methods for a unique constraint are as follows:

```python
def upgrade():
    with op.batch_alter_table("ClassificationTypes", schema=None) as batch_op:
        batch_op.create_unique_constraint(
            batch_op.f("uq_ClassificationTypes_class_type"), ["class_type"]
        )

def downgrade():
    with op.batch_alter_table("ClassificationTypes", schema=None) as batch_op:
        batch_op.drop_constraint(batch_op.f("uq_ClassificationTypes_class_type"), type_="unique")
```

As it has seen the first parameter of constraint methods are named according to the given naming convention instead of `None`.

It seems that SQLite breaks the classic rules of databases because of this. And Alembic has a section about the issues of unnamed constraint behavior of SQLite: https://alembic.sqlalchemy.org/en/latest/batch.html#dropping-unnamed-or-named-foreign-key-constraints
https://alembic.sqlalchemy.org/en/latest/batch.html#including-unnamed-unique-constraints

**Note:** I have tested upgrading and downgrading an SQLite DB, which has filled tables. However, it was a manual test. @IanMayo , do you think that we need tests to check whether constraints are created with names or not? I believe that it's possible to implement some tests using Alembic's Command API.


IanMayo:
@BarisSari - could you create an issue to produce the tests that constraints are properly created?

@robintw - could you check that you're up-to-speed with this new SQLAlchemy practice, please?

PR #369, opened 2020-05-13
Update Getting Started guides

### LINKED ISSUE #367: Update Getting Started to reflect reduced hard-coding.
There's a lot shared between the static `Getting Started` and the Binder/xterm version of `Getting Started`.

Both require updating now that we've removed lots of hard-coded values.

As we do it, it would be useful to explain to the user why they're about to enter information, as well as the detail of what they're about to enter.

robintw:
This PR updates the Getting Started guides - both the Binder notebook and the guide in the docs.

The main change is to add information on why we're asking all the questions during import, and suggestions for what to fill in as the answers.

For the docs guide I've also added and removed a few steps to make it more similar to the Binder guide, and to remove stuff that is now in the installation guide.

Fixes #367


IanMayo:
@robintw  - is this ready for merge?

robintw:
Yes, assuming you're happy with it.

PR #382, opened 2020-05-15
Add tests for uniqueness constraints

robintw:
The PR to add uniqueness constraints to the reference tables (#354) was accidentally merged without any tests. This PR adds tests for these uniqueness constraints.

### DIFF
--- a/tests/test_uniqueness.py
+++ b/tests/test_uniqueness.py
@@ -1,0 +1,46 @@
+import pytest
+from sqlalchemy import exc
+from sqlalchemy.orm.session import sessionmaker
+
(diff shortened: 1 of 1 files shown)


IanMayo:
@robintw - I think it's probably worth you double-checking that this PR matches our new understanding of uniqueness, please.

robintw:
@IanMayo I've fixed this to work with the new `level` parameter to `Privacy`. All of this still applies in terms of uniqueness as this PR only tests uniqueness for reference tables - I've updated the title to make this clear. There will be new tests added for uniqueness in the separate PR that adds unique constraints for the metadata tables.

PR #391, opened 2020-05-19
Export snapshot

### LINKED ISSUE #374: Export snapshot of database for offline usage
The database schema for Pepys organises the data into 3 logical groupings:
**Reference**: general "types" of things
**Metadata**: instances of specific things
**Measurements**: measured data

image

There would be value in an analyst being able to take away a database snapshot (prov via SQLite) that contains the contents of the Reference grouping, and possibly the Metadata grouping.

The benefits of this are:
* it reduces the amount of new entries that the analyst must produce
* it reduces the instance of data duplication (there is only one instance of `HMS Trumpeter`).

We should add a new feature to `pepys_admin` to support this export.  The feature will include `Export Reference data` and `Export Reference and Metadata data`, and will export to a SQLite filename specified by the current user.

Note: it is not expected to export a complete `Slave` instance of the database, just some useful metadata.

BarisSari:
Fixes #374


IanMayo:
@BarisSari - warning. I'm talking the clients about how we decide what parts of the `Metadata` section should be exported.

While writing the email, I realised there may not be set of tables that it always makes sense to export.  But, I did realise that the answer may be in the `Privacy` attribute.  So, the analyst may choose to export all metadata content that is public, but not private.   

However, the above is still being discussed. I'll update you when I know more.

robintw:
This all looks good @BarisSari - just a couple of minor things:

1. Do you think it'd be worth moving all of the export-related functionality out of the main admin CLI file, and into a separate `export.py` file? I think it'd make things cleaner, and more obvious which files have which functionality.

2. When I export to a SQLite db using your code and then use Pepys Import to import some new data into that db (like the clients would if they exported and then took the exported database to sea on a laptop) then I get a message saying that the DataStore is initialising the tables. It seemed to just create the other tables it needed and accept that the reference and platform tables were already there and didn't overwrite them - but it might be worth writing a formal test for that, and checking that it works ok in all situations.

IanMayo:
That all sounds good, thanks @robintw .

* Yes, to new `export_snapshot.py` file, please
* Hmm, I guess we need to initialise (create Pepys schema) the new snapshot file before we move the data into it.

BarisSari:
I made the changes. New database is initialised at the beginning and all export-related functions are moved to `export_snapshot.py`.

PR #395, opened 2020-05-20
Hotfix - Don't check if schema is created for Clear Schema command

BarisSari:


pepys_import/core/store/data_store.py
```diff
@@ -1028,16 +1028,14 @@ def clear_db_schema(self):
         """Delete the database schema (ie all of the tables)"""
         if self.db_type == "sqlite":
             meta = BaseSpatiaLite.metadata
+            with self.session_scope():
+                meta.drop_all()
+            with self.engine.connect() as connection:
+                if self.db_type == "sqlite":
```
IanMayo, line 1034: we're already in a `if` block that checks the `db_type`

BarisSari: Yes, I removed it. The second with clause is also unnecessary. I removed it as well:
https://github.com/debrief/pepys-import/pull/395/commits/7c271b6cade19a8df188ffd1d40c0e0e92786198

PR #397, opened 2020-05-20
Add note on encrypting username and password to docs

### LINKED ISSUE #396: Config file documentation needs to record encryption marker
In the config file, I believe we use marker characters to indicate that a password has been encrypted:
https://github.com/debrief/pepys-import/issues/203

The documentation for the config file should indicate that the password can be marked as encrypted (without identifying what the encryption algorithm is).

robintw:
Fixes #396

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -28,8 +28,8 @@ Configuration file variables
 These settings control how pepys-import connects to the database. The specific
 variables are:
 
- - :code:`db_username`: Username used to connect to the database server (default: :code:`postgres`). Only used for PostgreSQL connections.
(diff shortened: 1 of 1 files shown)


IanMayo:
Thanks, my observation related to the `[database]` section.

I just checked, we do support encryption of user & pwd fields in that section.  Hmm, should we duplicate your paragraph at the foot of the `database` section?

Aah, better idea.  We append a note to each parameter indicating something like 'can be encrypted, see Encryption note below`.  Then all 4 potentially encrypted fields can pull the user to the description.

robintw:
Thanks @IanMayo - fixed now.

PR #403, opened 2020-05-20
Hotfix - Replace sensor_name/platform_name attributes

### LINKED ISSUE #372: Replace manual sensor_name/platform_name attributes with SQLAlchemy relationships
In the definitions of the `State`, `Contact` and `Comment` classes, we have defined instance variables called `sensor_name` and `platform_name`. When we add one of these classes (in, for example `Datafile.create_state`) we manually set these variables based on the sensor object that we have. As far as I can see, they're only used in the extended validator to give more useful errors to the user.

The same functionality could be implemented using SQLAlchemy relationships, and then these would automatically stay up-to-date, rather than having to be manually set on creation. We're starting to use SQLAlchemy relationships in #353 - so I think it'd be good to use them here too.

What do you think @BarisSari?

BarisSari:
Fixes #372


BarisSari:
I couldn't switch all platform_name and sensor_name attributes because `State, Contact` tables don't have a relationship with `Platform`, and `Comment` table doesn't have a relationship with `Sensor` table. Do you think that it's okay, @robintw ? Or is it more confusing now?

robintw:
Hmm, I'm not sure.

I think there might be a way to do this in SQLAlchemy - something like method 2 in [this answer](https://stackoverflow.com/questions/17580649/sqlalchemy-relationships-across-multiple-tables) - what do you think? It'd be nice if it was all done automatically by SQLAlchemy.

BarisSari:
> Hmm, I'm not sure.
> 
> I think there might be a way to do this in SQLAlchemy - something like method 2 in [this answer](https://stackoverflow.com/questions/17580649/sqlalchemy-relationships-across-multiple-tables) - what do you think? It'd be nice if it was all done automatically by SQLAlchemy.

Thanks for the link, @robintw . It seems like I missed the possible attributes of `relationship` in SQL Alchemy documents! I'm trying now.

Issue #399, opened 2020-05-20
Bug in importing bulk data


IanMayo:
I've wiped the remote Postgres database, and want to shove a load of data in.

Aah, it wasn't completely clear.  I used the Pepys_Admin interface to `import reference data` and `import metadata`.  That may have caused a duplication conflict. 

The import failed when I ran:
 `python -m pepys_import.cli --resolver default --path tests/sample_data/track_files/`

I've done that bulk regularly.  It has just failed while loading a file that we know (from the filename) is _bad_.

Note: if I import just that file, it works.  If I import the hole set of folders, we get the error reported below.

Error log:
```
NMEA Importer working on NMEA_bad.log
 83%|███████████████████████████████████████████████████▋          | 5/6 [00:00<00:00, 11.03it/s]
Traceback (most recent call last):
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/Cellar/python/3.7.6_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/ian/git/pepys-import/pepys_import/cli.py", line 94, in <module>
    main()
  File "/Users/ian/git/pepys-import/pepys_import/cli.py", line 38, in main
    process(path=args.path, archive=args.archive, db=args.db, resolver=args.resolver)
  File "/Users/ian/git/pepys-import/pepys_import/cli.py", line 90, in process
    processor.process(path, data_store, True)
  File "/Users/ian/git/pepys-import/pepys_import/file/file_processor.py", line 162, in process
    file, current_path, data_store, processed_ctr, import_summary
  File "/Users/ian/git/pepys-import/pepys_import/file/file_processor.py", line 260, in process_file
    data_store, full_path, highlighted_file, datafile, change.change_id
  File "/Users/ian/git/pepys-import/pepys_import/file/importer.py", line 103, in load_this_file
    self._load_this_file(data_store, path, file_object, datafile, change_id)
  File "/Users/ian/git/pepys-import/pepys_import/file/importer.py", line 123, in _load_this_file
    self._load_this_line(data_store, line_number, line, datafile, change_id)
  File "/Users/ian/git/pepys-import/importers/nmea_importer.py", line 107, in _load_this_line
    change_id=change_id,
[...]

Issue #379, opened 2020-05-15
Establish uniqueness criteria for tables


robintw:
This is really a 'point for discussion' issue with @IanMayo , with the hope that I will transfer this information to the wiki once the details have been agreed. We may also want to enforce some of these uniqueness criteria - I'll look into how to do this across multiple fields on both Postgres and SQLite.

The reason for establishing these criteria is two-fold:
 - We need to know what is valid to search by in the code. For example, it wasn't valid to search just by sensor name, as that isn't unique - you need to specify a platform too. That has been fixed, but we've got similar problems with searching for other entities such as Platform (#362)
 - When merging the databases (#375) we need a way to work out whether rows are the same or different. For example, if we have two rows from the Platform table, we need to work out whether they refer to the same platform. This requires understanding what makes a Platform unique.

So, the list below says what combination of fields make an entry unique across various tables. It's useful to read this list in conjunction with the [database schema diagram](https://docs.google.com/drawings/d/1kEEYdoU5H7rcvxCaf9Aeyh0ujvEizwZWJE4uNGGNK5o/edit).

**Reference tables:**
All reference tables are unique on the `name` field. This is already enforced by uniqueness constraints.

**Metadata tables:**
 - **Platforms**: unique on the combination of [`Name`, `Nationality`, `Type`] - that is, there are no duplicated platform names within a specific nationality and platform type (eg. not two UK submarines called Dolphin, but you can have a UK submarine and a UK frigate both called Dolphin)
 - **Sensors**: unique on the combination of [`Name`, `Host`] - that is, each Platform should have unique names for its sensors
 - **Datafiles**: unique on the combination of [`Reference`, `Hash`, `Size`]
 - **Synonyms**: unique on the combination of [`Table`, `Synonym`] - that is, each synonym must be unique for a specific table (eg. you can't have two Sensor synonyms with the same name)

**Measurement tables:**
These are harder to establish - and it partly depends on the answers in #375 as to whether we need to work out a way to establish uniqueness of this. The simplest is just to consider all fields when deciding uniqueness - presumably there should never be a case where all fields are equal for multiple rows?
 - **States**: TODO ?
 - **Contacts**: TODO ?
 - **Comments**: TODO ?

**Implementation:**
If we're going to rely on these uniqueness constraints, then we should probably enforce them in the database. We can do this with code like this in the table definition:

```
__table_args__ = (UniqueConstraint('customer_id', 'location_code', name='_customer_location_uc'))
```

[...]

Issue #362, opened 2020-05-12
Stop searching for Sensors/Platforms/etc by name - as names may be duplicated


robintw:
This is a fairly big change, but one that should be done sooner or later. A lot of places in the code search for things like Sensors or Platforms by name - but that isn't necessarily valid as we may have duplicate names for Sensors or Platforms. Generally, when we have been passing something like `platform_name` around, we should be passing `platform_id` instead.

I'm part-way through scoping this issue, and will post here with more details when I've worked out what needs to be done.

IanMayo:
> This is a fairly big change, but one that should be done sooner or later.

I think you mean `sooner rather than later`.

Yes, I support this change.

robintw:
Yes I do - it's next on my todo list.

PR #404, opened 2020-05-21
Hotfix - Fix NMEA import bug

robintw:
Ian reported a bug in #399, which occurs under the following circumstances:

1. Metadata is imported through Pepys Admin
2. NMEA_trial.log is imported
3. NMEA_bad.log is imported

The bug was caused by the issue described in #362 - basically, we ended up with the wrong platform linked to a sensor, because we were search for a platform by name and had a duplicate platform name in the database.

#362 is on the list of things to fix, but isn't top priority at the moment, so this hotfix temporarily fixes the issue by allowing _either_ a host name or a host_id to be specified when adding a sensor. A full fix will move to using host_ids exclusively, and searching for platforms by their proper uniqueness criteria (as defined by #379).


IanMayo:
@robintw - Travis has highlighted a couple of `black` linting issues.

robintw:
I think this is ready to merge now @IanMayo - the coverage is relatively low on the diff, but I'm not sure it's worth trying to increase that now given that this is only a temporary fix.

PR #408, opened 2020-05-24
Offer Default Values in CLI

### LINKED ISSUE #406: Offer some default CLI responses
I'm hoping to streamline the cli-resolver, by thinking through where we can offer sensible defaults.

Where a reference table contains less than, hmm, 7 entries, we could offer those entries in the CLI question.  Not only will this be quicker than typing in the name & pressing [tab], users may develop _muscle-memory_, and work more quickly through the data.

Looking at the schema, I _think_ these tables may contain 7 or less entries:
* Platform Types
* Sensor Types
* Privacies
* Datafile Types
* Media Types

Aah, there's at least one `metadata` table that we could do this with.  When we're importing contact data, frequently we have to assign `platform` and `sensor`.  But, once we've selected the platform, we can certainly show a list of sensors if there are 7 or less.

Nationalities will contain more than 7, but it's possible to identify 1 that will occur far more than others (`UK`), plus another 6 that occur very frequently.

We could automatically support offering the first 7 as menu options for a table with less than 7 entries.  But, the `Nationalities` solution is more challenging.  It _could_ be resolved by adding a 'Priority' attribute.  `1` would always be offered first,  `2` entries would follow, and anything else just gets supported by the existing fuzzy search.

Any other ideas?

The above _priority_ strategy could be added to other tables as the content of the `Reference` tables mature - and the analysts get a feel for what's used most often.

BarisSari:
Fixes #406 

Since I needed to change the Command Line Resolver, I thought that it might be a good time to refactor it a bit. Therefore, I merged all resolve and fuzzy search methods for reference data. Then, it became easier to add 7 or less options to the menu with only changing 1 method.


IanMayo:
@BarisSari - I couldn't see the code that processes the `nationality` - `priority`, to put them in the command line resolver.

Hmm, I couldn't see the bits that pull existing values (if less than 7 present) into the command-line menu.

Is this a partial PR?

IanMayo:
(sorry @BarisSari - commented deleted. I hadn't seen inside the `Large diff`)

BarisSari:
> (sorry @BarisSari - commented deleted. I hadn't seen inside the `Large diff`)

That's fine, @IanMayo . If you would like to see only that part of the code, here it is:

https://github.com/debrief/pepys-import/blob/33ff6cdf2953743bac828bcd298a19f99fdb0605/pepys_import/resolvers/command_line_resolver.py#L164-L181

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -107,56 +121,148 @@ def resolve_sensor(self, data_store, sensor_name, sensor_type, host_id, privacy,
...
-        # Choose Privacy
-        privacy_names = [
-            "Search an existing classification",
-            "Add a new classification",
-        ]
-        choice = create_menu(
-            f"Ok, please provide classification for new {data_type}: ",
-            privacy_names,
-            validate_method=is_valid,
+    def resolve_reference(
```
robintw, line 158: Could you add a docstring to this method, explaining the parameters? It's a little unclear at first look what text_name and field_name and so on do.

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -107,56 +121,148 @@ def resolve_sensor(self, data_store, sensor_name, sensor_type, host_id, privacy,
...
             return None
+        elif choice not in completer:
+            new_choice = create_menu(
+                f"You didn't select an existing {text_name}. " f"Do you want to add '{choice}' ?",
+                choices=["Yes", f"No, I'd like to select an existing {text_name}"],
+                validate_method=is_valid,
+            )
+            if new_choice == str(1):
+                plural_field = (
+                    re.sub(
```
robintw, line 246: Could you add a comment on what this regexp is aiming to do?

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -673,14 +444,16 @@ def add_to_platforms(
...
+        :type change_id: UUID
         :return:
         """
         print("Ok, adding new platform.")
         if platform_name:
             platform_name = prompt("Please enter a name: ", default=platform_name)
+            trigraph = None
             if len(platform_name) >= 3:
                 trigraph = prompt("Please enter trigraph (optional): ", default=platform_name[:3])
+            quadgraph = None
```
robintw, line 456: Do we really want to make it so we don't ask them for a trigraph/quadgraph if the platform name is shorter than 3/4. Can trigraphs/quadgraphs exist in these circumstances @IanMayo?

IanMayo: Thanks @robintw  - good question.

Yes, they can be shorter than 4 chars: https://en.wikipedia.org/wiki/ICGV_%C3%9E%C3%B3r

(I've just learned something new about uniqueness - I'll add it to that issue)

IanMayo:
@robintw  & @BarisSari  - Robin asked about the sort order in Privacy labels.  I'd like to get this version in front of the users, so I'm happy to defer privacy order for a future release.

So, for now it will just be the order in which they're presented in the `csv` file.

robintw:
One other thing I've thought of: we'll need to update the getting started guide and the Binder getting started notebook to take into account these changes.

Do you want me to do this?

IanMayo:
> One other thing I've thought of: we'll need to update the getting started guide and the Binder getting started notebook to take into account these changes.
> 
> Do you want me to do this?

Yes please, @robintw

pepys_import/resolvers/command_line_resolver.py
BarisSari: Added:
https://github.com/debrief/pepys-import/pull/408/commits/887fdf8132ae3643f9be5fe7892f2fba5b66a097

pepys_import/resolvers/command_line_resolver.py
BarisSari: Added:
https://github.com/debrief/pepys-import/pull/408/commits/887fdf8132ae3643f9be5fe7892f2fba5b66a097

BarisSari:
> One other thing I've thought of: we'll need to update the getting started guide and the Binder getting started notebook to take into account these changes.
> 
> Do you want me to do this?

Thanks for updating the guide and the notebook, @robintw . I added docstrings and a comment for the RegEx.

pepys_import/resolvers/command_line_resolver.py
BarisSari: Solved. Trigraph/quadgraph are asked even though platform name is less than 3/4 characters:
https://github.com/debrief/pepys-import/pull/408/commits/e2c48428d09b4765560facd81a944f4df2620c79

IanMayo:
Ok, @robintw - I'll leave it to you to review the changes, thanks.

PR #411, opened 2020-05-28
Hotfix - Accept Quoted Sensor Names

### LINKED ISSUE #400: Bug in Imported Sensor Name
When I import sample data with the following command: `python -m pepys_import.cli --path tests/sample_data/track_files --resolver default`, It creates a sensor named: `"` (15th row). The row above it looks strange, too.

Screen Shot 2020-05-20 at 13 58 06

BarisSari:
Fixes #400 

**Note:** This PR also adds if clauses to each Replay Importer because they should skip empty rows instead of throwing parsing errors.


importers/replay_comment_importer.py
```diff
@@ -27,6 +27,8 @@ def can_load_this_file(self, file_contents):
         return True
 
     def _load_this_line(self, data_store, line_number, line, datafile, change_id):
+        if line.text == "":
```
IanMayo, line 30: I _think_ you may be able to `strip()` the `line.text` before you do this test.

We should skip the line if it contains a couple of spaces, or a tab.

BarisSari: Done:
https://github.com/debrief/pepys-import/pull/411/commits/ffc17fa986e6084a19c037b931e69822eab3098f

pepys_import/file/highlighter/support/line.py
```diff
@@ -66,17 +67,33 @@ def tokens(self, reg_exp=WHITESPACE_DELIM, strip_char=""):
...
 
+            for match in finditer(quoted_name, child.text):
+                original_name = match.group()
+                quoted_sensor_name = original_name[1:-1].strip()
+                subtoken_sensor_name = SubToken(
+                    match.span(), quoted_sensor_name, int(child.line_start), child.chars
+                )
+                subtokens_sensor_name = [subtoken_sensor_name]
+
+                del_indexes = list()
```
robintw, line 86: Could you add some comments here explaining what we're trying to do? I'm a little bit confused by exactly what the aim of this chunk of code is

tests/test_admin_cli.py
```diff
@@ -239,7 +239,7 @@ def test_do_status(self):
         output = temp_output.getvalue()
 
         states_text = "| States       |              738 |"
-        contacts_text = "| Contacts     |               15 |"
```
robintw, line 242: Hmm. Why has this change increased the number of contacts recorded in the database?

BarisSari: Because we started to skip empty rows in Replay Importers and `sen_frig_sensor.dsf` had only errors due to that. It is imported by Replay Contact Importer now, and the number of Contacts increases. (I'm adding a note for that)

pepys_import/file/highlighter/support/line.py
BarisSari: Sure. I'll add more comments to the code and let you know 👍

BarisSari: @robintw , I reliased that the second for loop was really confusing. That's why I moved the quoted text's regex above of the main loop and added many comments:
https://github.com/debrief/pepys-import/pull/411/commits/04b3b18c7a3bb88c7ae586a32748a7849d5c66b5

What I was doing is as follows: Use the main loop to split a line into tokens-> Search for quoted texts in line(the second for loop)-> If there is any, find previously added tokens and delete them in tokens_array. Add the quoted text instead.

But right now, the code does the following: Search for a quoted text in line (use `re.search() `instead of for loop)->Find it's start and end indexes in the line-> In the main loop, check if the quoted text contains token. If it includes, then doesn't add the token to tokens_array.

I hope that the code is tidier now. If you have any other questions, I'd happily answer.

PR #416, opened 2020-05-28
Add extra progress bar when reading file

### LINKED ISSUE #413: Minor issue - delay before progress bar opens with very large file
I'm importing the `NMEA_TRIAL.log`, there is a noticeable delay (>10 seconds) after answering the questions, before the progress bar appears.  This gives the mistaken impression that the import has hung.

Screen recording [[here]](https://youtu.be/fD7EYb4Rls8)

_Maybe_ this is while the file is being loaded, or the number of lines in the file are being counted.

Are we able to either show a progress monitor for this operation, or to add the operation to the functionality that is covered by the existing progress bar?

robintw:
#413 reports an issue with a long delay without any changes in the UI when loading a large file. This was coming from the `fill_char_array_if_needed` function, which fills the underlying character array that the highlighting functionality uses.

This PR wraps the list comprehension there with tqdm to give an extra progress bar while that process is completing.

Fixes #413.

### DIFF
--- a/pepys_import/file/highlighter/highlighter.py
+++ b/pepys_import/file/highlighter/highlighter.py
@@ -1,3 +1,5 @@
+from tqdm import tqdm
+
 from pepys_import.file.highlighter.support.line import Line
 
 from .support.char import Char
@@ -112,7 +114,7 @@ class HighlightedFile:
         # We do this as a list comprehension as it's more efficient, but we have to
         # add it to list that already exists in self.chars, as references have already
         # been made to this list
-        self.chars += [Char(c) for c in file_contents]
+        self.chars += [Char(c) for c in tqdm(file_contents)]
 
     def create_lines(self, file_contents, lines_list):
         """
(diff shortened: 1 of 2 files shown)


IanMayo:
> which fills the underlying character array that the highlighting functionality uses.

Aah, we're getting into "special-case" territory here.  For huge system-generated files, it's unlikely that the clients will need extraction highlighting - since there's too much information for a human to actually inspect.  It's the ones with human input that have the strongest justifiication for extraction highlighting.

For such large files, extraction highlighting would either be switched off, or only run for the first 100, 200, x00 lines - sufficient to verify it's working as expected.

Anyway - I'm pleased it's fixed for this issue :-)

robintw:
Thanks @IanMayo. Just to follow-up briefly on the points you made:

1. Highlighting can currently be turned off for a particular importer by modifying the code (ie. setting a variable to False). We cannot yet turn off the highlighter for a specific file - but could potentially add that later on.

2. At the moment we can't run the highlighting for only the first `n` lines. Because of the way the highlighter works, this would mean that only those lines were actually processed by the importer - as the importer uses the highlighter methods to actually get hold of the text from the file. This could potentially be changed in future, but might be fairly difficult to do. Just wanted to make you aware.

PR #417, opened 2020-05-29
Give More Info on Selections in Pepys-Admin

### LINKED ISSUE #410: Give more information on what you're selecting in the Pepys Admin GUI
In the Pepys Admin GUI we have quite a few selection steps using things like iterfzf. Most of the time we don't explain what the user is selecting, why they're selecting it, and what it will mean when they select it. Sometimes this is obvious, sometimes it isn't.

Assuming there is a way of putting a bit of explanatory text above or below the iterfzf interface, we should alter all of these usages to put some explanatory text explaining what they're choosing and why (eg: "Choose platform to export").

## Added by Ian
Please also include this fix, related to `Export Snapshot` with this issue.
```
One UX-related issue I've just realised @BarisSari - there is no explanation anywhere that
you have to use TAB to select multiple entries in the privacies list. Is there a way of
putting a bit of explanatory text above or below the list to explain that? Actually, that 
explanatory text should explain that you're choosing which privacies of data to export 
(phrased in such a way that it comes across that you will export all data with those 
privacies, rather than just those specific Privacy table entries themselves). I've actually 
created an issue about this more broadly - I think there are various places in Pepys 
Admin where we ask the user to select things without giving them much information on 
what they're selecting and why - see #410.
```

BarisSari:
Fixes #410 .

Note: I realised that when selecting Privacy is canceled during the export reference and metadata, the Pepys Admin was failing. This PR has a fix for this issue as well.

### DIFF
--- a/pepys_admin/admin_cli.py
+++ b/pepys_admin/admin_cli.py
@@ -270,7 +270,7 @@ class AdminShell(cmd.Cmd):
             ).all()
             privacy_dict = {name: privacy_id for privacy_id, name in privacies}
         message = (
-            "Export all data with the selected classification(s). " "(Press TAB) for multi-select >"
+            "Export all data with the selected classification(s). (Press TAB) for multi-select >"
         )
         selected_privacies = iterfzf(privacy_dict.keys(), multi=True, prompt=message)
         if selected_privacies is None:
--- a/pepys_admin/admin_cli.py
+++ b/pepys_admin/admin_cli.py
@@ -267,6 +267,9 @@ class AdminShell(cmd.Cmd):
             ).all()
             privacy_dict = {name: privacy_id for privacy_id, name in privacies}
         selected_privacies = iterfzf(privacy_dict.keys(), multi=True)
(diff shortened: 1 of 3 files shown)


pepys_admin/admin_cli.py
```diff
@@ -266,7 +269,13 @@ def do_export_reference_and_metadata_data(self):
                 self.data_store.db_classes.Privacy.name,
             ).all()
             privacy_dict = {name: privacy_id for privacy_id, name in privacies}
-        selected_privacies = iterfzf(privacy_dict.keys(), multi=True)
+        message = (
+            "Export all data with the selected classification(s). (Press TAB) for multi-select >"
```
robintw, line 273: Can you just fix the brackets here, so it is `(Press TAB for multi-select)`?

BarisSari: Fixed:
https://github.com/debrief/pepys-import/pull/417/commits/cfddc0bc3625262fe6df9a0e97949c3106eae214

PR #420, opened 2020-05-29
Reorganise Top Level of Pepys Admin

### LINKED ISSUE #415: Minor issue - reorganise top level of Pepys Admin
If/when we add the `Help` option ( #414 ) to Pepys Admin, the top level is full.

We're going to need a _slot_ for the `merge database` option.

We can solve it by tidying the `Export` functionality.

I propose that we move 3 and 4 down into children of a new `3). Export.`

I propose that we also move 7 and 8 down into children of a new  `4) Snapshot`. The terminology of these commands will be renamed to `Create snapshot with Reference data` and `Create snapshot with Reference data & Metadata`

When we do introduce `Merge database` it can go under `4) Snapshot`.

(I'm doing the above to try to allow 5 and 6 to keep their number, so we don't have to update existing help/guidance)

BarisSari:
Fixes #415


IanMayo:
Sorry @BarisSari - there are some conflicts to sort.

BarisSari:
> Sorry @BarisSari - there are some conflicts to sort.

Solved, @IanMayo .

PR #422, opened 2020-05-31
Move alembic status to the end

### LINKED ISSUE #407: Tidy database migration report status
We output some Alembic migration-related information in `pepys-admin` Status report.

It would be neater if we:
* moved the Alembic status to the end of the `Status` report
* preceded the Status with `## Database version`
* appended the status with a horizontal bar of statuses

BarisSari:
Fixes #407 .

If we don't modify Alembic's output, we might print the database version in two ways:
**With verbose**
Screen Shot 2020-05-31 at 19 52 20
**Without verbose**
Screen Shot 2020-05-31 at 19 52 45

In the issue, Ian suggested "_appended the status with a horizontal bar of statuses_".
I'm not sure if it's worth to modify Alembic's output because it might be challenging. @IanMayo, If you think that that's okay to go with one of the options, I won't try to manipulate anything. If not, I'll work on it.


IanMayo:
Sure - let's not try to get inside Alembic status reporting.

The non-verbose versions seems ok.

PR #427, opened 2020-06-02
Fix Failing Alembic Commands

### LINKED ISSUE #368: Various alembic commands fail
When running `alembic history` I get the following error:

```
  File "/Users/robin/anaconda3/envs/pepys/bin/alembic", line 8, in <module>
    sys.exit(main())
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/config.py", line 577, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/config.py", line 571, in main
    self.run_cmd(cfg, options)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/config.py", line 551, in run_cmd
    **dict((k, getattr(options, k, None)) for k in kwarg)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/command.py", line 430, in history
    _display_history(config, script, base, head)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/command.py", line 399, in _display_history
    base=base or "base", head=head or "heads"
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/base.py", line 215, in walk_revisions
    head, base, inclusive=True, assert_relative_length=False
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/revision.py", line 765, in _iterate_revisions
    requested_lowers = self.get_revisions(lower)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/revision.py", line 321, in get_revisions
    resolved_id, branch_label = self._resolve_revision_number(id_)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/revision.py", line 501, in _resolve_revision_number
    self._revision_map
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/util/langhelpers.py", line 230, in __get__
    obj.__dict__[self.__name__] = result = self.fget(obj)
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/revision.py", line 123, in _revision_map
    for revision in self._generator():
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/base.py", line 103, in _load_revisions
    for file_ in Script._list_py_dir(self, vers):
  File "/Users/robin/anaconda3/envs/pepys/lib/python3.7/site-packages/alembic/script/base.py", line 876, in _list_py_dir
    return os.listdir(path)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/robin/Documents/IanMayo/pepys-import/migrations/versions'
```

I get the same error when running `alembic show head`.

I assume it is due to Alembic not expecting the `postgres_versions` and `sqlite_versions` folders - but I expect @BarisSari can advise.

### LINKED ISSUE #393: Fix issue with automatically-generated geometry migrations for SQLite
When autogenerating a migration with alembic for SQLite, it automatically generates the following changes for the geometry tables, even when there has been no change in them:

```
INFO  [alembic.autogenerate.compare] Detected type change from INTEGER() to Geometry(geometry_type='POINT', srid=4326, management=True) on 'Contacts.location'
INFO  [alembic.autogenerate.compare] Detected NOT NULL on column 'Geometries.geometry'
INFO  [alembic.autogenerate.compare] Detected type change from NUMERIC() to Geometry(management=True) on 'Geometries.geometry'
INFO  [alembic.autogenerate.compare] Detected type change from INTEGER() to Geometry(geometry_type='POINT', srid=4326, management=True) on 'Media.location'
INFO  [alembic.autogenerate.compare] Detected type change from INTEGER() to Geometry(geometry_type='POINT', srid=4326, management=True) on 'States.location'
  Generating /Users/robin/Documents/IanMayo/pepys-import/migrations/sqlite_versions/2020-05-19_3a23906a866f_.py ...  done
```

BarisSari:
Fixes #368 
Fixes #393 

PR is ready for review. I think the best way to test this PR is by running the following commands:
```bash
> alembic history
# Expected output
(pepys-import) baris@bariss-MacBook-Pro pepys-import % alembic history
INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
f103f27c4575 -> b03fb305981b (head), Add level attribute to Privacies Table
480d1f29560f -> f103f27c4575, Change pennant to identifier and make required
cd4a4f059f9e -> 480d1f29560f, Add priority field to Nationality
07e4b725c547 -> cd4a4f059f9e, Add foreign key for GeomSubType.parent
e752bda39400 -> 07e4b725c547, Switch to foreign keys for fields in Contacts
6c1e62ee267f -> e752bda39400, Rename reference table cols to name
ccc37f794db6 -> 6c1e62ee267f, Switch to UUIDs for PKs for SQLite
96aa1196c23b -> ccc37f794db6, update, add constraints
7df9dcbd47e7 -> 96aa1196c23b, Add uniqueness constraint to all names in reference tables
bcff0ccb4fbd -> 7df9dcbd47e7, base2
<base> -> bcff0ccb4fbd, base

> alembic show head # it shouldn't print anything but it shouldn't fail, either.

> alembic -c alembic_sqlite.ini show head
# Expected output
(pepys-import) baris@bariss-MacBook-Pro pepys-import % alembic -c alembic_sqlite.ini show head
Rev: b03fb305981b (head)
Parent: f103f27c4575
Path: /Users/baris/Documents/PycharmProjects/pepys-import/migrations/sqlite_versions/2020-06-05_b03fb305981b_add_level_attribute_to_privacies_table.py

    Add level attribute to Privacies Table
    
    Revision ID: b03fb305981b
    Revises: f103f27c4575
    Create Date: 2020-06-05 12:43:44.318254

> alembic -c alembic_postgres.ini show head
# Expected output
(pepys-import) baris@bariss-MacBook-Pro pepys-import % alembic -c alembic_postgres.ini show head 
Rev: d96db9a0620f (head)
Parent: 30194fdfc033
Path: /Users/baris/Documents/PycharmProjects/pepys-import/migrations/postgres_versions/2020-06-05_d96db9a0620f_add_level_attribute_to_privacies_table.py

    Add level attribute to Privacies Table
    
    Revision ID: d96db9a0620f
    Revises: 30194fdfc033
    Create Date: 2020-06-05 12:42:57.737339


> alembic upgrade head  # Upgrading the database to the latest version
> alembic downgrade bcf # Downgrading the database to the first version
    
# Run the following commands when your config file points to the PostgreSQL database
# Postgres migration scripts are not changed. But it is worth to check if there is any error
> alembic upgrade 515:d96 --sql
> alembic downgrade d96:515 --sql

```
Checking the changes in `Database Migration` section would also help you to understand why it's necessary to use different config files: https://pepys-import--427.org.readthedocs.build/en/427/database_migration.html#helpful-commands (There are also new notes in `How to use it` section)


migrations/README.rst
```diff
@@ -168,7 +168,24 @@ For doing that the following command might be used:
...
-| If you would like to use relative identifiers, such as :code:`alembic downgrade head -1`, you might check it out: `Relative Identifiers <https://alembic.sqlalchemy.org/en/latest/tutorial.html#relative-migration-identifiers>`_
+| If you would like to use relative identifiers, such as :code:`alembic downgrade -1`, you might check it out: `Relative Identifiers <https://alembic.sqlalchemy.org/en/latest/tutorial.html#relative-migration-identifiers>`_
+|
+| **Note-4:** During the migration of SQLite Database, it's possible to see this error:
+
+.. code-block:: bash
+
+    AddGeometryColumn() error: "UNIQUE constraint failed: geometry_columns.f_table_name, geometry_columns.f_geometry_column"
+
+You can ignore this error because it says that there is already a geometry column entity for your table that has a Geometry column (i.e. States, Geometries, Contact). This error happens because SQLite doesn't support *ALTER TABLE* statement.
```
robintw, line 179: If we see this error, does the rest of the migration continue correctly? I'd have thought it'd have stopped part-way through, potentially leaving the database in an unclean state?

migrations/README.rst
```diff
@@ -168,7 +168,24 @@ For doing that the following command might be used:
...
+Instead, it creates a new one, copies values from the previous table, and drops the existing table. However, during the creation of a new table, it tries to push the Geometry column to the :code:`geometry_columns` table again.
+
+| **Note-5:** If you change a column's name, you should implement different classes for :code:`upgrade()` and :code:`downgrade()` methods. Otherwise, this class breaks your upgrade because it doesn't have that column to rename.
+|
+| An example can be found in :code:`migrations/sqlite_versions/2020-06-03_f103f27c4575_change_pennant_to_identifier_and_make_.py`. What you should do is as follows:
+
+- :code:`copy_from` post-hook will create all necessary tables. Copy the table that has a column change to :code:`upgrade()` and :code:`downgrade()` methods.
+- Remove the global class.
```
robintw, line 187: I'm rather confused about Note 5. I don't understand why we have to copy a definition of the table in there for this, and I'm not sure what the copy_from post-hook will do, or what it means to remove the global class (presumably not delete it from `sqlite_db.py`?). From looking at the copy_from post-hook, it looks like it will copy the table definitions there automatically - is that right? So if so, does the developer who changes a column name need to do anything with this copying, or is this all 'just for information'?

Could you provide some more details, or possibly discuss it with me on Slack sometime?

migrations/README.rst
```diff
@@ -227,6 +255,27 @@ If you face this error, it means that the :code:`pepys-import` repository should
 
 ------------
 
+.. code-block:: bash
+
+    Database tables are not found! (Hint: Did you initialise the DataStore?)
+
+If you migrated your **SQLite** database and started to see this message, it would have been because of the extra tables created by Alembic. When a table with Geometry column is included in migration,
+please check the SQLite database if there are any temporary tables (with *tmp* keyword) such as :code:`idx_Geometries_tmp_geometry_node` etc. If there is any, please drop these tables. After that,
+please open your migration script and add :code:`spatial_index=False` argument to the Geometry column. An example is as follows:
```
robintw, line 264: Is this something that the developer does when creating the migration script? Or something that the end user might need to do? I'd rather not have the end user having to list and drop tables in the SQLite database.

Can we just set spatial_index=False all the time?

migrations/README.rst
```diff
@@ -227,6 +255,27 @@ If you face this error, it means that the :code:`pepys-import` repository should
...
+
+.. code-block:: python
+
+    # Change this column
+    geometry = deferred(Column(Geometry(geometry_type="GEOMETRY", management=True), nullable=False))
+    # Add spatial_index
```
robintw, line 270: Hang on, this has confused me a bit more - we're saying that we're adding a spatial index here, but we're setting `spatial_index=False`?

migrations/README.rst
BarisSari: Yes, it doesn't affect the `geometry` column because it means that there is already a column (which is created in the previous migrations). If you migrate your database in this PR, you will see that this error is thrown for a few migrations but it doesn't change anything. 

I know that it sounds inconvenient. When I searched for it, I found out that geometry columns accept `management` parameter. If it's true, it should automatically drop this value: (https://geoalchemy-2.readthedocs.io/en/latest/spatialite_tutorial.html#declare-a-mapping)

`Setting management to True indicates that the AddGeometryColumn and DiscardGeometryColumn management functions will be used for the creation and removal of the geometry column. This is required with SpatiaLite.`

In our columns, we already set this parameter to `True` but Alembic still doesn't drop it when a table with geometry column is dropped. I don't have any idea why it's happening.

migrations/README.rst
BarisSari: I think my explanation is a bit confusing. What I meant is as follows: Post-hook copies the correct tables to the migration script, but it is not enough to have these classes. Let me explain it in a code snippet:

```python
class Platform(BaseSpatiaLite, PlatformMixin):
    identifier = deferred(Column(String(10), nullable=False))

def upgrade():
        with op.batch_alter_table("Platforms", schema=None, copy_from=Platform.__table__) as batch_op:
            batch_op.alter_column(column_name="pennant", new_column_name="identifier", nullable=False)

def downgrade():
        with op.batch_alter_table("Platforms", schema=None, copy_from=Platform.__table__) as batch_op:
            batch_op.alter_column(column_name="identifier", new_column_name="pennant", nullable=True)
```

As you can see that, in `upgrade()` function, Alembic tries to drop a column named "**pennant**". On the other hand, in `downgrade()` function, it tries to drop a column named "**identifier**". So, if you don't implement a different class for `upgrade()`, it is going to fail because it will not find a column named "**pennant**" (it takes it from `copy_from` parameter). The correct implementation should be like this:
```python
def upgrade():
        class Platform(BaseSpatiaLite, PlatformMixin):
            pennant = deferred(Column(String(10), nullable=False))
        with op.batch_alter_table("Platforms", schema=None, copy_from=Platform.__table__) as batch_op:
            batch_op.alter_column(column_name="pennant", new_column_name="identifier", nullable=False)

def downgrade():
        class Platform(BaseSpatiaLite, PlatformMixin):
            identifier = deferred(Column(String(10), nullable=False))
        with op.batch_alter_table("Platforms", schema=None, copy_from=Platform.__table__) as batch_op:
            batch_op.alter_column(column_name="identifier", new_column_name="pennant", nullable=True)
```

If it's still not clear, we can discuss it on Slack. I'll edit the documentation after we agreed on a good explanation.

migrations/README.rst
BarisSari: This problem might occur only if there is a wrongly implemented migration script. So, it's something for developers. If they see that there are extra redundant tables, it points out that new migration script has something wrong.

> Can we just set spatial_index=False all the time?
I thought the same but I think it wouldn't work when we tried to use `BaseSpatiaLite.metadata.create_all()` because it should be created for once. (The first file in migration scripts does this job if migration is used to create all tables)

migrations/README.rst
BarisSari: It's a typo, sorry about that. I meant to add `spatial_index=False` parameter.

BarisSari: Fixed:
https://github.com/debrief/pepys-import/pull/427/commits/a68109ce663f06ef89f03955c8773059cb479741

migrations/README.rst
robintw: Ok, that's fair enough.

migrations/README.rst
robintw: Ok, could you clarify in the docs here that this is for developers only, and shows that the migration script is wrong.

migrations/README.rst
robintw: I'm still a bit confused about this: surely if we define a class inside the function with just one member variable, then it'll think the class only has one field? Or is this just a simple example with only one member variable?

We should probably discuss this on Slack once you're back.

migrations/README.rst
BarisSari: Done:
https://github.com/debrief/pepys-import/pull/427/commits/56fb60e75509ae1a0c815f32667b0062e02e9d02

migrations/README.rst
BarisSari: I'm sorry that I've seen this message now. It was just a simple example to show you the difference.

IanMayo:
Have added bounty to the Stack Overflow question related to this issue: https://stackoverflow.com/questions/62775640/alembic-generates-arbitrary-type-changes-for-geometry-columns

IanMayo:
@BarisSari - could you fix the conflict please?

(it does look quite simple)

BarisSari:
> @BarisSari - could you fix the conflict please?
> 
> (it does look quite simple)

Sure, @IanMayo. However, this PR is not ready yet. I hope it will be in the next couple of days.

BarisSari:
I've made a thorough search about why we face many problems related to SQLite. I added a new section named _"Why do we have many SQLite related problems ?"_ to the [documentation](https://pepys-import--427.org.readthedocs.build/en/427/database_migration.html) (at the bottom).

Even though I solved most of the problems for SQLite, GeoAlchemy2 is also problematic and there are still two problems:
1. Arbitrary _nullability_ change for `Geometry` Columns
2. Error that is thrown during the migration: `AddGeometryColumn() error: "UNIQUE constraint failed: geometry_columns.f_table_name, geometry_columns.f_geometry_column"`. This error doesn't break anything but it is still out there. What it tries is adding the `geometry` column of the table that changed to the `geometry_columns` table. But there is already a row and it causes an error. For example, let's say we added a new column to the Media table. Previously, a row created in the `geometry_column` table (`f_table_name='Media', f_geometry_colun='geometry', ...`). Since a temporary Media table is created first, and this table should have the geometry column, it calls `AddGeometryColumn()` function and this error is thrown. Because the old Media table still exists. It will be removed after its data is inserted into the new table. 

If that's okay for both of you, @robintw and @IanMayo , I suggest reviewing and merging this PR at the moment. I'll create issues in Alembic's repository or on StackOverflow. But I don't want this PR to wait any longer. I can open hotfix PR's when/if I get any answers.

BarisSari:
Hello @IanMayo . Robin and I discussed the problems on Slack and he suggested writing the advantages and disadvantages of making `--sql `parameter work so that you might tell us your ideas whether that's necessary or not.

**Advantages**
- It might be useful when something fails. For example, let’s say one migration failed and we don’t know what it did exactly. We might run `downgrade --sql` to generate a rollback query, and run it in the database to rollback. 
- When we’re not sure what a migration script will upgrade, we might again generate a query first to examine if it will cause any trouble.
- For a developer who is not familiar with Alembic, it might be difficult to understand the migration script. Therefore, he/she might generate SQL scripts and check if the migration script is correct or not.

**Disadvantages**
`copy_from` parameter should be passed to each `Operations.batch_alter_table()` for SQLite. In a migration script, we have this operation for each table that changed. Otherwise, the changed tables aren't reflected and Alembic throws an error. Here is related information from Alembic's documentation:

> In the preceding sections, we've seen how much of an emphasis the "move and copy" process has on using reflection in order to know the structure of the table that is to be copied.
This means that in the typical case, "online" mode, where a live database connection is present so that Operations.batch_alter_table() can reflect the table from the database,
is required; the --sql flag cannot be used without extra steps.

> To support offline mode, the system must work without table reflection present, which means the full table as it intends to be created must be passed to Operations.batch_alter_table() using copy_from:

You might say that it doesn't sound like a disadvantage. But it means that we should pass all necessary tables, foreign key tables, imports, etc. to each migration script. That's why I wrote a post-hook script that automatically passes these values from `sqlite_db.py`. So, at least, we don't need to manually copy classes and imports.

- When we pass `copy_from` to each operation:
1. It does create temporary index tables for Geometry columns if `spatial_index=False` is not passed to geometry columns, which might break our `is_schema_created()` function (it basically checks the number of tables in the database). 
My post-hook (`copy_from`) also passes `spatial_index=False` parameter automatically to geometry columns.
2. It throws the error mentioned in the previous message (`AddGeometryColumn() error: ...`). By the way, this error will be seen by the client when the migration option is run in Pepys-Admin.
3. If post-hook somehow stops working, migration script will fail at least 90%. Because there will be missing imports, missing classes, etc.

What do you think, @IanMayo ?

IanMayo:
@BarisSari - do you know why the performance test failed?   Is it due to new constraints?

BarisSari:
> @BarisSari - do you know why the performance test failed? Is it due to new constraints?

No, I don't know why it failed, @IanMayo . I don't think that my changes affect the test's performance. There might be another reason.

robintw:
I've found some of the benchmarking tests to be a bit flaky recently: I think Travis have had quite a lot of load on their servers, and it's been slowing down some of the tests.

I think it's fine to increase the benchmark threshold by a bit to allow a bit of flexibility.

IanMayo:
Thanks @robintw - do you have spare time to relax the performance test for this PR?

IanMayo:
@robintw  has explained that the root of this issue is that supporting `--sql` could bring some benefits, but adding that support could be challenging.

Based on that, I'm don't think we need to support `--sql`, because:
* client don't have DB Admins that only understand SQL. They do have python devs (though none that are currently familiar with Alembic)
* the upgrade will always happen from `PepysAdmin` - so, Alembic will always be available
* it's (hopefully) unlikely that we will need to roll back a migration via SQL.  But, if we really need to, I propose that we would go through the chore of crafting the relevant SQL
* in time, the database schema should mature - particularly as we implement functionality to "drive" the areas that don't get touched much right now.

robintw:
@IanMayo Performance threshold now updated.

IanMayo:
I know @robintw has been quite involved in this PR already, but I'll wait for his approval before I merge it.

migrations/README.rst
```diff
@@ -242,4 +253,44 @@ You should upgrade your DB and then run the revision command:
...
+
+------------
+
+**FOR DEVELOPERS ONLY**
+
+.. code-block:: bash
+
+    Database tables are not found! (Hint: Did you initialise the DataStore?)
+
+If you migrated your **SQLite** database and started to see this message, it might mean that your migration script it wrong.
```
robintw, line 266: Very minor typo here - should be "is" rather than "it"

BarisSari: Thanks, @robintw! Typo fixed: https://github.com/debrief/pepys-import/pull/427/commits/98e6d4a4e9beb19f098d78951309cb36a1de3c04

PR #429, opened 2020-06-02
Add CSV import docs and minor change to import to print path

### LINKED ISSUE #423: Document use of CSV seed file path
In #419 @BarisSari identified the `--path` parameter for PepysAdmin

It is documented here: https://pepys-import.readthedocs.io/en/latest/usage.html#pepys-admin

I think we could do with some more detail there.  Importing the `csv` seed data will be performed by a member of clerical staff.

So, I believe we need to be more explicit in opening the Command Prompt, navigating to the Pepys folder, specifying the full command that is to be typed in, then some coverage of how to navigate the Admin UI to:
* clear existing metadata/reference data that may have been created during testing
* load the reference data/metadata

robintw:
Fixes #423.

I've tried to make the instructions fairly simple - can you check and see whether they are suitable for the target audience @IanMayo?

Also includes a minor change to the import procedure to print out the path that is being used to import from.


IanMayo:
Thanks @robintw 

I think the "clear schema" will not need to be run on every occasion.

* they may not have all tables ready at the start, so may wish to return to process them again at a later date
* they may think of a new `category` of synonym in the future, and wish to load them into the database
* they may decide to enter all the platforms of another nation in one go
* once they start working with `Tasks` and `Participations`, they may collate a list of all the small elements (serials) within an exercise, and load them just before the exercise happens

My understanding is that rows in the `seed-file` that are already present will be silently skipped, and only new rows added.

Is that correct?

If the above is correct, I suggest the guidance indicates that `Clear schema` is an optional step, and not necessary when new rows are being added.

robintw:
All rows that are present _with exactly the same values as an existing record_ will be skipped, all others will be added.

I'll rephrase it to try and make this clearer.

IanMayo:
@Robin - I think we can reconsider the structure of the steps in the guidance here.

We run `pepys-admin` twice, but it's only at step 4 where we inform the user that they may not need to clear the database - thus they didn't need to go into `pepys_admin`, and 3 previous steps have been unnecessary.

Instead, if we just run `pepys-admin` once (specifying the `--path` parameter), then we reduce the number of steps that aren't performed to one.

robintw:
I've updated the instructions now - let me know what you think @IanMayo.

PR #433, opened 2020-06-03
Hotfix - Stamp Latest Migration Automatically

### LINKED ISSUE #398: Better way to stamp latest migration when creating schema in DataStore
At the moment there are two ways of creating the database structure: either using `alembic upgrade head` (which would upgrade a blank database to the latest revision), or using the SQLAlchemy commands in the DataStore class to initialise based on the SQLAlchemy table definitions.

In the latter case we also manually 'stamp' the latest alembic version into the database, so that Alembic knows what version the database is at, and can then apply new revisions later.

The problem is that we keep forgetting to update this bit of code when we create a new alembic migration - and that means Alembic thinks the database is out of date when it isn't.

There are various potential solutions:

1. Just get better at remembering to do this!

2. Change to using alembic migrations to create the database structure in DataStore, then the revision will be stamped automatically. We decided not to do this before as it would lead to an automatic alembic upgrade, which may potentially lose data. However, we could do an automatic upgrade if - and only if - there is no schema in the database already (ie. it is blank).

3. Create some sort of automated tool to update our source code on commit, to ensure that the latest revision ID from the alembic migrations folder is included in the source code in the right place.

I think number 2 is possibly best, but welcome discussion from @BarisSari and @IanMayo.

BarisSari:
Fixes #398.


BarisSari:
Hello @IanMayo and @robintw . Even though I changed the `DataStore.initialise()` method, I'm still not quite sure if this is the right thing to do. Because Alembic slows down our tests and I believe that it leaves some database connections open (we're already using `with` context to create a connection in `migrations/env.py`). I debugged the tests many times to figure out the problem but unfortunately, I couldn't find out. If I don't dispose some of the created engines, after %80 tests pass, python.exe stops working. It throws max connection error.

There are a few suggestions on Alembic which I already tried:
https://alembic.sqlalchemy.org/en/latest/cookbook.html#building-an-up-to-date-database-from-scratch
https://alembic.sqlalchemy.org/en/latest/cookbook.html#connection-sharing

What would you prefer me to do?

robintw:
Hmm, I'm not sure what the best way forward is. Those problems do sound like it means this particular solution won't work for us.

Is there some way of adding a post-processing task when creating a migration that automatically updates the stamped migration in our Python file? I could see us creating a file containing just:

```
LATEST_POSTGRES_VERSION = "blah"
LATEST_SQLITE_VERSION = "blah"
```

and importing that file in data_store, to use those values to initialise the alembic version table. We might then be able to get a post-processing script (like the one that adds the copy table command) to modify that file when alembic creates a revision? What do you think?

BarisSari:
> Hmm, I'm not sure what the best way forward is. Those problems do sound like it means this particular solution won't work for us.
> 
> Is there some way of adding a post-processing task when creating a migration that automatically updates the stamped migration in our Python file? I could see us creating a file containing just:
> 
> ```
> LATEST_POSTGRES_VERSION = "blah"
> LATEST_SQLITE_VERSION = "blah"
> ```
> 
> and importing that file in data_store, to use those values to initialise the alembic version table. We might then be able to get a post-processing script (like the one that adds the copy table command) to modify that file when alembic creates a revision? What do you think?

It's good thinking! I hope that this approach will be much easier and won't affect performances.

BarisSari:
This PR is ready for review, @robintw and @IanMayo .

PR #435, opened 2020-06-03
Rename pennant to identifier and make required field

### LINKED ISSUE #432: Rename `pennant` to `identifier`
After some walking through of potential datasets with the client, we've identified the need to rename `pennant number` as `identifier`.

The database will store helicopters as platforms. Helicopters don't have a pennant number, but they do have a unique number painted on their tail.

Similarly a warship may have two "ship's boats" .  These can have a name, but no pennant number.  We discussed using their build-date as identifier.  So this would still be unique:
* Name:  Neptune
* Type: Sea-Boat
* ID: Jan_2020
This would also be unique, if another ship had a sea-boat called Neptune:
* Name:  Neptune
* Type: Sea-Boat
* ID: Mar_2005

So, we keep `Name`,`Platform-Type`, `Identifier` as a unique combination.

**Note:** the above is part of a working practice/business process that an organisation using Pepys will have to develop, it's not part of the software.

robintw:
This PR renames the `Platform.pennant` field to `Platform.identifier`, and makes this a required field.

This includes:
 - Changing the database definitions for SQLite and Postgres
 - Adding a migration for the changes, for both SQLite and Postgres
 - Updating the default data CSVs to include an identifier for each platform
 - Updating the UI for the command-line resolver, to ask for the identifier with all the other required fields
 - Updating tests to deal with the change to the UI
 - Updating tests that manually create a Platform to give an identifier to this platform

Fixes #432


robintw:
This needs merging before #436, as I'll need to do some manual alteration to that PR to accommodate the changes in this one.

PR #440, opened 2020-06-04
Remove support for synonyms for Sensors

### LINKED ISSUE #387: Synonyms don't work correctly for Sensors (and possibly other tables?)
As quoted from Ian in #379,

"GPS-B" may be the synonym for "GPS-2" on "HMS DRAGON". But, "GPS-B" may be the synonym for another sensor on another platform. So, maybe there are some tables we can't support synonyms for. I'll have to give this more thought.

This means we could run into issues using synonyms at the moment, as they may end up referring to the wrong sensor on the wrong platform.

robintw:
This PR fixes #387 by removing support for Sensor synonyms

 - Blacklist the Sensors table in the `add_to_synoynms` function
 - Remove searching for synonyms in `get_sensor`
 - Remove options to add synonyms in the command-line resolver
 - Update tests


pepys_import/core/store/data_store.py
```diff
@@ -414,6 +414,10 @@ def add_to_platforms(
         return platform_obj
 
     def add_to_synonyms(self, table, name, entity, change_id):
+        # Blacklist certain tables, and don't Synonyms for them be created
+        if table in ["Sensors", "GeometrySubTypes"]:
```
BarisSari, line 418: Can you use variables from `constants` here? I think we already import it.

robintw: Thanks Baris - good catch. Fixed now.

PR #444, opened 2020-06-05
Add ordering for privacy

### LINKED ISSUE #438: Add ordering for Privacy
When merging databases, we need to use the 'highest' privacy of the two entries when merging entries - and this requires knowing how privacies relate to each other.

We decided this was going to be a `Level` column added to the Privacy table. We'd then need to modify the code that displays privacies in the command-line resolver to display them in order, the code that adds new privacies to display the privacies first to see their levels, and then allow a level to be entered too, plus the code that imports Privacies from a CSV file to import levels too.

BarisSari:
Fixes #438


pepys_import/resolvers/command_line_resolver.py
```diff
@@ -192,11 +192,12 @@ def resolve_reference(
                 .all()
             )
         elif db_class.__tablename__ == constants.PRIVACY:
-            objects = data_store.session.query(db_class).order_by(desc(db_class.level)).all()
```
IanMayo: This is very minor @BarisSari , but wouldn't it be (very, very, very slightly) more efficient to use `.limit(7)`, than to use `.all()` then trim the first 7 values?

BarisSari: Yes, it's more efficient but I didn't want to fetch the table again for showing it in 'add a new privacy' option. That's why I fetched all values and added the first 7 values to the options. If a user selects 'add a new privacy', it then shows all values in the table without sending another query to the database.

IanMayo: Aah yes, very good :-)

Issue #394, opened 2020-05-20
Importing metadata/reference data - skip duplicates


IanMayo:
Soon the clients are going to start populating the reference and metadata tables.

This may be an iterative process, as their understanding of how to organise the data matures.

`PepysAdmin` includes commands to `import metadata` and `import reference data`.

The clients should be able to run these repeatedly with `.csv` files that evolve as they add new values.  So, when populating these tables, we should skip rows that exist already.

Actually, duplicate entries should break our uniqueness tests.  But, when doing the csv import, pepys-admin should just skip them, rather than fail.

robintw:
I've investigated this, and conveniently we already skip values that exist. However, there are a couple of other issues that are worth noting at this stage, even if they don't actually need fixing now:

1. It matters what order the CSV files are imported in. For example, the reference data must be imported first, as the metadata refers to the reference data (eg. a Platform entry refers to a PlatformType entry - and so the PlatformType entry has to be imported first). We can give the user instructions to run the `import reference data` first, followed by `import metadata` - but it might be worth adding a combined `import reference & metadata` command which does it in the right order? (And obviously keep the individual import commands, but make it clear that they import _only_ the reference/metadata)

2. It also matters what order the metadata CSV files are processed in. For example, a Sensor entry refers to a Platform, so the Platform CSV file must be imported before the Sensor CSV file. This works at the moment, because the list of files is sorted alphabetically, and P comes before S. However, this could be a bit fragile and may be worth coming back to later to do in a more robust way. I've had a look at the [schema diagram](https://docs.google.com/drawings/d/1XIJ9t3Lj2a0EXEFogNWmtAP_bNPCx-b2aLlnn9kmzzA/edit) and there are some other relationships that won't work with this alphabetical sorting. For example, Tasks is needed before Participations, but they will be imported in the wrong order if we go alphabetically. However, as far as I'm aware we're not actually using those tables at the moment - so it shouldn't be a problem for the moment.

3. Searching for foreign keys in the CSV files are done only by name at the moment. For example, the Sensor CSV file refers to the host Platform as just a platform name - which is then used to search for the platform. This will need changing when we fix #362, and will be a bit of a pain to fix as I imagine it will require us to specify lots more platform details in the Sensors CSV file. This could do with some more thought I think.

IanMayo:
**Order that files are read in**: Yes, there is also a single dependency in `Reference` tables.  `GeometryTypes` must be imported before `GeometrySubTypes`.

[...]

PR #449, opened 2020-06-08
Hotfix - Don't allow duplicate entries to be added for Sensor, Platform, Datafile, Synonym

robintw:
This PR stops duplicate entries being added for Sensor, Platform, Datafile and Synonym. This is particularly useful when importing from CSV files (see #394).

I was _sure_ this had already been implemented, but tests today for something else revealed that it was only implemented for the reference tables but _not_ for the metadata tables.

This implements this for the metadata tables - it just returns the existing entry rather than adding an entry if an entry already exists with the same details.

### DIFF
--- a/pepys_import/core/store/data_store.py
+++ b/pepys_import/core/store/data_store.py
@@ -280,6 +280,22 @@ class DataStore:
         elif privacy is None:
             raise MissingDataException("Privacy is missing/invalid")
 
+        # Check if entry already exists with these details, and if so, just return it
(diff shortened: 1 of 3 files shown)


robintw:
This is ready for review now: the reason for the low coverage is because we have a few `assert` statements in there that we can't test because they're checking for 'impossible' situations (eg. that the database has duplicate entries that defy the database's own unique constraints). I think the asserts are worth having to cover an insane worst-case scenario, but I don't think we can test them.

PR #451, opened 2020-06-08
Improve Display of Alembic Version

### LINKED ISSUE #447: Improve display of Alembic version
image

I think the Create Date field above gives me what I’m looking for.  The database connected to in the above instance was last updated on 3rd June.

But, when I tried the above using an old database (the remote PostGres one at anywheredata.com), `2. View status` failed while collating the table data, since in trying to create the summary tables, it referred to a missing database field.   This failure stopped the `alembic` status from being displayed.

Solutions?
* Separate `View status` into `View summary of contents` and `View summary of database version`
* Catch the database listing error, print the error, then continue to show the alembic status 

Any other thoughts/strategies?

BarisSari:
Fixes #447 

If DB is not up-to-date, `status` option prints the following:

image


BarisSari:
@robintw , benchmark test for short file started to fail because mean is about 0.30-0.32s. As you might see my changes, I don't think that I did anything to affect the performance of file import. Do you have any idea what could be the reason of failing?

robintw:
@BarisSari I found in some of my tests recently that it was a bit flaky - I think Travis might have changed the specs of the servers the tests are being run on, so they're taking a little bit longer and sometimes going across the threshold.

Sometimes I've had success with just restarting the build, and the next time it's been ok - otherwise just increase the threshold to something like 0.4.

BarisSari:
> @BarisSari I found in some of my tests recently that it was a bit flaky - I think Travis might have changed the specs of the servers the tests are being run on, so they're taking a little bit longer and sometimes going across the threshold.
> 
> Sometimes I've had success with just restarting the build, and the next time it's been ok - otherwise just increase the threshold to something like 0.4.

I ran 3 times and all of them failed. I increased it to 0.4.

PR #466, opened 2020-06-14
Try to connect to the database at the beginning

BarisSari:
When a DataStore object is created, it creates a connection string and an engine that is going to connect to the created string. However, it works lazily and it doesn't try to connect to the database URL if `engine.connect()` is not called. However, when the welcome text and the status are printed, it seems like DataStore is connected to the given database, which is misleading. This PR solves this problem.

### DIFF
--- a/tests/test_data_store_api_spatialite.py
+++ b/tests/test_data_store_api_spatialite.py
@@ -779,6 +779,7 @@ class FirstConnectionTestCase(TestCase):
         assert "ERROR: SQL error when communicating with database" in output
         assert "Please check your database file and the config file's database section." in output
         assert "Current database URL: 'sqlite+pysqlite://:@:0/" in output
+        assert "__init__.py" in output
 
 
 if __name__ == "__main__":
--- a/tests/test_data_store_api_spatialite.py
+++ b/tests/test_data_store_api_spatialite.py
@@ -1,6 +1,8 @@
 import os
 import unittest
+from contextlib import redirect_stdout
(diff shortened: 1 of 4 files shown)


tests/test_data_store_api_spatialite.py
```diff
@@ -760,5 +762,24 @@ def test_create_invalid_synonym(self):
...
+                db_username="",
+                db_password="",
+                db_port=0,
+                db_name=db_name,  # Give a file that is not a database
+                db_type="sqlite",
+            )
+        output = temp_output.getvalue()
+        assert "ERROR: SQL error when communicating with database" in output
+        assert "Please check your database file and the config file's database section." in output
+        assert "Current database URL: 'sqlite+pysqlite://:@:0/" in output
```
robintw, line 781: Earlier we set the db_name to an `__init__.py` file - why is the current database URL not showing that name here?

BarisSari: The absolute path differs on local machines and on Travis. That's why I didn't include it, but I think it's better if I add one more assert which checks "__init__.py" string

BarisSari: Added:
https://github.com/debrief/pepys-import/pull/466/commits/4884893c9ba67bf0bd5b81b5f4531d979c36264f

BarisSari: Added:
4884893

PR #468, opened 2020-06-15
Add EAG importer

### LINKED ISSUE #464: Support for EAG files
For the trip next week, the client have learned about a file format that they can expect to receive.

We have been given a sample data-file, and a description of the structure. Significantly, the file format is in ECEF format: https://www.oc.nps.edu/oc2902w/coord/coord.pdf

Unfortunately we can't put the sample file in this repo, and possibly not the file description.

These will be passed directly to @robintw 

Once the importer is complete, we can make a new release, and the clients can test it against their other samples.  If it's successful, we can convert an existing Pepys track to EAG format, include that file in our sample data, and write/run unit tests against it.

I've been able to make the following assumptions on the file format
1. The filename will end in `.eag.txt`
2. The filename will start with the date of the recording, then an underscore, then the `callsign` for the aircraft.  This callsign will be used as a synonym for the aircraft. 
3. Column A is the GPS time since the previous Sunday (ms)
4. Columns B, C, D can be ignored
5. Columns E, F, G represent the ECEF for x,y,z in metres
6. Columns H and I can be ignored
7. Column J is the heading (in degrees)
8. Column K is the GPS time of day (derived from column A)

robintw:
Fixes #464.

This adds an importer for EAG files, along with a sample EAG file (created manually, rather than being a real file from the client) plus tests.


importers/eag_importer.py
```diff
@@ -0,0 +1,205 @@
...
+
+    def convert_ecef_to_location_and_height(self, ecef_x_token, ecef_y_token, ecef_z_token):
+        """Converts Earth-Centred, Earth-Fixed X, Y and Z co-ordinates into a Location object containing
+        latitude and longitude, plus a height value in metres.
+
+        X, Y and Z should be provided in metres.
+        
+        All formulae taken from https://microem.ru/files/2012/08/GPS.G1-X-00006.pdf
+        """
+        ecef_x = float(ecef_x_token.text)
```
BarisSari, line 171: Is it possible that any of X, Y, or Z coordinates are wrong? I don't know how robust it this file but float conversion might fail if any of them is a not convertible string.

robintw: You're right - I thought I had a try-catch block around that, but I didn't. I'll fix it now.

robintw: Fixed now in 58cfa2b

robintw:
Thanks for pointing out the issue with the float conversion - that's fixed now, are you happy to approve this @BarisSari?

importers/eag_importer.py
```diff
@@ -0,0 +1,213 @@
...
+            # Calculate the timestamp, using the time since last Sunday and the calculated last Sunday date
+            timestamp = self.calculate_timestamp(time_since_sun_ms, last_sun_date)
+            time_since_sun_ms_token.record(self.name, "timestamp", timestamp)
+
+            # Platform is based on the callsign - the user will link this as a synonym to a defined Platform
+            platform = data_store.get_platform(platform_name=callsign, change_id=change_id,)
+
+            # Sensor Type is always fixed to GPS
+            sensor_type = data_store.add_to_sensor_types("GPS", change_id=change_id).name
+            sensor = platform.get_sensor(
```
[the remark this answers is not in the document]
robintw: Leaving this field blank means that it will be requested through the resolver - so in normal operation the user will be asked to fill it in. However, the testing I did for that has thrown up another couple of issues which I'll fix in a few new commits.

robintw:
I've fixed a couple of issues I found - and have tested that it asks for a sensor name correctly. I'll create a separate issue about enforcing fields are non-null and non-empty (which are different things).

I think this is ready to be merged now.

IanMayo:
Blank sensor name.  Aah, of course.  Duh.

PR #470, opened 2020-06-15
Offer first 7 objects when resolving a reference data

### LINKED ISSUE #469: Platform-Types and Sensor-Types not offered in CLI
This morning I worked with the analyst to load some sample data.

We worked with the file called `sen_tracks.rep`.

The CLI ran as expected, but she was not offered the existing items for `sensor-type` or `platform-type`.  We opened pepys-Admin and verified that both tables contained values - which they did.  So, she used `Search for platform-type`, which worked fine, and `Search for sensor-type`, which also worked ok.

I see that these tables don't have "priority" column.  I do think they both contain more than 7 entries. Maybe that's the issue.  Maybe we only provide the choices when there are 7 or less.

It may be better to offer the first 7 in this instances.  In the longer term we should probably introduce a "priority" table, to let the analysts select the 7 most likely values in these types tables.

BarisSari:
Fixes #469

### DIFF
--- a/pepys_import/resolvers/command_line_resolver.py
+++ b/pepys_import/resolvers/command_line_resolver.py
@@ -204,7 +204,7 @@ class CommandLineResolver(DataResolver):
             )
             current_values += "\n"
         else:
-            objects = data_store.session.query(db_class).all()
+            objects = data_store.session.query(db_class).limit(7).all()
         objects_dict = {obj.name: obj for obj in objects}
         # CamelCase table names should be split into words, separated by "-" and converted to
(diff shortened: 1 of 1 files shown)


IanMayo:
Tested this by extending platform types to beyond 7 entries:
image

CLI successfully trimmed to first 7 entries:
image

But, search for platform-type allowed selection of value that wasn't in the first 7:
image

BarisSari:
Thanks for quick experimentation, @IanMayo . I opened the page for asking your review but you already did it! 🙂

PR #471, opened 2020-06-15
 Reduce risk of accidental deletion

### LINKED ISSUE #430: Reduce risk of accidental deletion
When writing the instructions for importing CSV reference data earlier, and including a step to delete the database schema, it struck me that it would be very easy for a user to accidentally delete the entire database schema or the entire database contents by selecting the wrong option in Pepys Admin.

A few notes/comments on this:

1. I remember something about different users connecting with different config files which have different permissions. Does this mean that most users don't have the ability to delete the schema? Would they, however, have permissions to delete the entire contents (which is one of the other options in Pepys Admin).

2. Is it worth adding a 'Are you sure?' prompt to Pepys Admin for the destructive operations? These would be:
 - `Clear database contents`
 - `Clear database schema`
 - `Migrate` (as it could potentially cause data loss - and definitely changes the schema)

3. Do the client have backup plans for their Postgres instance? Are they automated and tested properly. I'd hate for them to lose everything because of either a mistake by one of their staff, or (worse) a programming error by one of us.

Thoughts @IanMayo?

BarisSari:
Fixes #430


IanMayo:
I'll make a couple of minor changes to the wording.

BarisSari:
> I'll make a couple of minor changes to the wording.

Sure, @IanMayo .

PR #473, opened 2020-06-15
Forbid empty string as value for fields that are used for uniqueness checks

### LINKED ISSUE #472: Ensure that fields required for uniqueness are non-null and non-empty
We should ensure that fields required for uniqueness checks (eg. name, identifier, nationality for Platform) cannot be null and cannot be empty (ie an empty string).

All of these fields have been checked and they all have `nullable=False` set in their definition, which means they cannot be set to NULL. However, they can still be set to `""` - this needs further investigation.

robintw:
Fixes #472.

This ensures that empty strings cannot be set as the value for String fields that are used for uniqueness checks (like `name` in Platforms and `hash` in Datafiles). This is done in two ways:

1. By checking that the values are not empty when the user inputs them in the command-line resolver
2. By adding a CHECK constraint to the database to ensure that the database will not accept empty strings for these fields


robintw:
When you're back from vacation (no rush), would you be happy to put together these migration scripts @BarisSari?

BarisSari:
> When you're back from vacation (no rush), would you be happy to put together these migration scripts @BarisSari?

Sure, @robintw . I'll create scripts this weekend.

BarisSari:
Hello @robintw . I created migration scripts. Then, I tested upgrading and downgrading PostgreSQL. It worked without any problem. It also works correctly when I upgrade an SQLite database. But when I try to downgrade SQLite, Alembic throws an error. It creates constraints correctly but it doesn't detect them in the downgrade method. I'll debug it later today.

robintw:
Hang on, I think this was still waiting for @BarisSari to debug the issues with the downgrade method - is that correct Baris?

BarisSari:
> Hang on, I think this was still waiting for @BarisSari to debug the issues with the downgrade method - is that correct Baris?

Yes, that's correct. I'll open a Hotfix PR later on for the downgrade method of SQLite.

IanMayo:
Oops. Sorry @BarisSari & @robintw  - I saw the new scripts and thought it was complete.

PR #493, opened 2020-06-19
Add Nisida importer

### LINKED ISSUE #506: Introduce `Track Number` (opt) to Contacts table.
A sensor can simultaneously track multiple contacts.  Some sensor break them down into separate entities. The most common way of reporting these is via track numbers, something like "S424".

We should support this by adding an optional Track Number field to the contacts table.

It should appear before the "Subject" field.

This field will be used in the Nisida importer.

### LINKED ISSUE #509: Introduce `Remarks` field for measurement tables
In the Nisida file format, lots of entries allow for a `Remarks` field.

We _were_ implementing this through creating a row in both the `measurement` and the `comments` tables.

But, if we add a `Remarks` field to the measurements tables (other than the `Comments` table) then we keep the remark with the fields for that row.

In time we may need to produce a "Comments & Remarks" query that combines `Comments` with non-null `Remarks` from other measurement tables.

robintw:
Adds an importer for Nisida format files.

Also implements a number of changes to the database schema including:
 - Adding `remark` columns to all measurement tables
 - Changing the type of the `Comments.content` field to be unlimited length text
 - Changing the SRID of the `Geometries.geometry` field to match all the other geometry fields we have
 - Changing `start` and `end` in `Activations` to be optional (the importer code ensures one must be set)
 - Making `name` optional for both `Activations` and `Geometries`

Fixes #506 
Fixes #509


robintw:
Good comments, thanks - I'll deal with them tomorrow. I think we need to have a broader think about how best we test importers - some of the existing tests for other importers are extremely limited, and we're often not properly checking error situations in importers. I'll think about a sensible and relatively quick way of testing this now, and then open another issue to discuss broader testing of importers.

PR #504, opened 2020-06-25
Use "." to cancel or exit in Pepys-Admin

### LINKED ISSUE #442: Consistent 'exit' key
In `pepys-import`, users enter `.` to cancel import.

But, in `pepys-admin`, users enter `0` to exit.

If we switch `pepys-admin` to use `.` for exit, it will _release_ `0` to be used for something else in the future (such as potentially offering another _existing item_)

The RST & Binder documentation would have to be updated accordingly.

BarisSari:
Fixes #442


BarisSari:
> Thanks for this @BarisSari. It all looks good, and all the docs seem to have been updated properly.
> 
> However, I'm just wondering whether there is anything we can do to reduce duplication across the files. We seem to be writing the same `default` method for all of our `XYZShell` classes. Do you think it'd work to extract that method to a mixin (like we've done for the various database objects)? If so, then `postcmd` might be able to be extracted too. What do you think?

That was a good suggestion, @robintw . I created a Base class named `BaseShell` and removed duplicated `default`, `postcmd` methods.

PR #507, opened 2020-06-28
Update Latest Versions

BarisSari:
In #473, I added manually generated migration scripts but apparently, I forgot to commit `latest_revisions.json` and therefore the values are old. This PR corrects the values.


BarisSari:
This problem will be solved when #493 is merged.

PR #514, opened 2020-07-02
Hotfix - failing with extra token on first line

### LINKED ISSUE #513: Failing with trailing POS in UNIT line
The users are processing their first NISIDA file.

The first line is the `UNIT/` line we need for the date.  But, it looks like the `POS/` from the second line has been pulled onto the first, so it looks like this:
```
UNIT/ADRI/JAN19/SRF/POS/
311002Z/3623.00N/00412.02E/GPS/359/03/-/
311030Z/3653.02N/00430.06E/GPS/356/13/-/
```
But, it should be like this:
```
UNIT/ADRI/JAN19/SRF/
POS/
311002Z/3623.00N/00412.02E/GPS/359/03/-/
311030Z/3653.02N/00430.06E/GPS/356/13/-/
```
I've asked them to remove the `POS/` from the first line. We don't need them to put it back onto the second line, since we don't rely on the `POS/` header.

But, we should modify Pepys so that it robustly handles extra fields in the `UNIT/` line.

IanMayo:
Supports #513


IanMayo:
Aah, hold on. My workaround is breaking the tests.

I've just spoke to the clients, and they don't need this workaround for this weekend's trials - they only expect a couple of NISIDA files, and they understand how to manually check for (and fix) an invalid `UNIT` line.

But, @robintw - would you mind stepping into this branch and fixing the tests please?

(no rush at all)

PR #521, opened 2020-07-02
Update versions of dependencies

robintw:
This PR updates the versions in requirements.txt and requirements_dev.txt.

Where possible, if the dependency seems to have a sensible versioning approach, the version specifiers have been changed to automatically allow bugfix releases, but not to allow major releases that may have breaking changes.

A few packages, such as `pip` and `wheel` have had their version specifiers removed so that they always use the latest version - this saves some time in the deployment process as currently these get installed at the latest versions by the deployment script, and then are downgraded by the `requirements_dev.txt` install.


requirements.txt
```diff
@@ -1,12 +1,12 @@
...
+psycopg2==2.8.*
+pint==0.14
 tabulate>=0.8.7
 pyfiglet>=0.8.post1
-lxml==4.5.0
-python-dateutil==2.8.1
-prompt_toolkit
+lxml==4.5.*
+python-dateutil==2.8.*
+prompt_toolkit==3.*
```
IanMayo, line 9: @robintw - is there some logic behind whether we use `==` or `>=` in the comparisons?

robintw: Yes, mostly.

I'm using `==X.Y.*` or `==X.*` when I want to pin the dependency to that specific major or minor version (eg. `==3.*` will accept 3.0, 3.1 and 3.2.5 but not 4.0 or 5.0). It's easier than writing `> X.Y and < A.B`

I'm using `>=X.Y` if we just want the latest that is greater than a particular version, and aren't particularly concerned about going to the next major version (either because that project's versioning doesn't use major/minor versions, or because it doesn't really matter for that dependency and we can just use the latest version).

Just as an additional note: if these version specifiers 'get it wrong' and install something that we don't want, then we'd pick it up with the automated tests anyway.

PR #525, opened 2020-07-03
Stop using platform_name to identify a platform, as there may be duplicates

### LINKED ISSUE #136: 7.2.1 PostGIS deployment
* [x] PostGIS installed on network
* [x] Pepys database present in PostGIS
* [x] Pepys user account defined for PostGIS
* [x] Pepys installed on shared drive
* [x] Database config file produced for PostGIS (in shared drive)
* [x] Environment variable configured to point to config file

Update sequence
* [ ] Download/verify/transfer
* [ ] Archive current version (in shared folder)
* [ ] Unpack new version to shared folder
* [ ] Select `Pepys-Admin` from `Start Menu`. If not present click on `bin\install_pepys.bat`, and try again
* [ ] Go into menu, select `Migrate` to update database Schema as necessary

### LINKED ISSUE #362: Stop searching for Sensors/Platforms/etc by name - as names may be duplicated

### LINKED ISSUE #485: Potential new UI flow for resolving platforms
(Taken from #390, which was split into two)

Aah, this may have exposed a pattern related to how we manage duplicates.

We read NELSON from the file
We see if we have NELSON in the database
If we don't we walk user through creating new platform
If we do, we say to user We have UK / Frigate / F23 called NELSON, is this the respective platform?
If user says yes - load it for that platform
If the user says no, work through creating a new platform.

robintw:
This PR makes changes across the code to stop treating platform_name as a unique identifier for a Platform, switching instead to using the platform ID when passing data around, and to identifying by a combination of name, identifier and nationality when communicating with the user.

The three main user-visible changes, are:

1. The format of the `Sensors.csv` file used for pre-populating metadata has changed. It now has to include the name, nationality and identifier of the platform that the Sensor is hosted by, so that we can properly identify which platform this is. Therefore, the header of this file has changed to `name,sensor_type,host_name,host_nationality,host_identifier,privacy`. **Note:** This will mean that the users have to change their Sensors.csv file, or they will get an error when populating the metadata from the CSV files (the other files should import fine, but the Sensors file will not).

2. When a platform name like `DOLPHIN` is read in a file it is *not* automatically matched to a platform with the name of DOLPHIN - even if only one platform of that name exists (as if it were automatically matched to that one, then there would be no way to add an additional platform with the same name!). Instead, a menu like the example below is shown, giving options of which platform to use, and options to search or add a new platform.

```
Select a platform entry for DOLPHIN:
   1) DOLPHIN / 123 / France
   2) DOLPHIN / 123 / United Kingdom
   3) Search for existing platform
   4) Add a new platform, default name 'DOLPHIN'
   .) Cancel import     
```

When a platform is selected, the link between the name `DOLPHIN` and that particular platform is cached *for this specific file only*. It seems reasonable to do this caching - as otherwise the user would have to select a platform for every single line in the file! - and I can't see it causing problems (if there were two platforms with the same name referenced in a single file then they'd have to have different names within the file, or we wouldn't be able to tell the difference - and therefore the caching would link these two names to separate Platform objects).

3. When fuzzy-searching for a Platform (eg. choosing `3` in the menu above), we now display the full NAME / IDENTIFIER / NATIONALITY string in the autocomplete, so that the users can see exactly which platform they are selecting.

There are many more changes that only developers should see. These include:

 - Removal of some duplicated methods - eg the class method Platform.get_platform(), which was barely used, replacing it with the data_store.get_platform method which was used far more.
 - Changes to the `find_platform` and `get_platform` methods to allow specification of name, identifier and nationality - or alternative a `platform_id` (for use in code paths where the Platform object has already been resolved - and passing the ID saves resolving it entirely again)
 - Adding a `get_cached_platform` method the Importer base class, and changing all importers to use it. This is what allows us to cache the link between the platform name in the file and the Platform object. It works in a very similar way to the already existing caches for the sensors on each platform. As discussed above, I don't think this will cause any problems with caching the wrong things.
 - Indexing the `self.measurements` dict by platform ID rather than platform name, as the name may not be unique - and that could mean that different platform's measurements are combined together for validation, which could give incorrect validation results
 - Updating tests to work with the new functions and new parameters, plus adding new tests to check that an import using duplicate names works correctly, and to test the new option to choose an existing platform when resolving works correctly.

(Also good to note that the benchmarking tests warned me of an expected problem: when we switched to resolving a platform for each row, before introducing the platform caching, I got test failures showing that the benchmark took far too long, as it was checking the database for a platform for each row).

Fixes #362 
Fixes #485.


IanMayo:
Good piece of work @robintw , thanks.  Once it's merged I'll email a copy of `sensors.csv` to the analysts, for them to put it into the "safe seed data" folder.

pepys_import/core/store/common_db.py
```diff
@@ -558,6 +524,10 @@ def sensor(self):
     def sensor_name(self):
         return association_proxy("sensor", "name")
 
+    @declared_attr
+    def platform_id(self):
+        return association_proxy("sensor", "host")
+
```
IanMayo, line 530: @robintw - can you explain the role/functionality of `association_proxy` to me please?

tests/test_importers.py
```diff
@@ -581,5 +582,112 @@ def test_parse_timestamp(self):
...
+            plat1 = (
+                self.store.session.query(self.store.db_classes.Platform)
+                .filter(self.store.db_classes.Platform.platform_id == plat1_id)
+                .all()
+            )
+
+            assert len(plat1) == 1
+            assert plat1[0].name == "DOLPHIN"
+            assert plat1[0].nationality_name == "United Kingdom"
+
```
IanMayo, line 680: @robintw - I'm sure there's an easy explanation for this 🤣 

Since nationality is a foreign key, shouldn't it be storing the ID for `United Kingdom`, rather than the string?

Aah, maybe `nationality_name` is a function that retrieves the name for the foreign key.

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -97,26 +97,50 @@ def resolve_datafile(self, data_store, datafile_name, datafile_type, privacy, ch
...
+
+            # The order-by clause is important, to get the same ordering of
+            # options on different platforms/db backends, so that our tests work
+            platforms = (
+                data_store.session.query(data_store.db_classes.Platform)
+                .join(data_store.db_classes.Nationality)
+                .filter(data_store.db_classes.Platform.name == platform_name)
+                .order_by(
+                    data_store.db_classes.Platform.identifier.asc(),
+                    data_store.db_classes.Nationality.name.asc(),
```
BarisSari, line 116: We are ordering Nationalities by their priorities when we resolve Nationality. Do you think that it could be better for users if we use the same order here?

pepys_import/core/store/common_db.py
robintw: Sure, it provides a way of getting a 'fake' attribute on a SQLAlchemy model that actually comes from a relationship. So, for this example we have a SQLAlchemy model that represents a row in the `States` table. This has a SQLAlchemy relationship defined which links the `State` entry to the `Sensor` entry that it refers to (via the `sensor_id` field). This association proxy basically just gives a nice name to an attribute that is on the other end of a relationship. In this case, the relationship is the Sensor relationship, and here we're giving a nice name to the `host` property of the sensor found through that relationship.

Basically, it's a nicer way of saying `State.sensor.host` - instead, you can just do `State.platform_id`.

tests/test_importers.py
robintw: `nationality_name` is another one of these association proxies, which means we can easily access the name of the nationality, even though it is stored as a nationality ID (and in this situation we can't just check the ID as it'll be different each run, so we want to check the name)

pepys_import/resolvers/command_line_resolver.py
robintw: Good idea - I'll do that now!

robintw: Fixed in 86b05a4

PR #531, opened 2020-07-09
Create file associations for .sqlite and .db files with Pepys Admin

### LINKED ISSUE #526: Add file association to load .sqlite files in Pepys Admin
Needs to be added to deployment scripts.

Will have to work out whether this is for .sqlite or .db or both, or whether there is a more generic way we can do this.

robintw:
This PR modifies the install script so that it adds file associations for .sqlite and .db files to be loaded with Pepys Admin. It also assigns them the Pepys icon.

The recommended way of changing file associations in Windows is using the `ftype` and `assoc` commands, but these only work if you have administrator privileges. 'Under the hood' these commands edit the registry, and the relevant part of the registry is available under HKEY_CURRENT_USER and HKEY_CURRENT_MACHINE, with the former being writeable by the current user, and the latter writeable only by administrator. We create the registry keys under HKEY_CURRENT_USER, which means that a normal user should be able to created them (it works on my machine).

**Note**: The extra security policies they have at the client may mean that users are not able to modify file assocations themselves, even in the HKEY_CURRENT_USER part of the registry (sometimes malicious programs try to do this to change file associations to malware) - so this may or may not work for the client, we'll have to test it. If it doesn't work then they can probably do this via Group Policy, but we'd have to talk to the administrator.

@IanMayo: It'd be good if you could test this on your Windows machine, and check it works without admin permissions. Just running the install script should be all you need to do, then try double-clicking on a .sqlite or .db file.

Fixes #526

### DIFF
--- a/bin/install_pepys.ps1
+++ b/bin/install_pepys.ps1
@@ -83,4 +83,19 @@ if (!($env:Path -split ';' -contains $pepys_bin_path)) {
         [EnvironmentVariableTarget]::User)
 }
 
+# Create file associations for .sqlite and .db files to open in Pepys Admin
+$pepys_admin_run_command = [System.IO.Path]::GetFullPath(".\pepys_admin.bat") + " --db %1"
 
+# All of the assigning to $null below is just to stop the default output showing exactly
+# what registry keys have been created
+
+# Create the file extension entry for .sqlite and assign it to the filetype sqlitefile
+$null = New-Item -Path HKCU:\Software\Classes -Name .sqlite -Value sqlitefile -Force
+# Do the same for the .db extension
+$null = New-Item -Path HKCU:\Software\Classes -Name .db -Value sqlitefile -Force
+# Specify the 'shell open' command for the filetype sqlitefile (that both extensions reference)
+$null = New-Item -Path HKCU:\Software\Classes\sqlitefile\shell\open -Force -Name command -Value "$pepys_admin_run_command"
+# Set the default icon for the filetype to the Pepys icon
+$null = New-Item -Path HKCU:\Software\Classes\sqlitefile -Force -Name DefaultIcon -Value $icon_string
+# Refresh the Windows Explorer icon cache, so the icons show immediately
+ie4uinit.exe -show


IanMayo:
@robintw - on my MS-Windows VM, `.sqlite` files are already registered with `DB Browser for SQLite`.

So, on right-click I only get these options:
image

If I go into `Choose another App` it offers me Pepys:
image

It just offers the file path.

I accept that a user may have .sqlite registered with another app. I'm surprised that Pepys wasn't offered alongside DB Browser.  I'll restart the VM and see if it changes.

robintw:
Hmm, that's interesting. If anything, I'd have thought it would have over-written the DB Browser option, and made Pepys the default.

I'll have a go installing DB Browser for SQLite myself, and seeing how it handles it. I think the 'Open With' menu is actually configured by different registry entries than the standard file association.

robintw:
I've spent a while looking into this just now - and it's all very complicated. There are multiple ways to set file associations in Windows: some of them require admin permissions and some of them don't, and some work with the Open With list and some don't.

The easiest way forward would be to test this with the client and see if it works in their configuration. If they don't have DB Browser for SQLite installed, and have never used Open With on a SQLite file, then it *should* work as it is. It's possible that none of this will work for the client, as their security policies may stop us making changes to the registry anyway - so we should probably try it to find out before we spend too long looking at other ways of modifying the registry to make it work better.

I think part of the problem is that we're using a .bat file rather than a .exe file - but there's not much we can do about that. Another part of the problem is that we're not doing this with admin permissions: there are two main places this information is stored in the registry, and I think they can end up conflicting if, for example, DB Browser was configured with admin permissions, and Pepys is configured without admin permissions. If we had admin permissions then we could *probably* just use the `ftype` and `assoc` commands that Windows provides, and things would probably work more easily.

So, I think the best thing to do is trial this and see - what do you think?

As a final note: if we can't get this working reliably, then we could always add a Pepys Admin shortcut to the Send To menu. We know how to do that, and know it works for them.

IanMayo:
That's all good, thanks Robin.  I'll talk it through in the phone chat tomorrow, and probably try and do a trial next week.

PR #533, opened 2020-07-10
Added CheckConstraints to all non-nullable String columns

### LINKED ISSUE #497: Add check constraints to ensure field values are not empty strings
We've already added SQL CHECK constraints to ensure that string fields that we're using as part of the uniqueness criteria for tables (eg. `name` or `identifier`) aren't set to the empty string (see #473). Should we also do this for many (or all) other string fields that are set to `nullable=False`?

For example, empty strings for the `name` fields of the reference tables make no sense, and the same applies for fields in `Synonym` and others.

robintw:
The added CheckConstraints check that the column is not an empty string.

Includes migrations for SQLite and Postgres.

@BarisSari: The Postgres migration seems to work correctly, but the SQLite migration gives an error when we try to downgrade. I suspect this is a problem related to #427 - does it need that merging first to create the right migration? Or do I just need to copy in the most recent definitions of the tables?

Fixes #497.


BarisSari:
> @BarisSari: The Postgres migration seems to work correctly, but the SQLite migration gives an error when we try to downgrade. I suspect this is a problem related to #427 - does it need that merging first to create the right migration? Or do I just need to copy in the most recent definitions of the tables?

I think you can copy the most recent definitions of the tables and add `copy_from` parameters to `with` clauses, @robintw . I can also do that for you if downgrade still doesn't work.

I'm still waiting for an answer on SO for the #427. So, it would be better not to depend on it for now.

robintw:
Thanks @BarisSari.

I've added the `copy_from` parameter, and copied the table definition into the class, but I'm still getting an error: `ValueError: No such constraint: 'ck_Tasks_name'` - even though that constraint is defined in the table definition that I've copied.

Any ideas what's going on? Ideally, if you could let me know what you think is wrong, then I can do the fix - as I could do with getting more familiar with these sorts of Alembic issues.

BarisSari:
To be frank, I don't why it doesn't find the constraint, @robintw . However, I had faced this issue before and I found a workaround in my Alembic PR (https://pepys-import--427.org.readthedocs.build/en/427/database_migration.html#possible-errors-and-solutions). You can find it in `For Developers Only` section:

> 
> `ValueError: No such constraint: 'XXXX'`
> 
> This error might happen in both PostgreSQL and SQLite databases. However, when it happens in SQLite, it might mean that your class that passed in copy_from parameter to batch_alter_table doesn’t have the necessary constraint. You might compare the related class in pepys_import.core.store.sqlite_db with the class in the migration script to find out the missing constraint. Then, you might add the constraint to the class using Table.append_constraint() method:
> 
> ...
> # Append constraint
> Sensor.__table__.append_constraint(CheckConstraint("name <> ''", name="ck_Sensors_name"))
> with op.batch_alter_table("Sensors", schema=None, copy_from=Sensor.__table__) as batch_op:
>     batch_op.drop_constraint("ck_Sensors_name", type_="check")
> ...
> 
> If you face this error during the upgrade, please add append_constraint() line inside of the upgrade() function. If you face it during the downgrade, add the line inside of the downgrade() function. This error should be corrected now.
> 

_Note:_ I'm bothered because of SQLite related issues. I'm checking the documentation of Alembic from time to time and I don't see most of the issues we have. It might be that I'm missing some important points.

BarisSari:
@robintw, I played with the table definition a bit and I think the problem is Alembic doesn't detect the constraints in the `Column` definitions. When I moved the constraint to the `__table_args__` tuple, it worked:

```python
    class Task(BaseSpatiaLite, TaskMixin):
        __tablename__ = constants.TASK
        table_type = TableTypes.METADATA
        table_type_id = 4

        task_id = Column(UUIDType, primary_key=True, default=uuid4)
        # It was: name = Column(String(150), CheckConstraint("name <> ''", name="ck_Tasks_name"), nullable=False)
        name = Column(String(150), nullable=False)
        parent_id = Column(
            UUIDType, ForeignKey("Tasks.task_id", onupdate="cascade"), nullable=False
        )
        start = Column(TIMESTAMP, nullable=False)
        end = Column(TIMESTAMP, nullable=False)
        environment = deferred(Column(String(150)))
        location = deferred(Column(String(150)))
        privacy_id = Column(
            UUIDType, ForeignKey("Privacies.privacy_id", onupdate="cascade"), nullable=False
        )
        created_date = Column(DateTime, default=datetime.utcnow)
        # Constraint added here
        __table_args__ = (CheckConstraint("name <> ''", name="ck_Tasks_name"), )
```

robintw:
Thanks @BarisSari. I've implemented that, and it all seems to be working.

Could you investigate whether this is a known problem with alembic, and if it doesn't seem to be known then could you submit a bug report to alembic itself?

BarisSari:
> Could you investigate whether this is a known problem with alembic, and if it doesn't seem to be known then could you submit a bug report to alembic itself?

Sure, I'll do that and let you know.

Issue #482, opened 2020-06-16
Add testing of imports using command-line resolver


robintw:
All our testing of the importers currently use the default resolver, because that's far easier for tests. However, this means that we're not testing the main way that users will import files - which is through the command-line resolver. The command-line resolver has good tests itself, but it isn't tested in an 'end-to-end' way.

I think it'd be good to have some tests that do an import of a file using the command-line importer. Ideally we'd do this once for each importer type.

IanMayo:
Yes good idea.

It will be fragile to importer modifications (including wording).

But, if the wording comes from a constant (or Python f"" string) - then we should also be able use that constant in the unit tests.

PR #535, opened 2020-07-11
Add end-to-end automaton testing of the import CLI, plus fix the bugs it found

robintw:
This PR adds an interesting new sort of test for the end-to-end testing of the import CLI. It uses an 'automaton' which is implemented by monkey-patching the `prompt` calls in the command-line resolver code. This automaton gives reasonable answers to each question asked by the command-line resolver: either choosing from the list, where a list or auto-completion list is given, or giving one of a predefined set of answers.

This means, whenever the test is run, the interface is 'exercised' with a load of potential interactions, which can flush out various bugs. The tests use randomness in choosing which answers to give, so are run five times each to get a reasonable coverage, and I have implemented tests for importing a REP file and a Nisida file. If the test fails then the output that will be shown includes the random seed used to run the test, so it can be recreated for testing purposes. A lot of the interactions 'go round in circles' a bit, as it gives invalid input, so the question is asked again, and again, until it randomly picks a valid input. Still, that is a good test of our invalid input handling code! Basically, if the CLI can deal with this automaton hammering it with crazy stuff, then it should be able to deal with a real human operating it sensibly!

This is a relatively naive automaton at the moment, and doesn't do absolutely everything that is possible in the GUI, but it has already brought a number of bugs to my attention. These bugs are fixed in this PR, and are mostly around checking input from the user more carefully to deal with empty input, input that is too long, input that isn't properly parsed etc.

This works towards a solution to #482, but I'd like to add a few other more manual tests before we close that issue.


robintw:
Thanks @BarisSari.

I've updated each run to repeat 10 times, giving 20 repeats overall. I think (hope) that will be sufficient. This adds around 2 minutes to the run time of the test suite. I've also marked these tests with an `automaton` marker, so that they can be easily de-selected for running the test suite quicker (`-m 'not automaton'`).

BarisSari:
I think you might switch Postgres to SQLite to fasten tests, @robintw. By the way, I still see that you haven't fixed the random seed. Do you think that we should keep it this way?

robintw:
I originally used SQLite, but that didn't actually turn up lots of bugs because SQLite doesn't enforce various things (for example, it doesn't enforce lengths of string fields - so we could add a very long string to a field that was set to a max length of 4 with no errors). A number of the bugs that these tests found only occurred when these field lengths were enforced.

We could fix the random seed, but I wonder whether keeping it random would give better coverage overall - albeit at the risk of giving a random failure at some point (which is frustrating, but probably better than not getting that failure at all). What do you think?

BarisSari:
> I originally used SQLite, but that didn't actually turn up lots of bugs because SQLite doesn't enforce various things (for example, it doesn't enforce lengths of string fields - so we could add a very long string to a field that was set to a max length of 4 with no errors). A number of the bugs that these tests found only occurred when these field lengths were enforced.

Aah, I see. Yes, it makes sense to use Postgres then even though it is slower.

> We could fix the random seed, but I wonder whether keeping it random would give better coverage overall - albeit at the risk of giving a random failure at some point (which is frustrating, but probably better than not getting that failure at all). What do you think?

Yeah, I think it will give better coverage overall. Actually, that's completely fine for me to have random failures (I already have them thanks to the Alembic 🙂 ). I just thought that it would be weird for a new developer. Maybe you might add some notes to the test class to explain this situation.

robintw:
Good idea, I've added a long docstring (partly copied from the first comment in this PR), and renamed the tests to make them clearer.

IanMayo:
Oops, conflict based on other merges, @robintw

PR #538, opened 2020-07-13
Add end-to-end tests for importing files with the CLI

### LINKED ISSUE #482: Add testing of imports using command-line resolver

robintw:
This is a companion PR to #535. This PR tests the import of a couple of files (a REP file and a GPX file) by using a static list of responses to the CLI, testing the full import process with known responses. This is different to #535 which generates random responses that fulfil the requirements of the CLI.

This and #535 combined fix #482 
This didn't show up any bugs at the moment, but it will fail if we change the CLI interface at all. This is actually good, as we don't want to be accidentally changing the CLI interface (I think we've done this accidentally before, when bugs have meant a lot more questions have been asked).


IanMayo:
@robintw - merge conflict to be resolved, please.  (No rush - I know you're off tomorrow)

PR #539, opened 2020-07-15
Fix issues with EAG importer

### LINKED ISSUE #510: EAG files - file contains two tracks
Clients have received an EAG file.  Sender has reported that file contains tracks for two aircraft.

Client is sending redacted copy of file for us to examine.

### LINKED ISSUE #511: EAG file not recognised by Pepys
Clients are running 0.0.17, which should include EAG support.

When they right-click on EAG file, and `Send to Pepys`, it reports `Files processed 0 times`.

robintw:
This fixes two issues with the EAG importer:

1. Fixes #511, where files weren't recognised by the EAG importer because of how the filename checks were performed. This PR makes the filename checks more robust, as described in that issue.

2. Fixes #510, where files can have multiple tracks. This PR adds support (and tests) for multiple tracked EAG files.


IanMayo:
Great piece of variable naming here @robintw :
image

IanMayo:
I've got one request, please. We have a sample file where there are three header rows, but only two tracks (24th June). In this instance, the callsign is missing from the third header row. Could you change your sample so:
* there are only two tracks in the data
* there are three header rows
* the third header row misses the callsign

robintw:
What do you want to happen in that case @IanMayo? Do you want the file to be rejected as invalid, or the missing callsign to be ignored? I'd be tempted to reject the file as invalid - as when we're parsing the header we don't actually know which of the track IDs (and therefore callsigns) will be used later on in the file (though we could just store it as `None` and catch it later on if it was to be used)

IanMayo:
Hello @robintw ,
the file is only invalid if a row of data has an index that isn't present in the headers.  In the scenario I described above, the track data is only for two callsigns, so it isn't a problem that there isn't a callsign in row 3.

robintw:
Thanks @IanMayo - fixed in 1cb25c7

PR #540, opened 2020-07-15
Add more tests for the migration option

BarisSari:
This PR adds 3 tests for the migration option in Pepys-Admin. For both PostgreSQL and SQLite, it tries to apply the migration to an empty database and a database with the sample data imported. Actually, the second database is already initialised before the import of sample data, so basically, the database is not migrated because it is already at the latest version. However, I think it is good to check that database stamped correctly, and running the migration option doesn't make any changes.


IanMayo:
Yes, @robintw  - I had a similar thought at the end of last week.  

If we try to ensure we have sample data that represents all data-types we import [1], then we can populate a database with data that covers all tables/fields in use.  Then we could run all of the migrations on that dataset - and (hopefully) ensure a migration isn't introducing a constraint that can't be met because of existing data.

Hmm, since I believe SQLite constraints can be subtly different to PostGis ones - should we do the above for PG as well as SQLite?

[1]. This may represent a change in strategy, since currently we may only test for importer support for all data-types in the unit tests, rather than data-files.  I'm unsure if it's a large task to go back & ensure the sample data covers all supported data-types.  But, it's probably worth it - since it demonstrates the format(s) of input data that we expect.

robintw:
It'd be good to do that for Postgres too. I think we'd have to create an out-of-date Postgres database and populate it manually, and then export that to a .sql file (or another postgres-specific file) and commit that to the repo - and then inside the test create a Postgres db and run the .sql file in it.

As for covering data-types in the files we import: what exactly do you mean by data-types here? We definitely have datafiles that cover all of the different tables we import, and I think we cover most (and probably all, but I'd need to check) the fields that we can import. Interestingly, most of our unit tests for the importers are actually based on real files that we import, rather than just strings of data that we test with - this has its advantages and disadvantages, but here I think it is an advantage.

IanMayo:
Good.

For the data-types I was mostly thinking about the Nisida and Replay formats - since those importers support a range of types of element (especially Nisida).  But, if we mostly drive the unit tests from test data then we do probably have good support.

Aah, one issue - when we're populating the initial database, we should probably skip the very large test file (NMEA_TRIAL.log) - since it bring a performance impact.

BarisSari:
Thanks for the review. I'll populate databases with old code and create tests. I'll add them to this PR.

BarisSari:
Hello @robintw and @IanMayo . I used Pepys 0.0.17 (released on Jun 16) to create old databases and I added them to this PR. These old databases have all sample data but `NMEA_trial.log` and `nisida`. We can't import `nisida` data because we didn't have an importer for that at that time. All tests are passing when we upgrade old version databases to the latest versions.

IanMayo:
Nice work @BarisSari .

Since Nisida introduces some data-types that aren't in other importers, it would be useful to be able to test the migration of those tables when they contain data.

Is it possible to import the nisida file once the migrations have reached the correct point?

There is a chance that we may wish to do this in the future:  as we add a new file/data-type, we may have to modify a schema, and it would be good for the testing to handle the migration, and then load the data-file.

This _could_ look like a table/array in that lists alembic versions plus the data-files that should be imported once those migrations have been completed. 

(if this is silly, just say ...)

BarisSari:
> Nice work @BarisSari .
> 
> Since Nisida introduces some data-types that aren't in other importers, it would be useful to be able to test the migration of those tables when they contain data.
> 
> Is it possible to import the nisida file once the migrations have reached the correct point?
> 
> There is a chance that we may wish to do this in the future: as we add a new file/data-type, we may have to modify a schema, and it would be good for the testing to handle the migration, and then load the data-file.
> 
> This _could_ look like a table/array in that lists alembic versions plus the data-files that should be imported once those migrations have been completed.
> 
> (if this is silly, just say ...)

It's not possible if we use the migration option in Pepys-Admin. Because it directly upgrades the database to the latest version with one command. I don't think that we can interrupt it. I think we can achieve this goal if I modify the current Nisida Importer and put it in Pepys 0.0.17. If that works, we don't need any workarounds. Do you want me to do that?

For the future, I suggest upgrading the database to a specific version using Alembic command (calling `alembic upgrade VERSION-NUMBER` instead of `alembic upgrade head`), importing some data, and running the migration option in Pepys-Admin.

robintw:
I'm not sure it's a great idea to modify past versions of Pepys once they've already been released - so I don't think adding the Nisida importer to v0.0.17 would be a good idea.

Can your suggestion for the future be implemented as part of these tests? Eg the table that Ian suggested, giving version to be imported, followed by any files to import after that, followed by the next version to be imported? And using alembic commands themselves (well, the Alembic Python API) to upgrade to a specific version, rather than using the Pepys Admin migration command.

I'm happy to postpone this more advanced testing to a later PR if you'd prefer: but I'm also aware that we're winding down work on this phase of the project shortly.

IanMayo:
Good advice @robintw - I don't think we should modify existing versions.  If/when we introduce new file formats we would again have to modify released versions.

I'd appreciate us having a go at the table of versions/data-files, please.  It's a fairly complex piece of functionality - so I'd like us to start learning lessons on it.

BarisSari:
Okay, I'll create a new table of versions/data-files 👍 But if you need old database @robintw , I think we can merge this one and I might work on another branch. Because I don't know how long it will take to add this functionality.

robintw:
Ok great, I'll review this shortly and then it can be merged, and you can start the new work in another branch. Thanks! :)

PR #541, opened 2020-07-16
Configure coverage.py to ignore certain parts of the code that are untestable or don't need to be tested

robintw:
This PR is just a bit of housekeeping to configure the coverage checker for our tests (the coverage.py library) to ignore certain bits of the code. This will mean that our coverage reports are more accurate.

Basically this configuration excludes anything marked with a comment saying `# pragma no cover`, plus all `__repr__` and `__str__` methods (just used for debugging), plus places that raise `NotImplementedError`.

I've also added some of the `# pragma no cover` comments to bits of the code that can't be tested - for example, assert statements that test 'impossible' situations that would be impossible (or extremely difficult) to set up for testing. For example, there are assert statements in the code that check that the merge isn't in a situation where the database constraints have been broken - these are now skipped when coverage is determined. Many of these comments are on a line that starts a block (eg. `else:`) - in this case they skip the whole block.


IanMayo:
@Robin - do you need to take this out of draft mode?

PR #542, opened 2020-07-16
Add more tests

robintw:
This PR adds more tests in various places in the code which weren't covered well by tests.


robintw:
I've got some more tests to add sometime - particularly to some of the importers - but I think this should be reviewed and merged now, and I'll continue with the other tests later.

tests/test_admin_cli.py
```diff
@@ -1123,6 +1144,32 @@ def test_do_export_reference_data(self, patched_input):
...
+    def test_do_export_reference_data_invalid_filename(self, patched_input):
+        with open("already_existing_file.db", "w") as f:
+            f.write("Hello, world")
+
+        patched_input.side_effect = ["already_existing_file.db", "test.db"]
+        temp_output = StringIO()
+        with redirect_stdout(temp_output):
+            self.shell.do_export_reference_data()
+        output = temp_output.getvalue()
+        assert "Reference tables are successfully exported!" in output
```
BarisSari, line 1157: I think it would be better if we also assert the error message of `already_existing_file.db`

robintw:
Both fixed now, thanks @BarisSari.

PR #545, opened 2020-07-21
Add check that database is at latest revision before merging

robintw:
This PR adds a check to the merging process to make sure that the slave database that is being merged into the main database is at the latest revision, and refuses to merge if it isn't. This is the latest revision according to the `latest_revisions.json` file that is present in the master installation - so this means that the slave SQLite database has to be at the latest version according to that installation - and therefore the schema should match the master Postgres database (as that will have been migrated when the installation was last upgraded).

@BarisSari: I'm going to write some tests for this tomorrow, but it'd be useful to use the 'out of date' SQLite database that you've created for tests in #540 as part of these. Therefore, I'll leave writing the last few of these tests until that PR is finished and merged, then we won't duplicate work.

### DIFF
--- a/pepys_admin/snapshot_cli.py
+++ b/pepys_admin/snapshot_cli.py
@@ -1,19 +1,16 @@
 import cmd
-import json
 import os
 import shutil
-import sqlite3
 import tempfile
 
 from iterfzf import iterfzf
 from prompt_toolkit import prompt as ptk_prompt
 from prompt_toolkit.completion import PathCompleter
 
-from paths import MIGRATIONS_DIRECTORY
 from pepys_admin.base_cli import BaseShell
 from pepys_admin.merge import MergeDatabases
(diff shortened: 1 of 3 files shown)


BarisSari:
> @BarisSari: I'm going to write some tests for this tomorrow, but it'd be useful to use the 'out of date' SQLite database that you've created for tests in #540 as part of these. Therefore, I'll leave writing the last few of these tests until that PR is finished and merged, then we won't duplicate work.

I've added an old SQLite database to that PR, @robintw . It's waiting for your review.

robintw:
Thanks @Baris. I've added tests using that database here, and I think this PR is ready for review now.

IanMayo:
Thanks @robintw - it looks just right. You need to mark it as `Ready for review`.

robintw:
Fixed now @IanMayo, sorry about that.

PR #546, opened 2020-07-23
Add Advanced Migration Tests 

BarisSari:
This PR adds advanced migration tests that are discussed in #540 . 

**Note:** I checked the date of the merge of Nisida Importer PR (on July 1st) and decided the alembic version number according to that.


tests/migration_tests/test_migration.py
```diff
@@ -118,3 +125,175 @@ def test_do_migrate_from_old_version_postgres(self, patched_input):
...
+        while True:
+            try:
+                command.upgrade(config, "+1")
+                new_version = get_alembic_version(connection)
+                if new_version in version_datafile_dict:
+                    import_files(version_datafile_dict[new_version], data_store)
+                if new_version == latest_sqlite_version:
+                    print("Upgrade to head is successful!")
+                    break
+            except Exception as e:
```
robintw, line 226: Do we want to fail the test as well if the alembic migration gives an error? Or are we expecting some (minor) Alembic errors that we just have to ignore?

(Same applies in the Postgres test)

BarisSari: Actually no, we don't want it anymore. Previously, I wasn't reading the `latest_revision.json`, so it was trying to upgrade from the latest to a non-existing version and it was giving an error. But I added another break for the while loop and I think we are okay to remove it. I'll do that for both tests.

BarisSari:
> Thanks @BarisSari. I'm happy for this to be merged now.
> 
> One minor comment is that it might be easier to store the version-datafile list in a Python data structure (a list of lists, or a dict) rather than creating a new table in the database just for this. However this is really implementation preference, so I'm happy for it to stay as it is.

Yes, you're right. It would be much easier 🙂  We talked about a table and I didn't think of any other options. Let's wait for @IanMayo's comment. If he also agrees, I might change it.

IanMayo:
Good morning team :-)

Yes, storing the details as text-files does seem tidier. 

But, does the implementation only test the Nisida files?   There are lots of other files we should load, in order to check the migrations.

robintw:
I think actually storing the values as a Python dict, rather than a text file, would probably be the tidiest way.

I'll let @BarisSari answer the other part, but I think all of the other files were imported in the 'original old database' file that we're then migrating from.

BarisSari:
Good morning, @IanMayo .

> Yes, storing the details as text-files does seem tidier.
Like @robintw explained above, I think it would be easier to store the values in a Python dict.

> But, does the implementation only test the Nisida files? There are lots of other files we should load, in order to check the migrations.
Old databases were populated by Pepys 0.0.17. Therefore, all sample data were already imported except Nisida files. When we discussed earlier, we said that this would be the "oldest" database for the clients. That's why I chose this version.

IanMayo:
Thanks @BarisSari - In Slack we've established that any new migrations are checked against legacy data formats.

Yes, I support switching to storing the version/data details in Python dictionary files.

BarisSari:
> Thanks @BarisSari - In Slack we've established that any new migrations are checked against legacy data formats.
> 
> Yes, I support switching to storing the version/data details in Python dictionary files.

Okay, I'll make changes today.

PR #548, opened 2020-07-27
Add more importer tests

robintw:
This PR adds more tests for various importers.

I've taken a function I wrote for testing the Nisida importer and made it more generic and used it for all of these. This function, called `check_errors_for_file_contents` takes a string to process through the importer (it puts it into a temporary file, and processes that file with the importer), plus an error string to check is included in the errors output.

For example:

```
check_errors_for_file_contents(
            ";SENSOR: 100112 120600 SENSOR @A 16 15 00.00 N 030 4587 00.00 E  252.41 107.26 TA SUBJECT held on TA",
            "Error in longitude minutes value 4587.0. Must be between 0 and 60",
            contact_importer,
            filename="test.rep",
        )
```

This puts the `;SENSOR...` string into a temporary file, runs the importer, checks that `Error in longitude minutes value 4587.0. Must be between 0 and 60` is in the output errors listing, and creates the temporary file with the name `test.rep` (we need to specify a filename that will pass the `can_load_this_filename` checks).


robintw:
After a few more changes, we've now got 100% coverage for the importers, which I think is a Good Thing.

PR #549, opened 2020-07-28
Add a few more tests

robintw:
Adds a few more tests to cover some bits of the code that weren't covered previously.


tests/test_view_data_cli.py
```diff
@@ -111,5 +114,127 @@ def test_postcmd(self):
...
+
+    def test_do_cancel(self):
+        temp_output = StringIO()
+        with redirect_stdout(temp_output):
+            self.shell.do_cancel()
+        output = temp_output.getvalue()
+        assert "Returning to the previous menu..." in output
+
+    def test_default(self):
+        # Only cancel command (0) returns True, others return None
```
BarisSari, line 213: I think there is a typo, it should be `command (.)`, right?

robintw: Thanks, fixed now.

tests/test_view_data_cli.py
```diff
@@ -189,7 +189,7 @@ def test_do_run_sql(self, patched_prompt):
             in output
         )
         assert (
-            "| NMEA_bad.log                     | None |  243 | 8ddb840fee218872d2bb394cc654bdae"
+            "| NMEA_bad.log                     | None |  243 | b201a229fb2c4a80bd657066e4bf9c8a"
             in output
```
IanMayo, line 193: @robintw  - has the contents of `NMEA_bad.log` changed?  Do you understand why the hash has changed?

robintw: It changed in one of the other PRs, so when I pulled from `develop` I got the new version of the `NMEA_bad.log` file, and had to update the hash.

That file changed because in one of the other PRs adding tests I added a blank line to the file to test that the parser ignored blank lines.

IanMayo: As expected, thanks.

PR #552, opened 2020-07-28
Fix bumpversion config

### LINKED ISSUE #551: Bumpversion fails in two updates
`setup.cfg` includes includes instructions for `bumpversion` on how to update two files.

But, the conversion doesn't work correctly, it results in invalid version numbering.

This can be tested by running `bumpversion patch`.  But, note that it should only be run in a branch, since it pushes the changed code during processing.

robintw:
Bumpversion was using manually-configured search/replace strings for finding the version number to replace, and these led to incorrect results. Actually, the bumpversion defaults work fine for us, so I've removed those config lines, and it uses the defaults.

Fixes #551.
I've also run black on setup.py, as it seemed to have single quotes around the version string rather than black's preferred double quotes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,12 +4,8 @@ commit = True
 tag = True
 
 [bumpversion:file:setup.py]
-search = version='{current_version}'
-replace = version='{new_version}'
 
 [bumpversion:file:pepys_import/__init__.py]
-search = __version__ = '{current_version}'
-replace = __version__ = '{new_version}'
 
 [bdist_wheel]
 universal = 1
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -4,12 +4,8 @@ commit = True
 tag = True
 
 [bumpversion:file:setup.py]
-search = version='{current_version}'
-replace = version='{new_version}'
 
 [bumpversion:file:pepys_import/__init__.py]
```
IanMayo, line 8: @robintw  - just checking.  We still need the text at line 6, just to indicate to BumpVersion that it should include `setup.py`, right?  I'm asking because we don't provide such an identifier for the third file that gets "bumped", `setup.cfg`.

robintw: Yeah we do. `setup.cfg` is acting as the bumpversion config file here, and bumpversion automatically updates its own config file.

PR #561, opened 2020-08-14
[WIP] Fix Enhanced Validator

### LINKED ISSUE #556: Fix Validator when speeds are measured or calculated to be zero
Quote from Slack below:

Next error:
```
"REP Importer - Enhanced Validator Error on Timestamp:2020-07-10 00:03:06.015000, Sensor:1084, Platform:  Northumberland":
"Calculated speed (0.047 meter / second)is more than the measured speed  * 10 (0.000 meter /second)!"
```

This looks like a special case.  As the speeds get very slow, rounding errors (well, maybe quantisation) may result in a zero speed. Hmm, I guess that if the measured speed is zero we mark any speed up to 1 m/sec as valid.

BarisSari:
Fixes #556

### DIFF
--- a/pepys_import/core/validators/enhanced_validator.py
+++ b/pepys_import/core/validators/enhanced_validator.py
@@ -40,6 +40,13 @@ class EnhancedValidator:
         except AttributeError:
             location = None
 
+        if (
+            speed == 0.0 * (unit_registry.metre / unit_registry.second)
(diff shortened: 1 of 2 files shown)


BarisSari:
Would you like me to change anything in this PR, @IanMayo?

IanMayo:
(sorry, attending mini BBQ for kids exam results)

PR #571, opened 2020-09-10
Increase limit of number of rows to show

### LINKED ISSUE #570: Admin - increase limit of number of rows to show
We currently limit the number of table rows that we show to 50:
image

Since the terminal allows the analyst to scroll back through the displayed records we could actually show a lot more.

We should increase the limit to **500**.

We still shouldn't limit the number of rows that get sent to `.csv`, or the number returned in an SQL call.

BTW - we should remove the `10` number from the preceding comment. 

IMHO - we should also move the new `500` value to a named constant near the top of the file. Something like `MAX_ROWS_DISPLAYED`.

BarisSari:
Fixes #570


BarisSari:
It seems like Black had a huge upgrade and it causes a conflict with our current codes. I'll open another PR for that. Until it's merged, I'll keep this one as a draft.

robintw (approved):
Looks great, thanks.

IanMayo (approved):
Nice work.

PR #573, opened 2020-09-10
Apply latest Black to the repository

BarisSari:
We don't specify the Black version in our `requirements_dev.txt`, and it has a new version that was released at the end of August (https://black.readthedocs.io/en/stable/change_log.html#b1). Therefore, Travis CI Build failed because there some changes (https://github.com/psf/black/pull/1612). This PR makes this repository compatible with the latest Black version. It also solves flake8 warnings that occurred during the change.

I also updated pre-commit hook. Otherwise, it could try to revert all these changes. The following commands should upgrade the libraries of pre-commit hook:
```bash
pre-commit uninstall
pre-commit clean
pre-commit install
```


IanMayo (approved):
A huge piece of work @BarisSari 

Thanks.

robintw (approved):
Looks good, thanks.

PR #576, opened 2020-09-14
Fix mod_spatialite download URL, and add error handling

### LINKED ISSUE #575: Trouble with new release
I've built a new release, but on testing it, it won't start.

Config:
* Environment variable set: 
image
* Config file:
image

Steps:
* download [release](https://github.com/debrief/pepys-import/releases/tag/0.0.21)
* unpack
* click on `bin/install_pepys.bat`. Command runs, shortcut shown
* from Start Menu, select `Pepys Admin`

image

robintw:
Fixes #575.

The error in #575 was because Pepys Admin couldn't load mod_spatialite, and the reason it couldn't load it was because the deployment process hadn't downloaded it properly. It turns out that the spatialite website has changed, and the URL we had now returns a 404 error. I've updated the URL, and - because the folder inside the zip file has changed too - I've updated the `set_paths.bat` file.

I've also added error-handling to check if the extraction of spatialite fails, and if so, to stop the deployment script.

This should fix it now - can you check it works for you @IanMayo? (You'll need to clone this PR and then run the create_deployment.bat script, then take the zip and test deploying it).

I will add a separate issue to add more robust error-handling to the deployment process, and the batch files that the user runs.

### DIFF
--- a/bin/set_paths.bat
+++ b/bin/set_paths.bat
@@ -13,7 +13,7 @@ set PEPYS_PATH_PYTHON_ABS=%CD%
 popd
 
 REM Get absolute path to folder with spatialite DLLs in it
-set PEPYS_PATH_SPATIALITE=..\lib\mod_spatialite-NG-win-amd64
+set PEPYS_PATH_SPATIALITE=..\lib\spatialite-loadable-modules-5.0.0-win-amd64
 pushd %PEPYS_PATH_SPATIALITE%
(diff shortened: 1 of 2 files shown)


IanMayo:
Great - thanks @robintw  - will re-make release in the morning.

IanMayo (approved):
Great, thanks.  Using update to make build tomorrow morning will offer additional check

PR #580, opened 2020-09-17
Improve error-handling for deployment and execution batch files

robintw:
This PR improves error handling in our Powershell and batch files. Specifically, it:

 - Adds significant error handling to the `create_deployment.ps1` file, so that we now get alerted to any errors creating the deployment, and the deployment stops as soon as it encounters an error. This should significantly reduce the likelihood of creating an invalid deployment zip file
 - Adds error handling to the `set_paths.bat` file, so that if required paths are not found (due either to an invalid deployment zip, or a failed deployment) then it exits with errors
 - Adds error handling to all of the other batch files, like `pepys_import.bat` and `pepys_admin.bat` to catch errors that come from `set_paths.bat` and then stop before running any of the Pepys code. This means the user will get an error from `set_paths.bat` and then the program will stop (using the `PAUSE` command to print a 'Press Any Key to Continue' message). Thus, the user will know about the error immediately, and not run into problems in the Pepys Python code itself.

### DIFF
--- a/bin/install_pepys.ps1
+++ b/bin/install_pepys.ps1
@@ -9,7 +9,7 @@ $WshShell = New-Object -comObject WScript.Shell
 # Get the User's Send To folder location
 # This is safer than using a hard-coded PATH as network/user settings may mean the folder
 # is in an unexpected place
-$sendto_location = $WshShell.SpecialFolders("SendTo")
+$sendto_location = $WshShell.SpecialFolders("SendTo2")
 
 # On Windows icons are specified as a path, followed by a comma and a zero-based index into
 # the icons inside the file (as, some files like DLLs can have multiple icons in them)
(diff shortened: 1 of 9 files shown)


bin/install_pepys.ps1
```diff
@@ -9,7 +9,7 @@ $WshShell = New-Object -comObject WScript.Shell
 # Get the User's Send To folder location
 # This is safer than using a hard-coded PATH as network/user settings may mean the folder
 # is in an unexpected place
-$sendto_location = $WshShell.SpecialFolders("SendTo")
+$sendto_location = $WshShell.SpecialFolders("SendTo2")
```
IanMayo, line 12: @robintw  - I briefly googled, but couldn't any reference for `SendTo2` folder name.  Can you include a URL to some online reference in the previous comment?

robintw: Aah, glad you caught that - it's a mistake. I'll fix it.

bin/install_pepys.ps1
```diff
@@ -20,82 +20,113 @@ $icon_string = "$icon_path,0"
...
+    # We need to use full paths here or the shortcut will assume everything is relative to
+    # C:\
+    $Shortcut.TargetPath = [System.IO.Path]::GetFullPath(".\pepys_import_sendto.bat")
+    $Shortcut.IconLocation = $icon_string
+    # If we don't set the working directory then we won't be able to import other DLLs or use relative paths
+    # to our Python executable
+    $Shortcut.WorkingDirectory = [System.IO.Path]::GetFullPath(".")
+    $Shortcut.Save()
+}
+catch {
```
IanMayo, line 38: the introduction of `try/catch` is great. I've just done some Googling, and have seen we can access a description of the error:`
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_try_catch_finally?view=powershell-7#accessing-exception-information

robintw: Ah great. I've added a print out of all the error information each time we catch something - as with Pepys itself, I have it printing out the exception details first, followed by our `ERROR: Couldn't do X` string.

IanMayo (approved):
Great, thanks.

PR #582, opened 2020-09-23
Fix 'hangs' during import process

robintw:
This PR fixes two instances where pepys_import appears to 'hang' during the import.

1. When exporting the highlighted file. A message is now displaying saying that is what is happening, and a progress bar displayed while iterating over the characters in the file.

2. When submitting the measurement objects to the database. These are now processed in chunks of 1000 objects each, and a progress bar is shown while iterating over those chunks. This is slightly less efficient in terms of database access, but gives the user significantly more insight into the processing, which is a good trade-off.

This PR also includes a few flake8 fixes for `data_store_utils.py` which were picked up by the new pre-commit hooks.


IanMayo (approved):
Tested and working with 16Mb test file of REP data

PR #590, opened 2020-11-03
[WIP] - Accept exactly 3 decimals for time field

### LINKED ISSUE #584: REP importer - decimal places
> The rep importer expects exactly three decimal places of accuracy on time - anything more or less causes a crash

We should probably start by creating a "breaking" unit test, then fix it.

BarisSari:
Fixes #584


robintw:
@BarisSari - I think you might have misinterpreted the original issue. I think the problem the user was experiencing was that pepys _already_ only accepted time values with either 0 or 3 decimal places, and that we want pepys to be able to accept any number of decimal places.

@IanMayo Can you confirm whether my understanding of this is correct?

BarisSari:
aah, I've read the issue again and your understanding sounded more sensible, @robintw. If @IanMayo approves it, I'll close this PR and open a new one.

IanMayo:
Hello @BarisSari

Yes, the issue was that when Pepys encountered a time with something other than 3 decimal places it was falling over.  

We need to relax that constraint.

Thanks for stepping in @robintw 

As @BarisSari  suggested, I'll close this PR.

PR #591, opened 2020-11-04
Delete Datafile entry if there have been errors importing the file

### LINKED ISSUE #587: EAG parser - action on error
> The eag parser might fail to import any states due to an error, but will still store the file in pepys, causing 'file already imported' behaviour when the analysts fixes the file and puts it back through

We need to investigate this.   It looks like the EAG parser should not create the data-file if there is an error.

Hmm, but I think we create the datafile in the database before we start loading the rows.  

Hmm.  Maybe we need to delete the datafile record if there's a testing failure.  Or, we only create the datafile once we've passed the tests.

robintw:
This fixes #587 by deleting the Datafile entry if there were errors importing the file. This means we don't have 'orphan' Datafile entries hanging around - which potentially causes problems for re-importing files with errors.

The actual change is about two lines long (just a call to a delete function) - the majority of the PR is fixing the tests as the number of datafiles has changed.

### DIFF
--- a/tests/config_file_tests/test_local_and_core_validators.py
+++ b/tests/config_file_tests/test_local_and_core_validators.py
@@ -170,7 +170,7 @@ class TestLocalTestsFails(unittest.TestCase):
             platforms = self.store.session.query(self.store.db_classes.Platform).all()
             self.assertEqual(len(platforms), 18)
 
-            # there must be one datafile afterwards
+            # there must be no datafiles afterwards - as all files gave errors
             datafiles = self.store.session.query(self.store.db_classes.Datafile).all()
             self.assertEqual(len(datafiles), 0)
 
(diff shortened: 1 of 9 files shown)


tests/config_file_tests/test_local_and_core_validators.py
```diff
@@ -172,7 +172,7 @@ def test_local_basic_tests(self):
 
             # there must be one datafile afterwards
```
BarisSari, line 173: Can you change the comment? It might be confusing.

tests/config_file_tests/test_local_and_core_validators.py
```diff
@@ -212,7 +212,7 @@ def test_local_basic_and_enhanced_tests(self):
...
```
BarisSari, line 213: Same as above

BarisSari (approved):
Thanks for the PR, @robintw. It looks good, I asked for only a few minor changes.

robintw:
Thanks @BarisSari - I thought I caught all of those, but I missed a couple. Fixed now.

IanMayo (approved):

PR #596, opened 2020-11-05
Create new platform for unrecognised IDs in EAG files

### LINKED ISSUE #585: EAG files - missing IDs
> The eag files we received had IDs that weren't in the header - forcing analysts to delete those lines. Normally it was only one or two lines per file.

We could either ignore unrecognised IDs (to match the above workaround), or create new platforms for unrecognised IDs as they're encountered.  I'm checking with clients.

I've just spoken with someone familiar with the issue (CF), and he's happy that we create new platforms when necessary.

BarisSari:
Fixes #585


robintw (approved):
Looks good, thanks.

IanMayo (approved):
Great, thanks.

PR #598, opened 2020-11-05
Add GH Actions CI for Windows and Linux

robintw:
Previously we were running our Continuous Integration on Linux on Travis. This PR switches to Github Actions and sets up CI for both Windows and Linux.

There are a _lot_ of commits to this PR - sorry! This came about because of lots of experimenting I had to do while getting things working on Windows. It's almost impossible to test CI locally, so I had to do lots of experimentation.

I was originally working on getting our Postgres tests working on Windows - but ran into problems due to all Github Actions runners running as Administrator, and Postgres refusing to start as Administrator (and we need to be able to run the Postgres executable as part of our tests). I came up with a solution, but in the end that solution caused other problems. So, in discussion with Ian, I've left that code in the repository but commented it out (in case we need to use it in future) and switched to just running non-Postgres tests on Windows.

As with our Travis tests, Github Actions integrates with CodeCov. I can't test this fully until this is merged and we disable the Travis integration - as I can't tell if the CodeCov comments we're getting are coming from Travis or Github Actions. I should be able to test this with a few simple test PRs after this is merged.

The bulk of this PR is the extra configuration files and setup Powershell scripts for Github Actions, but there are also quite a few changes to our existing tests to make them work on Windows. Descriptions of some of the issues I found running our tests on Windows are available at https://github.com/debrief/pepys-import/wiki/Common-issues-with-tests-failing-on-Windows - these should be avoided in future tests to avoid failures on Windows.

As well as running tests, this PR also creates a Pepys deployment zip file for each commit. This can be downloaded by clicking 'Details' next to the 'Create Deployment' check, and then clicking the Artifacts link. At the moment this is a double-zip file - that seems to be the only sensible way to do it, so it needs extracting and then only the inside zip file sending to the user/whoever. We could extend this to automatically create actual Github releases if we wanted - as that would save Ian manually creating releases - but we'd need to decide exactly how to do that (in terms of version numbers, when to run it etc). I can investigate that if you want.


BarisSari:
Hello, @robintw. I have a comment before reviewing the PR. It seems like you needed lots of commits to try GitHub Actions out. Do you think that squashing these commits would be better? Otherwise, there will be lots of unnecessary commits in the main branch.

robintw:
That's a good suggestion @BarisSari - but I thought it was a problem to squash commits that have already been pushed. It seems like it can be done though, so I'll have a go at it.

robintw:
There we go @BarisSari - that seems to have worked. I've tried to strike a balance between not squashing enough commits and squashing too many commits. I've left the commits that did 'useful things' like fixing tests to make them work on Windows, but have squashed the ones that were just fiddling with Github Actions. Does this look more sensible now?

BarisSari:
Yes, it looks more sensible, @robintw. Thanks!

tests/test_admin_cli.py
```diff
@@ -899,18 +903,25 @@ def setUp(self, patched_prompt) -> None:
 
     @patch("pepys_admin.snapshot_cli.input", return_value="test.db")
     def test_do_export_reference_data_postgres(self, patched_input):
+        # Delete test.db file first, in case it is hanging around from another test
+        # If we don't do this, we can get into an infinite loop
+        path = os.path.join(os.getcwd(), "test.db")
```
BarisSari, line 908: I think you can use `CURRENT_DIR` variable instead of calling `os.getcwd()` again and again.

BarisSari (approved):
Thanks for the hard work, @robintw. I asked for only a small change but I'm approving the PR.

IanMayo (approved):
Great piece of work, thanks @robintw 

> We could extend this to automatically create actual Github releases if we wanted - as that would save Ian manually creating releases - but we'd need to decide exactly how to do that (in terms of version numbers, when to run it etc)

Yes, this auto-release would be useful. I propose we do a release for any tagged commit on `master`, and we use the tag as the version number.

Aah, the [create-deployment](https://pepys-import.readthedocs.io/en/latest/deployment.html) guidance will need to be updated.  I think we still need the manual instructions (as a fallback), but also the new ones.

robintw:
I think this is now ready to merge @IanMayo.

Once it's merged, we should disable Travis integration (ideally disable not delete, but I don't know if that's possible). I suspect only you can do that as you're the owner of the repo. Once that is done I can check that CodeCov is working properly from Github Actions, and fix it if necessary.

IanMayo (approved):

PR #625, opened 2020-11-18
Increase benchmark time thresholds

robintw:
## 🚀 Overview: 
Tests were failing intermittently on Github Actions because of variability in the time taken to execute the benchmarking code. This increases the time threshold allowed for a couple of the benchmarks, to give us more flexibility before the test fails. We will still get failures if the time taken for the test increases significantly.

## 🤔 Reason: 
We want reliable tests.

## Confirmations


IanMayo:
It looks like the time constraints need to be relaxed a little more, @robintw

robintw:
This has all passed now and is ready to be merged @IanMayo.

IanMayo (approved):

PR #626, opened 2020-11-19
Set up automatic releases on Github Actions

robintw:
## 🧰 Issue
As requested in a comment on PR #598

## 🚀 Overview: 
Added Github Actions steps to automatically produce a release for tagged commits in the repository.

## 🤔 Reason: 
This means a release is automatically created using a 'known good' VM that is configured for producing releases. It saves Ian from producing releases manually. The release that is created is set in 'draft' mode by default - so it can be checked and brie release notes can be added to the body, before making it public.

Unfortunately the only way to fully test this is to merge it first (I've done tests that show that it should work, but we can't test with a tag on the master branch until this is merged). Therefore, I suggest @IanMayo merges this, and then merges develop to master and creates a test tag on the master branch. This should then create a draft release, which can be deleted afterwards.

## 🔨Work carried out:

- [x] Altered GH Actions configuration
- [x] Tested manually

## Confirmations

### DIFF
--- a/.github/workflows/create_deployment.yml
+++ b/.github/workflows/create_deployment.yml
@@ -2,8 +2,8 @@ name: Create deployment
 on:
   push:
     branches:
-    - master
-  pull_request:
+    - automatic-releases
+  #pull_request:
 
 jobs:
   create-deployment:
@@ -33,10 +33,9 @@ jobs:
 
       - name: Release
         uses: softprops/action-gh-release@v1
-        #if: startsWith(github.ref, 'refs/tags/')
+        if: startsWith(github.ref, 'refs/tags/')
         with:
-          name: Version ${{ steps.get_version.outputs.VERSION }}
           draft: true
-          files: Release.txt
+          files: ./*_pepys-import.zip
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


IanMayo (approved):

.github/workflows/create_deployment.yml
```diff
@@ -1,10 +1,11 @@
 name: Create deployment
 on:
   push:
```
IanMayo, line 3: @robintw - I presume we can reinstate the next two lines, so that only tags on the `master` branch trigger releases.

Well, once we're happy it's all working ok.

robintw: Actually, no - as git tags aren't actually related to branches. I thought originally they were - but it seems not, see https://stackoverflow.com/questions/14613540/do-git-tags-only-apply-to-the-current-branch#:~:text=Here's%20another%20way%20of%20putting,only%20ever%20identify%20a%20commit.

That is, there is no way to find only tags from the master branch - so I think this will run on _any_ tags (but as we only use tags for releases, I think that's fine).

IanMayo: Aah, this is one workaround: https://github.community/t/run-workflow-on-push-tag-on-specific-branch/17519/4

robintw: Ok, I'll create another PR to test that later.

PR #632, opened 2020-11-20
Return to previous menu after cancelling add new platform

### LINKED ISSUE #618: UX issue - user can't switch back to `search for a platform` after cancelling add new platform
I'm working with the Getting Started binder.  I'm reminding myself about the ability to specify a different platform name, and for that to be used as a synonym in the future.

I ran the import to load `gpx_1_0.gpx`, accepting all the defaults.

I then edited the script, to load `gpx_1_1.gpx`.  At the `Search for an existing platform` or `Add a new platform` dialog, I pressed 2 to add a new platform.  At that point I can't get access to `Search for an existing platform` - I'm trapped in `Add platform`.

image

BarisSari:
## 🧰 Issue
Fixes #618.

## 🔨Work carried out:
- [x] Tests pass

## 🖥️ Screenshot (UPDATED)
It's my first GIF trial. Sorry, if it's too slow :) 
screnshot

image
image

## Confirmations


robintw (dismissed):
Looks good, thanks.

IanMayo:
Thanks @BarisSari .

`2. No, make further edits`.  

Hmm, when we click on 2, we don't make further edits, we go back up a level (as requested in the issue).

The user is at the _Review_ stage - looking at what has been entered for the platform. They're reviewing it in case any of the fields are wrong. So, maybe they realise they have called it a `Frigate` when it should be a `Destroyer`.   If that happens, we should let them make the edit.

So, I think we need:
```
Create this platform?
1) Yes
2) No, make further edits
3) No, restart platform selection process
.) Cancel import process
```

If 2 is clicked, then the previously entered values should be displayed as the user moves through the questions.  At the moment prompts are empty (except for the suggested `Name`).

IanMayo (changes requested):
Thanks @BarisSari .

`2. No, make further edits`.  

Hmm, when we click on 2, we don't make further edits, we go back up a level (as requested in the issue).

The user is at the _Review_ stage - looking at what has been entered for the platform. They're reviewing it in case any of the fields are wrong. So, maybe they realise they have called it a `Frigate` when it should be a `Destroyer`.   If that happens, we should let them make the edit.

So, I think we need:
```
Create this platform?
1) Yes
2) No, make further edits
3) No, restart platform selection process
.) Cancel import process
```

If 2 is clicked, then the previously entered values should be displayed as the user moves through the questions.  At the moment prompts are empty (except for the suggested `Name`).

BarisSari:
Okay, @IanMayo. I'll change the menu.

IanMayo:
@BarisSari - we should introduce the same logic for adding new `Sensors`, please.

Once the Jupyter build is done, I'll have a play - and give further feedback.

I've just looked a the schema, and can't see any other instances where we should have this logic.  Once we're working with `Tasks` and `Participations` we may need it, but I _currently_ think they will only be edited via Pepys-Admin.

BarisSari:
Hello @IanMayo. Okay, I'm making the same changes for `Sensors`, too.

IanMayo (approved):

Issue #610, opened 2020-11-12
Introduce new `Pepys-Viewer` app


IanMayo:
## 🐞 Overview
For some users, we wish to let them use Pepys-Admin to view data, but not to have access (or require training in) the rest of the Admin operations.

This could be met by providing a command-line parameter that gives an (apparently) different application:
* different welcome ASCII art (Pepys Viewer)
* application goes straight into the `6. View data` command mode
* exiting from this menu exits the application, rather than going back up to root level

The `View` menu may grow in the future as we add support for custom queries.

![image](https://user-images.githubusercontent.com/1108513/98944453-f8cb0780-24e8-11eb-9fc2-75f3041a194c.png)

**Note** this task will require both some Python changes, plus some MS-Windows trickery to create the new shortcut.

In the future the clients could choose to restrict who has access to `Pepys-Admin`.

## 🔗 Feature

N/A

robintw:
Just a thought: do we want the user to be able to access the `Status` option from Pepys Admin too? If so, we'll need to think about how we provide access to that.

IanMayo:
Thanks @robintw - I think _maybe not_ in the first version.  As the custom app loads it would show the version.  Plus we're not expecting this type of user to be interested in the Alembic version.

Was there some other piece of `Status` information you thought they would benefit from?

robintw:
Just that it gives the number of items in each table etc, which may be useful.

We'll have to be careful if we're instructing non-technical users to view database tables from the View Data menu, as many fields are missed out of the view there (because of space limitations) so it might confuse them.

robintw:
Just a reminder: when this is implemented, we must implement a training-mode version of Pepys Viewer too (see #614)

IanMayo:
Fixed in #642

PR #642, opened 2020-11-27
Add Viewer mode

robintw:
## 🧰 Issue
#610 

## 🚀 Overview: 
Adds a new Pepys Viewer app, which only provides access to the View Data commands within Pepys Admin. This can be called from the command-line using `--viewer`, or from two new Start Menu shortcuts: `Pepys Viewer` and `Pepys Viewer (training mode)`.

## 🤔 Reason: 
For some users, we wish to let them use Pepys-Admin to view data, but not to have access (or require training in) the rest of the Admin operations.

## 🔨Work carried out:

- [x] Added --viewer shortcut, and made it load just the viewing interface
- [x] Added Windows Start Menu shortcuts
- [x] Tests pass

## 🖥️ Screenshot
Screen Shot 2020-11-27 at 13 46 11

Screen Shot 2020-11-27 at 13 46 48

## Confirmations

## 📝 Developer Notes:
This depends on #624, and that should be merged first. I have already merged in the branch for #624 into this branch, as they were both modifying lots of the same files and I needed to sort out all the conflicts. _Hopefully_ Github will do the sensible thing and only show the new commits here, once #624 is merged.


IanMayo:
Thanks @robintw  - could you have a think about what changes should be made to `read the docs` please?

I'm pretty sure we should include the new command line parameters, but there may be a need for some higher level content - such as introducing the concept of the viewer.

IanMayo:
@robintw  - it looks like there is an issue with Create Deployment. _Maybe_ prompt_toolkit has moved.

robintw:
This create_deployment issue seemed to be an issue with installing black using the latest version of pip. Pip have released a new version (20.3) today, which switches to a new package resolver which is a lot more strict than the previous version. Pip was checking the metadata for the black package it had downloaded and raised an error as some of the metadata versions didn't match. This only seems to occur on the embedded version of Python we use in the Windows deployment, not on a normal Windows Python.

This has been temporarily solved by pinning black to an older version in `requirements_dev.txt`. This older version (only two versions ago) has a Python wheel available on PyPI which means pip can just download that and use it, without needing to build it itself - skipping the part that causes problems.

I've raised an issue on the black Github repo at https://github.com/psf/black/issues/1847. I have a suspicion they're going to tell me it's pip's problem, so I might end up raising an issue for pip too. I'll keep an eye on it all and let you know what happens.

As part of investigating this, I've noticed that failures in running the actual `create_deployment.bat` script aren't being picked up properly by Github Actions as a failure - I think the return codes from the various scripts aren't propagating properly. I'll add a separate issue to fix that.

robintw:
Argh, ignore that - pinning to an older black version means black wants to reformat all of our files! I'll work on this more tomorrow.

IanMayo (approved):
I ❤️  the viewer logo.

Issue #583, opened 2020-09-23
Validation error on REP data


IanMayo:
We have a very large file of REP data.  The data was computer generated, without (obvious) room for human error.

But, it fails enhanced validation: 
![image (3)](https://user-images.githubusercontent.com/1108513/93997530-67ae9e80-fd8b-11ea-8c76-e0de0867b843.png)

We should investigate the cause of the error, in case it represents some edge case.

Test data: 
[bulk_tracks.rep.zip](https://github.com/debrief/pepys-import/files/5267218/bulk_tracks.rep.zip)

This issue is relatively low priority.

robintw:
Fixed in #650.

PR #650, opened 2020-11-30
Fix enhanced validator to not compare bearing/heading if ship stopped

robintw:
## 🧰 Issue
#583 

## 🚀 Overview: 
The enhanced validator compares the heading specified in the input file against a bearing calculated from the current location and the previous location. If they are significantly different, it raises an error. This PR changes the code so that the comparison doesn't happen if the vessel is stopped (ie. the previous location and the current location are the same). When comparisons were happening in this situation, they were showing a bearing of 0, which didn't match the heading - and so was incorrectly raising an error.

## 🤔 Reason: 
Fixes a bug.

## 🔨Work carried out:
- [x] Fix bug
- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_import/core/validators/enhanced_validator.py
+++ b/pepys_import/core/validators/enhanced_validator.py
@@ -62,9 +62,12 @@ class EnhancedValidator:
             prev_time = prev_object.time
 
             if location and prev_location:
-                self.course_heading_loose_match_with_location(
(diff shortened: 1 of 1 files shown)


BarisSari (approved):
Nice and clean change. Thanks, @robintw.

IanMayo (approved):

PR #653, opened 2020-12-01
Don't print tables twice in Pepys-Admin

### LINKED ISSUE #652: Status tables shown twice in Pepys-Admin
## Version

0.0.23

## 🐞 Description

When I use `Status` in pepys-admin the 3 sets of tables are shown twice.

Experienced issue locally & in Jupyter.

## 🔢 Steps to Reproduce
1.  If necessary run an import to populate the database
2. Run pepys-admin
3. Run `2. Status`
4. Look at what's listed.

## 🖥️ Screenshots
image

BarisSari:
## 🧰 Issue
Fixes #652.

## 🔨Work carried out:
- [ ] Tests pass

## 🖥️ Screenshot
image

## Confirmations

### DIFF
--- a/pepys_admin/admin_cli.py
+++ b/pepys_admin/admin_cli.py
@@ -84,16 +84,13 @@ class AdminShell(BaseShell):
 
         with self.data_store.session_scope(), handle_status_errors():
             measurement_summary = self.data_store.get_status(report_measurement=True)
-            report = measurement_summary.report()
-            print(f"## Measurements\n{report}\n")
+            measurement_summary.report("Measurements")
 
             metadata_summary = self.data_store.get_status(report_metadata=True)
-            report = metadata_summary.report()
-            print(f"## Metadata\n{report}\n")
+            metadata_summary.report("Metadata")
 
             reference_summary = self.data_store.get_status(report_reference=True)
-            report = reference_summary.report()
-            print(f"## Reference\n{report}\n")
+            reference_summary.report("Reference")
 
         print("## Database Version")
         command.current(self.cfg, verbose=True)
--- a/pepys_admin/admin_cli.py
+++ b/pepys_admin/admin_cli.py
@@ -13,6 +13,7 @@ from pepys_admin.snapshot_cli import SnapshotShell
 from pepys_admin.view_data_cli import ViewDataShell
 from pepys_import.utils.data_store_utils import is_schema_created
 from pepys_import.utils.error_handling import handle_status_errors
+from pepys_import.utils.text_formatting_utils import custom_print_formatted_text, format_table
 
 DIR_PATH = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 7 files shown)


pepys_admin/admin_cli.py
```diff
@@ -84,16 +84,13 @@ def do_status(self):
 
         with self.data_store.session_scope(), handle_status_errors():
             measurement_summary = self.data_store.get_status(report_measurement=True)
-            report = measurement_summary.report()
-            print(f"## Measurements\n{report}\n")
+            measurement_summary.report("Measurements")
```
IanMayo, line 87: @BarisSari  - so `report()` should not return a parameter now, I guess.

BarisSari: We are styling the table and printing it inside of `report()`. But, it still returns not customised text. That's why it was printed twice: one in` report()` (table headers in bold), another with normal` print()` line 88.

IanMayo:
I'm totally poking my note in here @BarisSari  - and doing a level of scrutiny I don't normally do...

Can I suggest a refactoring of `data_store.getStatus()`?

It is only ever called for one type of table at a time, and it contains duplication that carries a bad smell.

I'd like to suggest:
1. Add 3 new methods in `data_store`, 3 versions of this - one per table-type.
```
  def measurementTables(self):
    return self.meta_classes[TableTypes.MEASUREMENT]
```
2. Modify `get_status()` to receive a set of tables as its parameter. 
3. Modify `get_status()` to also take a table title
4. Modify `get_Status()` calls to use `get_status(data_store.measurementTables(), "measurements")`
5. Modify `TableSummarySet` to take a title field, so:
```
  table_summaries_set = TableSummarySet(tables, titleStr)
```
6. Modify `TableSummarySet.report()` to not require a title

There's a good chance the above can be improved further - I'm just following the "Boy Scout Rule" :-)

pepys_admin/admin_cli.py
IanMayo: Hello @BarisSari  - yes, I understand. So, now we should modify `report()` so that it doesn't return a value (including the doc-string).

BarisSari:
Yes, @IanMayo. You're right that it has duplicated code and when we use it, we use it only for one table type. Okay, I'll do refactoring but I want you to know that it will break many things :) It will affect many tests which depend on returned outputs of `report()`. But, I agree that get_status() needs a refactoring. So, I think it will be worth it.

IanMayo:
Hello @BarisSari - aaah, I see.    In the tests we validate the contents of what is returned.  In that case I think the tidy solution is to return to the previous strategy.

`report()` should return text, then the _caller_ should call `custom_print_formatted_text(report)`.

BarisSari:
Okay 👍 I'm working on it

pepys_import/core/store/data_store.py
```diff
@@ -1644,3 +1618,12 @@ def get_logs_by_change_id(self, change_id):
             .filter(self.db_classes.Log.change_id == change_id)
             .all()
         )
+
+    def get_measurement_tables(self):
+        return list(self.meta_classes[TableTypes.MEASUREMENT])
+
```
IanMayo, line 1624: Hello - you're going to be cross with me here, I'm not trying to annoy you 🤣 

I've just realised we can remove even more duplication.

if the methods return a `TableTypes` enum value, then in `get_status()` we do:
`tables = list(self.meta_classes[tableType])`

Hmm, actually it can get even simpler. We no longer need those 3 methods.  We just pass `TableTypes.MEASUREMENT` to `get_status()`.

BarisSari: I was trying to find a middle way between your suggestion and minimum change of code (without breaking anything)😄 Yes, we can directly pass table type. I'm changing `report()` now. After that, I'll change `get_status()` again.

robintw (approved):
Looks good - thanks for doing the refactoring at the same time.

IanMayo (approved):
Nice work, thanks @BarisSari

PR #654, opened 2020-12-01
Add Access to contextual help from Pepys-Import & Pepys-Admin

### LINKED ISSUE #627: Access to contextual help from Pepys-Import & Pepys-Admin
## 🐞 Overview

The clients are developing best practices for how to word some fields, such as letter case, plus whether to use `HMS` prefix.

We _could_ include this in the UI.  Another option is to provide it a contextual help.  _Maybe_ at any question, entering `?` will open the contextual help for that field.

Please investigate if this is possible.  I propose that the each question is given an ID, and a CSV lookup table maps ID values to browser URLs.  These URLs will typically link to a page (poss with `#` anchor) in the local ReadTheDocs content.

Hmm, as an alternative.  The `CSV lookup table` could contain ID in column one, then the textual guidance in the second column.  After entering `?` or `HELP` this paragraph of text is displayed in the terminal (possibly in another color), with the prompt re-displayed.

Amongst all her other jobs, I'd appreciate @clenoble indicating her preference of content-in-browser, or content-in-terminal.

## Update
I guess we need:
* a method to insert a _New section_ divider. The method will take a `title` parameter, which gets displayed at the centre of the divider. Should look like the first one [here](https://github.com/debrief/pepys-import/issues/627#issuecomment-732243304).   Note the help reminder immediately after the divider
* a "framework" where prompts include an ID string.
* a table for HelpTexts: ID field plus guidance field. Guidance can be up to 2000 chars
* table can be populated from a `.csv` file via `Initialise data` in Pepys-Admin
* entering `HELP` or `?` at a prompt retrieves the guidance, displays it in italics, then provides original prompt.
* test-suite modified to verify that help present for every ID string (this will ensure we provide the new guidance when we add a new prompt).

## 🔗 Feature

#595 Improve user guidance

## 🔢 Acceptance criteria

During import process, entering `?` as answer for any question will either open the contextual help for that field in the browser, or display the content in the terminal.

BarisSari:
## 🧰 Issue
Fixes #627 

## 🔨Work carried out:
- [x] a method to insert a New section divider. The method will take a title parameter, which gets displayed at the centre of the divider. Should look like the first one here. Note the help reminder immediately after the divider
~a "framework" where prompts include an ID string.~
- [x] Add `help_id` and `search_help_id` parameters to `CommandLineResolver` methods.
- [x] Create a `constants.py` in resolvers directory. Use these numbers for help IDs instead of magic numbers.
- [x] a table for HelpTexts: ID field plus guidance field. Guidance can be up to 2000 chars
- [X] implement `DataStore.add_to_help_texts()`
- [x] table can be populated from a .csv file via Initialise data in Pepys-Admin
- [x] allow "HELP" and "?" inputs
- [x] entering HELP or ? at a prompt retrieves the guidance, displays it in italics, then provides original prompt.
- [x] Change dummy guidance texts in CSV file
- [x] test-suite modified to verify that help present for every ID string (this will ensure we provide the new guidance when we add a new prompt).
- [x] Tests pass
- [ ] Decide the style of help texts

## Confirmations


BarisSari:
@IanMayo, I have a few questions: 
- In which places there will be new section headers? I assume that at the beginning of each resolve operation, there should be one. Is that correct?
- Wrapping `prompt()` for an ID field will be a little bit difficult. Instead, I'm thinking to send question ids to the necessary methods. Let's say, we are resolving a platform and it calls `fuzzy_search_privacy()`. I'll send `question_id` (which has privacy guidance) to this method and it will show the related help texts. Does it sound sensible?

By the way, I'll write guidance texts to `Help Text.csv` but it might be better if you update it when it's done 🙂

IanMayo:
Thanks @BarisSari 

> In which places there will be new section headers? I assume that at the beginning of each resolve operation, there should be one. Is that correct?

Yes, at the start of each `resolve`.  

>  I'll send question_id (which has privacy guidance) to this method and it will show the related help texts. Does it sound sensible?

Sure, that sounds fine.

pepys_import/database/default_data/Help Texts.csv
```diff
@@ -0,0 +1,20 @@
+id,guidance
+1,RESOLVE_DATAFILE_TYPE
+2,RESOLVE_PRIVACY
+3,RESOLVE_DATAFILE
+4,RESOLVE_PLATFORM
```
IanMayo, line 5: @BarisSari - this isn't quite the structure I was expecting.  I was hoping for a legible ID.

```
id, guidance
RESOLVE_NATIONALITY,  The nationality of the platform.  Note: while a platform may have a fictional nation created  for exercise purposes - this should be the true nation it belongs to.  The fictional role undertaken by it should be captured within the Participation data.
RESOLVE_SENSOR_TYPE, The category of sensor.  These categories are steadily maturing, but please see an expert user if the relevant category isn't present.
```

BarisSari: When you said ID, I thought you meant Integer type. Yes, I think this structure is more sensible. I'll make changes for that.

By the way, I didn't write the guidance texts, yet. Please don't pay attention to that.

IanMayo:
@BarisSari it's going to be challenging to work out which specific prompt relates to which line in the `.csv` file.

I've added a comment stating that I'd like the _keyword_ to be used as the ID.   We could help navigate the help terms if we output this in brackets at the end of the printed help.

So, if I'm at a prompt and I want to "fix" the issue. I can enter `?` to view the help, and in that help I'll see the ID of the row I have to go to in the `.csv` (or in the table if _one-day_ we're doing it from inside `admin`).

BarisSari:
@IanMayo, yes it might help. I'll print the ID of the row as well. When I'm done, I'll give you many screenshots with small differences to decide the best style.

BarisSari:
@IanMayo, I have changed the Help Texts table. You can view it [here](https://github.com/debrief/pepys-import/blob/899d71a8536ed510e10b0337f8c93ac49fb15e16/pepys_import/database/default_data/Help%20Texts.csv
). I always find defining these kind of things difficult 🙂  So, you're welcome to update the guidance texts.

IanMayo:
> @IanMayo, I have changed the Help Texts table. You can view it [here](https://github.com/debrief/pepys-import/blob/899d71a8536ed510e10b0337f8c93ac49fb15e16/pepys_import/database/default_data/Help%20Texts.csv). I always find defining these kind of things difficult 🙂 So, you're welcome to update the guidance texts.

Yes, I feel your pain.

BarisSari:
Thanks for your understanding, @IanMayo.

I'm adding screenshots with some small differences. Please take a look at them. I'm testing them on MacOS for now. I'm open for suggestions:

----
#### Guidance text printed in Italic and it's not wrapped. It uses full width of the terminal (default - 80 width):
image
image
image

#### Guidance text printed in Italic and it's wrapped. It uses 60 width (chars per line) of the terminal (default - 80 width):
image
image

----
#### Guidance text printed in Italic and it's wrapped like above. Terminal width is 120 this time:
image

----
#### Guidance text printed in another color (I used color of Bootstrap's info button) and it's wrapped like above. Terminal width is 80:
image
image

----
#### Guidance text printed in another color and it's wrapped like above. Terminal width is 80, upper dashed line deleted:
image
image

IanMayo:
I have an issue where the help text is _hidden_ because the repeated question is shown after the help text, making it scroll back up the screen.

screencast

Solutions?
1. Put the help text in a different color (like @BarisSari  did with the blue text in the demo)
2. Let the user `<Press any key to continue>` before re-displaying the question.

I'm drawn to the second one. This is because it displays the guidance at the bottom of the command prompt (where the user eye/focus is), and only carries on once the user has read the text.

IanMayo (changes requested):
Thanks @BarisSari  - I'm not looking forward to updating all that help guidance :-)

One outstanding issue: https://github.com/debrief/pepys-import/pull/654#issuecomment-738165249

Would you mind implementing the "Press any key to continue" strategy?  It looks a really small change -  so it's not a great problem if we end up with another solution.

BarisSari:
Okay, I'm going to do that today

clenoble:
@IanMayo I would prefer to implement both solutions: display the help text in a 3rd color (pink or cyan) and <Press any key to continue>. The color will make easy to differentiate the help text from the instructions.

We also need to consider that on Windows, the terminal windows can take the height of the entire screen, so it is much more likely that the user will still view the help even once the question is added. This is a problem for binder only (and therefore only for training without windows install).

clenoble:
@IanMayo @BarisSari With color, there is no need to put the text in italic. Italic is not as readable as normal text so it should be used only rarely.

IanMayo:
Ok @Baris - we have way ahead.
1.  After displaying help, show "Press any ket to continue" before question is repeated
2. Display help guidance in pink, but not in italics

BarisSari:
Thanks, @IanMayo and @clenoble. I'm making changes at the moment. But I'll add `press enter to continue`, which is much easier than `press any key to continue`. I will search for it tomorrow morning.

IanMayo:
Actually @BarisSari - pressing `Enter` isn't too bad.  In their normal use, the analysts are using the `Enter` key for almost every interaction.

Actually `Enter` was the last key they pushed, in order for the help to open 🤣

IanMayo (approved):

PR #663, opened 2020-12-02
Fix Enhanced Validator to check if speed too high or too low

### LINKED ISSUE #659: Enhanced validation doesn't fail when speed too fast
I'm playing around with sample data.  I want a data-file to change.

I've modified this file to have a speed that is too much for the time step (200, rather than 2):
image

I'm _mocking_ the fact that a zero went in, instead of a decimal point in the highlighted line.

But, it passes enhanced validation.  It fails validation if I use a value of 0.0002

So, it looks like we're testing speeds that are too small, but not speeds that are too large.

Note: the above test file is in branch `feature/657_tutorial_data`

robintw:
## 🧰 Issue
Fixes #659 

## 🚀 Overview: 
Fixed the enhanced validator speed check to check if the speed in the input file is either too high or too low. To pass, the speed calculated from the location values must be between `measured_speed / 10` and `measured_speed * 10`.

## 🤔 Reason: 
Fixes bug, makes enhanced validator more robust.

## 🔨Work carried out:

- [x] Change if statement in Enhanced Validator
- [x] Add explicit tests: one with speed too low, and one with speed too high
- [x] Tests pass

## Confirmations


pepys_import/core/validators/enhanced_validator.py
```diff
@@ -169,7 +169,9 @@ def speed_loose_match_with_location(
         """
         distance = distance_between_two_points_haversine(prev_location, curr_location)
         calculated_speed = distance / time
-        if speed is None or calculated_speed <= speed * 10:
+        if speed is None:
+            return True
+        elif (speed / 10) <= calculated_speed <= (speed * 10):
```
IanMayo, line 174: wow @robintw - I didn't think it would be possible to represent the logic in this way.

robintw: Yeah, Python is cool 😄

BarisSari (approved):
Looks good. Thanks, @robintw.

PR #665, opened 2020-12-03
Fix GH Actions interactive review double commenting

robintw:
## 🧰 Issue
Reported on Slack by Baris.

## 🚀 Overview: 
The Github Actions workflow that provides interactive review was leaving multiple comments on PRs when multiple labels were added. This was because the workflow was only checking whether the `interactive_review` label was _present_ on the PR, not whether that label was the one that had been added (and thus triggered the workflow).

This PR fixes that, by checking more explicitly for the two conditions. We want to run the workflow if either:
 - This is a 'labeling' event (ie. a label has been added), and the label added is `interactive_review`, OR
 - This is a 'synchronise' event (ie. commits have been pushed to the PR) and the `interactive_review` label is present on the PR

It doesn't seem possible to do this in the trigger configuration itself, so we have two steps that check those conditions, plus another step that combines these into a single boolean flag, which the rest of the workflow uses as a condition to check whether to run.

## 🤔 Reason: 
Multiple comments are unnecessary and annoying.

## 🔨Work carried out:
- [x] Updated GH Actions workflow file
- [x] Tested manually in all conditions

## Confirmations


IanMayo (approved):
Great, thanks.

PR #668, opened 2020-12-03
try hosting the iframes in our repo

IanMayo:
MyBinder frames don't want to be embedded in pages from other domains:
image

See if we can host frames within MyBinder

### DIFF
--- a/demos/index.html
+++ b/demos/index.html
@@ -1,0 +1,1 @@
+<h1>Hello world</h1>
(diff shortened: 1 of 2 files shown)


IanMayo:
Nope, doesn't work.  It appears that we're getting the same error about frame parents.

Here's a direct link:
https://mybinder.org/v2/gh/debrief/pepys-import/hotfix/twin_frame_in_Jupyter?filepath=demos/twin.html

IanMayo:
Drop this, now we're favouring MS PPT & MS-Windows.

PR #674, opened 2020-12-07
Show one row for each platform when searching for a platform

### LINKED ISSUE #658: Triplication of results when searching for a platform
## 🐞 Description
Search suggestions include 3 times a given platform using different IDs.

## ⚙️ Current behavior
When searching for a platform name, a given platform is listed 3 times in the suggestions, once with the name, once with the trigraph, once with the quadgraph. 

## ⚙️ Expected behavior
A given platform should appear only once in the list of suggestions, in a way that includes the name, trigraph, quadgraph and synonyms if they exist.

## 🔢 Steps to Reproduce
In training mode, create a first platform.
For the 2nd platform, search using the 1st letter of the 1st platform created.
The 1st platform appears 3 times in the suggestion

## 🖥️ Screenshots
Screenshot_2020-12-02 Demo Training Mode - Jupyter Notebook

BarisSari:
## 🧰 Issue
Fixes #658 

## 🔨Work carried out:
~~Omit characters if the length of suggestion is more than 80 chars~~ (It already puts "..." if the line is longer than the terminal width. Please check the last screenshot)
- [x] Tests pass

## 🖥️ Screenshot
image
image
image
image

## Confirmations


robintw (commented):
Looks good.

My one concern is if we end up with so many synonyms that the synonyms part of the line gets so long that the other bits (such as nationality) get chopped off. Should we limit it to the first few synonyms? What do you think @IanMayo? Are they likely to have lots of synonyms for a platform? Of course, if we do limit it to the first few synonyms, then the user wouldn't be able to search by those synonyms, which may be a problem.

IanMayo (dismissed):

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -450,17 +449,25 @@ def fuzzy_search_platform(
         completer = list()
         platforms = data_store.session.query(data_store.db_classes.Platform).all()
         for platform in platforms:
-            completer.append(
-                f"{platform.name} / {platform.identifier} / {platform.nationality_name}"
-            )
+            p = f"{platform.name} / "
```
IanMayo, line 452: @BarisSari - would this code fail if the users used a '/' in a field?

Wouldn't it be more robust to build up a list?

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -479,17 +486,16 @@ def fuzzy_search_platform(
...
-            name_or_xgraph, identifier, nationality = choice.split(" / ")
+            platform_details = choice.split(" / ")
+            name, identifier, nationality = (
+                platform_details[0],
+                platform_details[-2],
+                platform_details[-1],
+            )
             # Get the platform from the database
             platform = (
                 data_store.session.query(data_store.db_classes.Platform)
```
IanMayo, line 497: @BarisSari - we used to perform an "or" operation here, the successive `filter` calls now look more like an "and".

Could you describe why we're changing our strategy please?

pepys_import/resolvers/command_line_resolver.py
BarisSari: Yes, it might fail because we split the selected option by "/" character. I'm changing the strategy. You're right that it would be more robust if we had a list

pepys_import/resolvers/command_line_resolver.py
BarisSari: Sure. `or` operation was used because some rows didn't have platform name. So, platforms were filtered by trigraph or quadgraph when rows without a platform name were selected. But right now, it is not possible to select a line which doesn't have a platform name. That's why I removed it.

pepys_import/resolvers/command_line_resolver.py
BarisSari: Actually, I checked the code again and it doesn't fail when user used a "/" in a field. Because we split by " / ", not only "/" characters. So, that's why we didn't have any problems until now. Do you still think that we should change it to list?

IanMayo: Yes, I still think we should switch to a list, please.  It looks much more robust.

BarisSari: okay

IanMayo:
> What do you think @IanMayo? Are they likely to have lots of synonyms for a platform? 

No, I'm not expecting a lot of synonyms.  It will probably just be something like accepting `HMS Nelson` for `Nelson`.

IanMayo (approved):

PR #683, opened 2020-12-07
Improve Run SQL mode

### LINKED ISSUE #679: Include examples of SQL strings in 2 dialects
I've fallen over a couple of times when trying to construct a PostGres SQL string.

Could the "Run SQL" command include (simple) examples of query strings in the 2 dialects please?

### LINKED ISSUE #681: Run SQL in Viewer Mode should only allow SELECT
In `pepys-admin`, RUN SQL is necessarily able to run any command - since it's the ultimate fallback in resolving issues.

But, in `Viewer` mode, the user should only be able to view data - and not make any amendments.

I'm no SQL expert, but a simple version of this could be:
* after trimming whitespace, check the first word is `SELECT`
* ensure the line doesn't contain any text after a `;` character (to prevent a second statement that does `Delete`)

robintw:
## 🧰 Issue
Fixes #681
Fixes #679 

## 🚀 Overview: 
Improves the Run SQL option in Pepys Admin/Viewer by:

- Adding guidance text, giving examples of queries
- Only allowing SELECT queries if you're in Pepys Viewer

Example showing guidance text:
image

Example showing what happens if you run a non-SELECT query in Pepys Viewer
image

## 🤔 Reason: 
- Provides more guidance to the user
- Stops Pepys Viewer users from accidentally changing entries in the database

## 🔨Work carried out:
- [x] Added guidance text
- [x] Added function to check whether a SQL query is just a SELECT query or not
- [x] Added while loop to keep checking if SQL query is valid or not
- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_admin/view_data_cli.py
+++ b/pepys_admin/view_data_cli.py
@@ -183,20 +183,49 @@ class ViewDataShell(BaseShell):
                 f"{selected_table} table is successfully exported!\nYou can find it here: '{path}'."
             )
 
+    def _check_query_only_select(self, query):
+        stripped = query.strip()
+
+        split_by_whitespace = stripped.split()
+
+        # If it doesn't start with SELECT then it is not acceptable
+        if split_by_whitespace[0].upper() != "SELECT":
+            return False
+
+        split_by_semicolon = stripped.split(";")
+        # If there is anything except whitespace after a ; then it is not acceptable
+        if len(split_by_semicolon) > 2 or split_by_semicolon[1].strip() != "":
+            return False
+
+        # If we've got here then we're ok
+        return True
+
     def _get_sql_results(self):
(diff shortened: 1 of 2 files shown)


IanMayo:
@robintw - do we need both sets of quotes in Postgres mode? I thought it was only one set.

robintw:
You're right @IanMayo - we don't need the quotes for the schema name. I've fixed it.

(By the way: [this page](https://www.prisma.io/dataguide/postgresql/short-guides/quoting-rules) explains some of the quoting rules. It seems that we have to quote the table name as we've started it with an uppercase letter).

IanMayo (approved):

IanMayo:
Had a play - checks worked fine.

PR #684, opened 2020-12-07
Improve training mode

### LINKED ISSUE #682: Training mode improvements
## 🐞 Overview
Move training db to folder in home directory
Change archive location in training mode to point to that directory, so that highlighted files etc get put there

## 🔢 Acceptance criteria
- Output files (output logs, error logs and highlighted files) all go to folder in home directory

robintw:
## 🧰 Issue
Fixes #682 

## 🚀 Overview: 
Switched training mode to store everything in a `Pepys_Training_Data` folder in the user's home folder and added specification of a custom archive folder (an `output` folder inside the above folder) where output/error logs and highlighted files will go

Example output from Pepys Import showing that the archive files are stored in the Pepys_Training_Data folder:
image

Example folder structure:
image

## 🤔 Reason: 
This means that output reports etc won't be stored in the main global archive folder - thus keeping all the training outputs separate.

## 🔨Work carried out:

- [x] Switched training mode to store everything in a `Pepys_Training_Data` folder in the user's home folder.
- [x] Added specification of a custom archive folder (an `output` folder inside the above folder) where output/error logs and highlighted files will go
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
I had to change the FileProcessor to do a `reload(config)` in its `__init__` function, as the config may have been changed since the config.py file was loaded by us going into training mode.


robintw:
I've had some issues with the tests for this PR, and I've come up with a solution that solves it - but wanted to get your views on the solution. It might be that we accept this solution and then come up with a larger change that would solve this problem more broadly later (see below).

When we enable training mode, we create a new config file with the relevant details in it, and then run `reload(config)` to reload the config file details into the constants in the `config` module. This worked fine when we were just altering the `DB_PATH` (etc) variables, but the changes in this PR mean we need to modify the `ARCHIVE_PATH` variable too. Because all of the modules are imported at the beginning of the Python process, this means we would also need to call `reload(config)` in the `__init__` function of the `FileProcessor` class, as without this it would still have the old config values that it loaded when the file was first parsed by Python.

_However_, calling `reload(config)` there means that we can't patch any of the config values that are used inside `FileProcessor`, (in a test) because as soon as a `FileProcessor` object is created we will reload the config module constants from the config file, and the patched values will be overwritten. Only two config file variables are used inside `FileProcessor`: `ARCHIVE_PATH` and `LOCAL_PARSERS`.

The only way I could see of getting past this was to extract these two variables into parameters for the `FileProcessor` class - so when it is created in `cli.py` it is created as:

```
processor = FileProcessor(
        archive=archive,
        skip_validation=skip_validation,
        archive_path=config.ARCHIVE_PATH,
        local_parsers=config.LOCAL_PARSERS,
    )
```

with those two arguments set from the config module. The `reload(config)` in the cli.py file will have already reloaded those values, and if they need to be patched by a test then the test can just call `FileProcessor(archive_path="Blah")`.

I'm not particularly keen on this solution, but I think it was the cleanest solution I could find at the moment.

This doesn't affect any other config file values at the moment, but it might in future depending what other work we do - and the underlying cause is how we manage the config file values. At the moment, `config.py` reads the config file when the config module is imported, and sets these constant values to values in the config file: that's why we have to call `reload(config)` to reload the values from the file - in this case `reload(config)` is Python code that actually reloads the module. I think it would be better to change how we handle the config file data, so that we have a specific function that parses the config file, and possibly change how we expose the actual config values. This would remove the need for `reload(config)` at all, and take all of this outside of the Python import mechanism, making it much simper. _However_, I don't think now (just before the training) is the time to do a major change to the config handling code - I think we should add it as a 'technical debt' item to work on when we have time.

What do you think @IanMayo, @BarisSari?

BarisSari (approved):
Looks good, thanks @robintw.

BarisSari:
> I'm not particularly keen on this solution, but I think it was the cleanest solution I could find at the moment.

I think your solution is the best solution for our current structure.

> I think it would be better to change how we handle the config file data, so that we have a specific function that parses the config file, and possibly change how we expose the actual config values. This would remove the need for reload(config) at all, and take all of this outside of the Python import mechanism, making it much simper. However, I don't think now (just before the training) is the time to do a major change to the config handling code - I think we should add it as a 'technical debt' item to work on when we have time.

Yes, we might read config values by calling a function. We didn't do it before because `config` is a module that should have global variables for whole uses. When we put it inside of a function, the state of config would be different for different modules of Pepys Import & Admin (according to the call time), and in my opinion, it wouldn't be robust. You're right that current `config.py` makes it difficult to test because of updated variables. But I think it should be like this.

IanMayo (approved):
> When we enable training mode, we create a new config file with the relevant details in it, and then run reload(config) to reload the config file details into the constants in the config module. 

I'd rather support training mode by creating a "real" file, but I recognise it's a relatively seamless way of integrating with the rest of the code.

As you suggest, I'm happy to treat it as technical debt to be handled when priorities permit,

PR #686, opened 2020-12-07
modify initial default data to suit training

### LINKED ISSUE #630: Initial data in training database
## 🐞 Overview

For running Pepys in training mode, it would be more realistic if there was some data in it already.

This task is initially to identify the specific data that should be present: which rows, in which tables.

Once that is done, it will switch to inserting/creating that data.

## 🔗 Feature

#595 

## 🔢 Acceptance criteria

- [ ] Training database created with sensible initial training data.

Note: here is the current reference data: https://docs.google.com/spreadsheets/d/1OlSXwRImHYOU6y62AL6KYQCbTY7ZIcxon17_SoAByGk/edit#gid=0
We should use a simplified version of that

IanMayo:
## 🧰 Issue

Fixes #630 

## 🚀 Overview: 

When we initialise a database, provide default data that:
1) matches what our Dec 2020 students would see in real life
2) helps users through tutorial (by creating some of the entities they will encounter, reducing the amount of human input required).

## 🤔 Reason: 

To reduce the amount of information that must be entered import process of tutorial.

## 🔨Work carried out:

- [x] Use trimmed down version of client lists for sensor-type & platform-type
- [x] Create metadata entries for first platform in tutorial (`ADRI`).

- [x] Tests pass

## Confirmations


IanMayo:
Hello @robintw & @BarisSari - I'd like to pick your brains on this, please.

In this PR I've modified the reference & metadata to create reference types similar to those actually in use, and a first platform/sensor in the metadata to make the first file import go more easily for the tutorial when starting from a blank database.

My changes will _almost certainly_ break the tests, but @BarisSari has a regexp-related strategy for handling that. 

My changes didn't work as expected though, since at database creation we only load the existing `reference` csv files, we don't load them for `metadata`.

I would like to propose that we also load existing metadata tables at database initialisation, for this reason:
* "Serious" organisations adopting Pepys won't be loading our metadata/reference data to their master database - they'll be be investing effort in _designing_ reference data to organise the data, and _importing_ existing metadata from their original sources (such as a list of UK warships).
* Anyone else that is just "investigating" (or playing with) Pepys would most likely want to do the tutorial anyway.

How does this sound?

robintw:
I think this sounds like a reasonable way forward. A few brief comments:

1. We'll need to make sure the users still have an experience of creating a platform during the tutorial - but I agree that not having to create one for the first exercise would be good.

2. We'll need to change a few other places that things are hard-coded. For example, the GPX Importer automatically assigns a Sensor Type of `GPS`. With the reference data in this PR, that should probably be hard-coded to `Location-Satellite`. In fact, that probably counts as a bug at the moment, if they're expecting to use Location-Satellite but actually get GPS.

3. It _might_ be worth having a parameter for the DataStore object as to whether to import metadata or not - used only for testing. It would make a number of our tests 'cleaner' if we were starting with no metadata - so that when we check the right number of metadata objects are present after testing we are checking for the `n` Platforms that are created by the test, rather than the `n + 2` platforms that are a combination of the pre-populated ones and the ones created for the tests. I see a number of tests are failing now for just the reference data change: I think a _very large_ number will fail if we add pre-populated metadata. Doing this would save us changing the tests again if we changed the default metadata in the future - and would make it far easier for tests that need to start from a 'blank slate' (for example, the merging tests would get rather difficult if they had pre-existing entries in those tables).

Ah, hang on - **update**: We can do this in a way that doesn't mean we have to have all the pre-populated data for the tests. `Datastore.initialise` just does the initialisation of the database, we then have `Datastore.populate_reference` and `Datastore.populate_metadata`. When we initialise the database during Pepys Import because the database is empty/non-existent then we deliberately call `initialise` and `populate_reference` - we could just add calling `populate_metadata` there.  Those functions are called inside `cli.py`, and so won't be called by any tests unless they're specifically testing the CLI - so we would have far fewer tests to change, and wouldn't need to update tests when we change the metadata contents.

So, we'd still have the tests that are currently failing to change - but most other tests could stay how they are, and would still be nice and 'clean'.

Does that sound sensible?

(This post is an example of me working things out as I type answers - sorry for the rather meandering text!)

IanMayo:
Thanks @robintw  - that all sounds fine.  Good news, in fact :-D

Yes, you're right about the GPX importer.

robintw:
This actually turned out to be a lot easier than I thought, and I _think_ it's all done now.

As expected, quite a few tests needed changing to deal with different reference and metadata being imported - but that was mostly changing the name of platform types (a lot of `Fisher`->`Fishing Vessel`), sensor types (`GPS`->`Location-Satellite`), platform names and so on. I also redid the 'missing data' set of CSV files to test missing data handling on CSV import - these needed to match the real ones, but with deliberate bits missing.

I have _deliberately_ not done a full search and replace for strings like `Fisher` or `PLATFORM-TYPE-1`. This is because in many places where those are used, they are still valid. For example, some code that does:

```
platform_type = self.store.add_to_platform_types("PLATFORM-TYPE-1", self.change_id)
```

still works fine. In fact, I think it is useful to have different type names there, as it shows us that it's something we have specifically added with a call like that, rather than something that we're assuming the database already has. Does that make sense? Basically it's less work (!) but also serves a useful purpose, I think.

BarisSari (approved):
I'm happy that we didn't need to change too many things.

IanMayo:
>  I think it is useful to have different type names there, as it shows us that it's something we have specifically added with a call like that

Yes, something that is **obviously wrong** should stand out.

PR #700, opened 2020-12-09
Fix file processing order so it is consistent on all platforms

### LINKED ISSUE #599: Different ordering of file processing on different platforms/systems
## Version
HEAD

## 🐞 Description
When a folder path is passed to Pepys, it processes all files in that folder. It seems that on different systems, it can process the files in different orders - which can lead to some tests failing if they depend on the order.

## ⚙️ Current behavior
The order in which files are processed differs between OS X and Linux, and possibly also between Travis Linux and Github Actions Linux. This causes problems when, for example, the first entry of a list is checked - but because the files were processed in a different order, it is actually the 3rd entry (for example) which is the one that should be checked.

## ⚙️ Expected behavior
Files should be processed in the same order regardless of platform and operating system, or tests should be designed which are not dependent on the ordering of files. Possibly both approaches should be taken.

## 🔢 Steps to Reproduce
It's easiest just to look at the [test results on Github Actions](https://github.com/debrief/pepys-import/actions/runs/347925850) which show three test failures. I've resolved these by getting each of these tests to just process a single file rather than a folder - which is probably a good thing to do anyway, as they don't need to process multiple files, and this speeds up the tests anyway.

## Discussion
I'm open to input as to whether we should do something with the file processing code to sort the file lists in Python before going through them, as the Python sorting functions should be platform-independent.

robintw:
## 🧰 Issue
Fixes #599

## 🚀 Overview: 
Fixed file ordering to force all the files to be sorted, with the same sorting on all platforms. The sample folder in the tutorial is now imported as `sample2` then `sample3` on both Windows and OS X.

## 🤔 Reason: 
Consistency.

## 🔨Work carried out:

- [x] Wrapped the `os.walk` function that is used to recurse into folders in a `sort_files` call
- [x] Extended `sort_files` to cope with strings or file objects
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
We had forgotten to use `sort_files` when we were recursing into directories using `os.walk` - I've fixed that, and extended `sort_files` a bit. This just meant that we weren't sorting files _at all_ when we passed a folder, and therefore we were getting arbitrary order on each platform (as os.walk provides files in arbitrary order).


BarisSari (approved):
Looks good, thanks @robintw

IanMayo (approved):

Issue #647, opened 2020-11-27
Research supporting import highlighting in XML


IanMayo:
Our import highlighting is really value for authors producing a new importer, or for an analyst who wants to double-check all the content got imported.

But, it doesn't work for XML, since we use an XML processing library.

This is a research task to consider how we could provide highlighting for the import.

I guess the options are:
1. extend our extraction code to support XML (though we'll need lots of code to handle the full range of XML :-( )
2. extend XML library to collect character index data necessary for producing highlighter 
3. use some new piece of code to "bind" the XML processing with our import highlighting

IanMayo:
@robintw & @BarisSari - either of you would be quite capable with this, though I think @BarisSari has more familiarity with the import highlighting.

But, I think the best outcome for this would be from one of you doing the research, and discussing findings with the other.  I don't mind who does what role 😸

robintw:
Actually it was me who worked on the import highlighting. I'll look into how best to do this and then discuss it with Baris.

robintw:
I've done some investigation into how to deal with highlighting for XML files, and have come up with a proof of concept that shows my plan works. There is a fairly rough notebook showing the proof of concept available [here](https://github.com/debrief/pepys-import/blob/highlighting-xml/XML%20Highlighting%20Playground.ipynb) but note that the highlighting output isn't actually showing properly on Github's view of the notebook. The actual output example highlighting for a tiny XML file looks like this:

![image](https://user-images.githubusercontent.com/296686/101615671-4cf0db00-3a06-11eb-9f11-8656706a8f16.png)

Now, I'll go into the various options I considered, and why I went down this route for the proof-of-concept.

My starting point was to decide that we definitely _do not_ want to implement our own XML parser using the `Token` objects that we already have provided through the highlighting functionality. XML is actually a fairly complicated format, and creating a parser that will work on a full range of real-world HTML files would be a _lot_ of work, and would likely end up being very fragile. Thus, we need to find a way to get character index information from an existing XML parsing library. Once we have this character index information (eg. that the content of the `<course>` tag starts at character 179 and finishes at character 183) we can easily fill in details in the character array stored as part of the `HighlightedFile` class, and get our highlighting working.

[...]

PR #714, opened 2020-12-11
Upgrade to Python 3.8

robintw:
## 🧰 Issue
In support of #647, as we need some Python 3.8 features for that to work

## 🚀 Overview: 
Upgraded to Python 3.8, by updating the CI and deployment scripts, and testing fully on Python 3.8 on Windows and OS X. A couple of tests have been updated to be a bit more flexible on numbers of tables being created to take into account different versions of spatialite.

## 🤔 Reason: 
- Allows use of some new ElementTree features in Python 3.8 for the XML parsing/highlighting. The key new feature is:

>The `.find*()` methods in the xml.etree.ElementTree module support wildcard searches like `{*}tag` which ignores the namespace and `{namespace}*` which returns all tags in the given namespace. (Contributed by Stefan Behnel in bpo-28238.)

- Keeps us up-to-date with Python: currently there are only security updates for Python 3.7, there will be no bugfixes to this version 

## 🔨Work carried out:

- [x] Tested on Windows and OS X with Python 3.8
- [x] Updated CI configuration for tests
- [x] Updated deployment script (which is also run on CI for making deployments) to download Python 3.8
- [x] Updated a couple of tests
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
**Important:** I would strongly suggest that all developers @BarisSari @IanMayo upgrade to using Python 3.8 once this is merged, so we're all on the same version.

Also useful to know, is that when doing debug print statements on Python 3.8 you can use a f-string like `f'{var=}'` to print out a nice representation of the variable with its name. For example:

```
>>> user = 'eric_idle'
>>> member_since = date(1975, 7, 31)
>>> f'{user=} {member_since=}'
"user='eric_idle' member_since=datetime.date(1975, 7, 31)"
```

### DIFF
--- a/.github/workflows/test_pepys.yml
+++ b/.github/workflows/test_pepys.yml
@@ -19,10 +19,10 @@ jobs:
           sudo apt update
           sudo apt install -y postgresql-13-postgis-3 sqlite3 libgdal-dev libproj-dev libproj-dev libsqlite3-dev spatialite-bin libsqlite3-mod-spatialite
 
-      - name: Set up Python 3.7
+      - name: Set up Python 3.8
         uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: 3.8
   
       - name: Install pip dependencies
         shell: bash -l {0}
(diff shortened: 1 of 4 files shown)


BarisSari (approved):
Thanks for the PR, @robintw. It looks nice. I'll use a new Python 3.8 virtual env when this PR is merged.

PR #717, opened 2020-12-14
Add support for highlighting XML files

### LINKED ISSUE #647: Research supporting import highlighting in XML

robintw:
## 🧰 Issue
Fixes #647 

## 🚀 Overview: 
Add custom XML parsing classes to allow recording extractions via highlighting on XML files, and update the GPX Importer to use these custom classes and do highlighting.

Example highlighting:
image

## 🤔 Reason: 
Highlighting was available for all file types except XML files, which was an anomaly - and having highlighting leads to far greater confidence from the clients that the importers are working properly.

## 🔨Work carried out:

- [x] Add custom XML parsing functions
- [x] Link custom XML parsing functions to HighlightedFile class using `record` method
- [x] Extend GPXImporter to use `record` calls from new parsing functions
- [x] Add tests for new custom XML parsing functions
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
There are quite a few changes in this PR, but I think I've documented them fairly well - please ask if there are any questions about how the new XML stuff works.

This implementation will be slower than the implementation that didn't use highlighting, but the speed doesn't seem to be an issue. I've done a test with a GPX file containing ~1700 states and it is still processed within a few seconds. @IanMayo is checking how large the GPX files they usually process are, and we can do more investigation into speed later if they tell us they use enormous files regularly.


IanMayo:
Nice piece of work @robintw .  I'm particularly impressed by the documentation of the new parser:
https://github.com/debrief/pepys-import/pull/717/files#diff-f93c5c5ee9b330caf7555bb84ec3635ceef1c05410dc0df843cd369d8d37d77d

IanMayo:
@robintw  - I tried to import a GPX file using the Binder.  I ran this command:
```
python -m pepys_import.cli --path tests/sample_data/track_files/gpx/test_land_track.gpx
```

But, the reports folder only contained the log, not the highlighted report.  Is highlighting switched off for GPX data?

BarisSari (approved):
Everything looks good, thanks @robintw. Out of curiosity, I have one question. It seems like we lost `sourceline` attribute of Element. Why did it happen? It seems like we still use the same `find`, `findall` methods with the original Element.

robintw:
Ah yes, @BarisSari: we were originally using the `lxml` library, which is based on the ElementTree API but provides a slightly extended version of it. `lxml` provides a sourceline attribute, but the ElementTree API that we've now switched to using doesn't. It should be possible to try and work out a sourceline based on the character that we're already extracting, but it would require searching for all newlines every time we wanted it, and it might slow things down quite a bit. On the other hand, we should only need the sourceline when we're recording an error, so the slowdown might not be too bad.

@IanMayo would you like me to try implementing that? Thinking about it, knowing which line the error is on would be quite useful in fixing it!

robintw:
@IanMayo It was a problem with Binder - I'd forgotten to update Binder to run on Python 3.8, and so some of the new code wasn't working properly. It now works fine on Binder - see screenshot below:

image

IanMayo:
Yes Robin, it would be useful to know that the line number where an error occurred.

IanMayo (approved):
Yes, working fine, thanks @robintw  - though it took me a few minutes to recreate your output to HTML, when I could have just copied the way you did it. Duh.

IanMayo (changes requested):
Changing to `request changes` to see if we can get the line number, for error reporting.

IanMayo (approved):

PR #718, opened 2020-12-15
Moving `Back` when creating new entity

### LINKED ISSUE #619: Investigate way of moving `Back` when creating new entity
@clenoble gave us some UX guidance:

> Having a back button gives you the reassurance to explore and possibly make 
> wrong choices without consequences. It's one of the most important principles of a good UX.

So, instead of going through all the steps to add a platform, then changing them at the end, Celine would like the user to be able to step back through the questions if, for example, they just realise they've entered the wrong thing.

This task involves investigating if its possible to have a special character (`.`) cause the process to move back a question, including repeatedly specifying it.

My first guess is that we would have to move from specifying the questions one by one in the code, to providing an array of `handlers`, each able to provide the question and store the response.  It would be this list of `handlers` that would be displayed at the end of the `add new item` process.

Note: having the questions in an array would also give us the ability to show `2/6` progress indicator for the user.

I _think_ that `add new item` is the only sequence of questions that would require `back` functionality.

The investigation should include the consideration of if the above can be implemented without introducing additional fragility.

BarisSari:
## 🧰 Issue
Fixes #619 

## 🔨Work carried out:
- [ ] Tests pass

## Confirmations


IanMayo:
Closing this issue, based on discussions on strategy. See discussion in #619

Issue #737, opened 2021-01-15
Suspected extra step in NMEA import


IanMayo:
## Version

0.0.25

## 🐞 Description

I'm testing a new release. I'm doing it by importing `NMEA_TRIAL.LOG`.

But, towards the end of the import process we go through the `RESOLVE PLATFORM` process again (see steps below).

## 🔢 Steps to Reproduce
Steps to reproduce the behaviour:
1. `python -m pepys_import.cli --path tests/sample_data/track_files/NMEA/NMEA_TRIAL.log `
2. Add new platform `PLAT`
3. Add a sensor `GPS`, search for sensor-type `LOCATION-SATELLITE`
4. Data gets parsed
5. Then I get invited to `Select a platform entry for PLAT` (see bottom of screenshot below).

## 🖥️ Screenshots

![image](https://user-images.githubusercontent.com/1108513/104727363-202e9280-572d-11eb-87c8-99c7a17ccff4.png)

IanMayo:
Fixed in #738

PR #738, opened 2021-01-15
Fix NMEA importer to stop asking user about platform twice

robintw:
## 🧰 Issue
#737 

## 🚀 Overview: 
The NMEA importer was asking the user twice about the platform, when it should only need to be told once. This change explicitly adds the returned platform to the platform cache, so it can be found next time (see developer notes below for more).

## 🤔 Reason: 
- Reduce user input needs

## 🔨Work carried out:

- [x] Add platform to the cache
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
This was tricky to debug because it seemed like the import was actually happening twice - there were two long progress bars, and Ian said in his original report that it was near the end of the import. Actually, the first long progress bar was the HighlightedFile characters array being filled (so we can do highlighting), and the second long progress bar was the actual import.

The code was like this:

```
platform = self.get_cached_platform(
    data_store, self.platform_name, change_id=change_id
)
# Keep track of the platform name, so we don't have to ask for each line
self.platform_name = platform.name
```
This is inside a loop over the lines in the file, and the first time round the loop we have `self.platform_name` set to `None`, which means the result can't be cached under a name - as we don't know a name until after we return from get_cached_platform. So, the next time round the loop we had a name, but the platform wasn't in the cache under that name. To fix this, we just add the line `self.platform_cache[self.platform_name] = platform` to explicitly put the platform in the cache under the new name.

### DIFF
--- a/importers/nmea_importer.py
+++ b/importers/nmea_importer.py
@@ -98,6 +98,10 @@ class NMEAImporter(Importer):
                 )
                 # Keep track of the platform name, so we don't have to ask for each line
                 self.platform_name = platform.name
+                # Explicitly store this new platform in the cache under the new platform name
+                # It won't be stored by default as the first time we called get_cached_platform
+                # we passed None as a name, and so it wouldn't have cached
+                self.platform_cache[self.platform_name] = platform
 
                 sensor = self.get_cached_sensor(
                     data_store=data_store,


IanMayo:
Thanks @robintw  - I've just tested it in Binder, and it worked perfectly.  Happy to Approve, when it comes out of draft status.

IanMayo (approved):

PR #740, opened 2021-01-19
Implement merge_platforms method

### LINKED ISSUE #739: 4.1 Merge Platforms action
## 🐞 Overview

The first maintenance task we're going to work on is `merge platforms`.

This issue is to perform the processing associated with merging platforms.
## 🔗 Feature

#600 

## API
Receives:
* list of platform IDs
* platform-id for the new `master` platform

Processing:
* move sensors from other platforms to the Target platform
    * if sensor with same name is already present on Target platform move measurements to that sensor
* move all comments from other platforms to the Target platform

Returns:
* success/failure error code

## Note
This is for the future, but I think it's worth being aware of.

The next release will include a `generic` merge capability - letting the user merge rows from `Reference` and `Metadata` tables.  Hopefully we will be able to avoid creating serparate merge methods for each specific table.  Hopefully we can use generic processing for the `Reference` tables - this may be possible through introspection.  For example, if `Nationalities` is being merged, go through tables and see which ones have `nationality-id` as a foreign key.  It's these rows that need to be changed to the "new" master Nationality.

BarisSari:
## 🧰 Issue
Fixes #739.

## 🔨Work carried out:
- [x] Create DataStore.merge_platforms() method
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:

I faced with a strange error while testing platform_id changes for Media table. I created the `Media` entities, and they were in the database. But I couldn't query them, it was always returning an empty list. That's why I updated the `platform` and `subject` attributes from `MediaMixin` class, which is inherited by `Media`. So, we can query Media objects now.


BarisSari:
@IanMayo, I have a few questions:

- We're merging Comments, States, Contacts, and Sensors of Platforms in this PR. Should we merge any other table's values?
- After the merge, should we delete the merged platforms and keep the master platform only?

IanMayo:
@BarisSari - I think that `Comments` is the only measurement table that we touch.   The other measurement tables have a `sensor`.  We keep the sensor, but change the sensor platform to the new one.

Oh no. We have a `subject` platform field in the `Contacts` table.  I don't think they have been used yet, but we should handle it in this merge action.

Hmm, while we don't import the data yet, we should update the `Platform` in `LogsHoldings`, `sensor_platform` and `subject_platform` in `Geometries`, and `platform`/`subject` in `Media`.

Actually, I don't think I know the correct answer there.  I know it will be hard to test tables that contain no data, or don't contain values in platform-related fields (yet).  Maybe we should check for their presence and throw an error in the merge code, indicating that merging them hasn't been implemented yet.

Any opinions @BarisSari , @robintw ?

The other one is easier. Yes, after the merge we should delete the merged platforms.

BarisSari:
>  I think that Comments is the only measurement table that we touch. The other measurement tables have a sensor. We keep the sensor, but change the sensor platform to the new one.

Yes, you're right. It's actually what I did. I couldn't explain it well :) Sorry about that.

> Oh no. We have a subject platform field in the Contacts table. I don't think they have been used yet, but we should handle it in this merge action.

I'm filtering Contacts by Sensor Platforms, but I will filter them with Subject Platforms, too. Then, I will update all objects.

> Hmm, while we don't import the data yet, we should update the Platform in LogsHoldings, sensor_platform and subject_platform in Geometries, and platform/subject in Media.

Yes, these fields are concerning me, and I think we should figure it out in this PR. For example, If I delete a platform without doing anything with these fields, we might lose them because of the referential integrity.

I can create dummy values in these tables, and write tests for merging platforms. It will make this PR a bit more complex, but I think we will eventually need it.

IanMayo:
Thanks @BarisSari  - wise advice.  Yes, let's create dummy data.

robintw:
This all sounds like a sensible way forward @IanMayo @BarisSari.

pepys_import/core/store/common_db.py
```diff
@@ -1354,29 +1354,17 @@ def sensor_name(self):
...
-            innerjoin=True,
-            uselist=False,
             foreign_keys="Media.platform_id",
         )
 
     @declared_attr
     def platform_name(self):
         return association_proxy("platform", "name")
 
-    @declared_attr
```
robintw, line 1368: Is there a reason this is removed? I wonder if, now we've got the platform relationship back in, can we add this back in too?

pepys_import/core/store/data_store.py
```diff
@@ -1647,3 +1647,99 @@ def get_logs_by_change_id(self, change_id):
...
+        Contact = self.db_classes.Contact
+
+        if isinstance(master_id, Platform):
+            master_id = master_id.platform_id
+
+        master_platform = (
+            self.session.query(Platform).filter(Platform.platform_id == master_id).scalar()
+        )
+        if not master_platform:
+            print(f"No platform found with the given master_id: '{master_id}'!")
```
robintw, line 1675: As this is going to be called from within a GUI, could you change this (and any other error messages below) to raise an exception instead? A `ValueError` with an appropriate error string would work fine here, and could be caught by the GUI to display the error nicely (the standard print statement doesn't play nicely with prompt_toolkit's GUIs, and will break various things...)

tests/test_data_store_merge_platforms.py
```diff
@@ -0,0 +1,680 @@
...
+                self.store.session.query(State).filter(State.sensor_id == sensor_2.sensor_id).all()
+            )
+            contacts_after_merge = (
+                self.store.session.query(State).filter(State.sensor_id == sensor_2.sensor_id).all()
+            )
+            assert len(states_after_merge) == 0
+            assert len(contacts_after_merge) == 0
+
+    def test_merge_platforms_with_different_sensor_names(self):
+        """Create two platforms, each platform will have a sensor named TEST-SENSOR.
```
robintw, line 138: I think this docstring is out of date with the test - this test has different sensor names

tests/test_data_store_merge_platforms.py
```diff
@@ -0,0 +1,680 @@
...
+                privacy=self.privacy.name,
+                change_id=self.change_id,
+            )
+
+            self.store.session.expunge(self.file)
+            self.store.session.expunge(self.comment_type)
+            self.store.session.expunge(self.platform)
+            self.store.session.expunge(self.platform_2)
+
+    def test_update_platform_ids_except_media(self):
```
robintw, line 493: I think this needs renaming now that it includes media

robintw (approved):
Looks good, nicely written. Just a few minor issues with out-dated docstrings/names etc.

pepys_import/core/store/data_store.py
BarisSari: Okay, I changed it: https://github.com/debrief/pepys-import/pull/740/commits/94ce9d8240683472553fd09c225987e45e4f70b3

tests/test_data_store_merge_platforms.py
BarisSari: Thanks for pointing out. I updated out of date docstrings and comments: https://github.com/debrief/pepys-import/pull/740/commits/94ce9d8240683472553fd09c225987e45e4f70b3

tests/test_data_store_merge_platforms.py
BarisSari: Thanks for pointing out. I updated out of date docstrings and comments: https://github.com/debrief/pepys-import/pull/740/commits/94ce9d8240683472553fd09c225987e45e4f70b3

robintw (approved):
All great now, thanks @BarisSari.

IanMayo (approved):

Issue #600, opened 2020-11-06
Database maintenance capabilities


IanMayo:
## 💬 Brief description of feature
Extend `pepys_admin` to include capabilities for developing & maintaining data.

This includes fixing broken data, and adding/maintaining reference data & metadata.

This (mostly) replaces the use of raw SQL commands to maintain data.

### Efficiency of analysis
* Avoid having to use SQL to resolve issues
* Task oriented perspective lets analyst work through tasks quickly

### Effectiveness of analysis
* Provide analyst with correct & consistent data
* Maintain audit log of maintenance interventions

## ☑️ Acceptance criteria
- [ ] Analyst able to perform required tasks without requiring use of SQL.

PR #743, opened 2021-01-26
Maintenance GUI - First Phase

robintw:
## 🧰 Issue
#600 

## 🚀 Overview: 
This is the first phase of the maintenance GUI, as described in https://docs.google.com/drawings/d/1WopPwQoUFGYnEZCQeglWsBalH7ye6ECWh2TyI0r7tfY/edit.

image

## 🤔 Reason: 
- Avoids analysts having to use SQL to resolve issues
- Task oriented perspective lets analyst work through tasks quickly
- 
## 🔨Work carried out:

- [x] Developed a number of standalone widgets to be used in the GUI, such as `FilterWidget`, `CheckboxTable`, `DropdownBox` and so on
- [x] Developed GUI
- [x] Added tests for the GUI, both running semi-interactively, and for individual methods/functions
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
As a quick way of getting into the GUI with the default initial data, you can run `python -m pepys_admin.maintenance.gui` from the root of the repo. Users will access the GUI through Pepys Admin (option 8).

Logging using loguru is disabled now, but I have deliberately kept the loguru dependency in the requirements.txt, so that if we need logging in future then it is easily available (and it also helps if we accidentally leave a log command in there, as we won't crash).

There is a _lot_ of new code in this PR, using new technologies for us such as prompt_toolkit and asyncio. The [fullscreen apps docs](https://python-prompt-toolkit.readthedocs.io/en/master/pages/full_screen_apps.html) from prompt_toolkit may be helpful to read.


pepys_import/core/store/common_db.py
```diff
@@ -136,6 +136,7 @@ def platform_type(self):
 
     @declared_attr
     def platform_type_name(self):
+        # return association_proxy("platform_type", "name", info={"name":"platform_type_name"})
```
BarisSari, line 139: I think this is a leftover. Can you delete it?

BarisSari (changes requested):
Thanks a lot for your hard work, @robintw. In general, it looks great! However, I think it would be better if we consider the following issues:

image

I think I have found a bug. When I first opened the GUI, I selected name, "=", and then pressed to ENTER without writing any value. And I have faced this error. After that, I couldn't write anything. The same error occurred repeatedly.

(Actually, I tried again, and it seems like I can't write anything in "Enter value here" after I select "name" in the first column. It returns to a dropdown menu which never opens.)

----

image

It seems like the position of `<add filter condition>` is visually wrong. Can we put it to the left side, or can we move it to the end of the right side like `< delete >`?

----

image

I don't know how I did, but I had this error :) I think I created two new conditions without filling the first. And it gave this error.

----

Maybe we can put examples in a directory called `examples`. It might be tidier.

robintw:
Thanks for the feedback @BarisSari. I think I've dealt with all of your changes now - I've fixed the two bugs you found (one to do with empty dropdowns - they now give a message 'No entries' - and one to do with adding > 2 empty filter rows), removed the old comment, and moved the examples to an `examples` directory.

I'm corresponding with the prompt_toolkit author about the positioning of the add button - he thinks it is a bug in the latest prompt_toolkit master.

pepys_admin/maintenance/gui.py
```diff
@@ -0,0 +1,876 @@
+import time
```
IanMayo, line 1: @robintw - it would be great it we could reduce the size of this file, if there's a tidy way of doing it.

I guess we could start off by pulling out the help content.

BarisSari (dismissed):
Thanks, @robintw. It looks good. But as Ian suggested, it would be better to split `gui.py`.

pepys_admin/maintenance/gui.py
robintw: @IanMayo I've shrunk it somewhat by removing the help text to `help.py`, and putting the `create_column_data` function into `utils.py` (actually, I'll be extending that function significantly in the work I'm going to do tomorrow/Friday, so it may move into a completely separate file after that). That has reduced it by a few hundred lines at least. I think pretty-much everything else makes sense to keep within the main class.

IanMayo: Great, thanks.

IanMayo (approved):
There's some ninja-level async wizardry going on with the UI code here @Robin - well done.

PR #744, opened 2021-01-26
Convert FilterWidget output into SQL Alchemy query

### LINKED ISSUE #741: API for converting FilterWidget output into SQLAlchemy query
## 🐞 Overview
The FilterWidget which will be used for database filtering in the GUI produces an output of the filters that are selected in the UI. The output looks something like this:

```
[["name", "=", "HMS Floaty"], ["AND"], ["Nationality.name", "=", "name here"]]
```

We need a function that converts this sort of filter specification, along with a table name (in this case `Platforms`) into a SQLAlchemy query. The data structure is a list of lists, where each entry is either a single element list containing `AND` or `OR` or a three-element list containing `column, operator, value`. The operators that are currently supported are `"=", "!=", ">", "<", ">=", "<=", "LIKE"` (with LIKE only for string columns).

The output is based on the input to the filter widget, which is a dictionary specifying column information - like this:

```
column_data = {
    "platform_id": {"type": "id", "values": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]},
    "name": {"type": "string", "values": ["HMS Name1", "HMS Floaty", "USS Sinky"]},
    "identifier": {"type": "string"},
    "nationality_id": {"type": "id", "system_name": "Nationality.nationality_id"},
    "nationality_name": {"type": "string", "system_name": "Nationality.name"},
    "timestamp": {"type": "datetime"},
    "speed": {"type": "float"},
}
```

You can see here that some columns have a `system_name` property, and that this name - if present - is the name that is returned in the FilterWidget output. This should allow for easier handling of denormalised columns - as the user will see `nationality_name` in the UI, but behind the scenes it will be output as `Nationality.name` (or whatever is configured as the system_name).

Note: At some point, we'll probably need to implement some code to take a table and create the column_data for it, while denormalising the table (to get things like `nationality_name`). If it seems helpful to implement this functionality as part of this issue then that'd fine - but it's not essential at this stage.

## 🔗 Feature
#600 

## 🔢 Acceptance criteria
- Working function to do conversion
- Comprehensive tests showing it works in a range of situations

BarisSari:
## 🧰 Issue
Fixes #741 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_admin/maintenance/widgets/filter_widget_utils.py
+++ b/pepys_admin/maintenance/widgets/filter_widget_utils.py
@@ -2,11 +2,11 @@ import operator
 from typing import List
 
 from pepys_import.core.store.data_store import DataStore
-from pepys_import.utils.table_name_utils import table_name_to_class_name
+from pepys_import.utils.table_name_utils import camel_to_snake, table_name_to_class_name
 
 operator_dict = {
     "=": operator.eq,
-    "!=": not operator.eq,
+    "!=": operator.ne,
     ">": operator.gt,
     "<": operator.lt,
     ">=": operator.ge,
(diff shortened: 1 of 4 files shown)


robintw (commented):
Thanks @BarisSari, this looks good.

Just one question: is the `camel_to_snake` function still used anywhere? I think it used to be used for converting from the system name to the declared attr name - but now we're just using the declared attr name, I don't think it is used any more. If it's not used, we can probably remove it.

BarisSari:
Thanks for the review, @robintw. You're right, we're not using it anymore. I deleted it

robintw (approved):

PR #746, opened 2021-01-28
Hotfix - Delete merged platforms

BarisSari:
## 🧰 Issue
Related to #739 

## 🤔 Reason: 
I forgot to delete merged platforms in the previous PR. This PR fixes this issue.

## 🔨Work carried out:
- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_import/core/store/data_store.py
+++ b/pepys_import/core/store/data_store.py
@@ -30,6 +30,7 @@ from pepys_import.utils.sqlite_utils import load_spatialite, set_sqlite_foreign_
 from pepys_import.utils.value_transforming_utils import format_datetime
 
 from ...utils.error_handling import handle_first_connection_error
+from ...utils.sqlalchemy_utils import get_primary_key_for_table
 from ...utils.text_formatting_utils import custom_print_formatted_text, format_error_message
 from .db_base import BasePostGIS, BaseSpatiaLite
 from .db_status import TableTypes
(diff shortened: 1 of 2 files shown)


IanMayo:
@BarisSari  - do we record the merge in the log/changes tables?  The `changes` table tries to link to an ID in the relevant table.   We have two options for recording the change:
1. Include change entries for deleted platform rows, knowing the ID will no longer be present
2. Just include an entry for the new `master` platform, including IDs of deleted platforms in the description.  Note: The IDs are long, so that could be very long entry if a dozen or so platforms are merged.

Any thoughts @robintw 

It's worth getting a clear understanding here, since it's effectively a precedent for how we log other maintenance changes.

BarisSari:
> @BarisSari - do we record the merge in the log/changes tables? 

No, @IanMayo. We're not recording.

> The `changes` table tries to link to an ID in the relevant table. We have two options for recording the change:
> 1. Include change entries for deleted platform rows, knowing the ID will no longer be present
> 2. Just include an entry for the new `master` platform, including IDs of deleted platforms in the description.  Note: The IDs are long, so that could be very long entry if a dozen or so platforms are merged.
> 
> Any thoughts @robintw
> 
> It's worth getting a clear understanding here, since it's effectively a precedent for how we log other maintenance changes.


I'm not sure what would be best for us. Okay, let's wait for Robin's thoughts.

robintw:
I think it's best to include entries in the Logs table that reference the deleted IDs (it is the Logs table that has the link to the ID, not the Changes table. We don't actually have a formal foreign key for that field, and I've checked and other parts of the code that deal with the Changes table (like the merging master/slave databases code) don't assume that the ids will necessarily match up - so it shouldn't break anything.

We could put the ID of the platform that it was merged into in the `new_value` field in Logs (although that's not quite what that field was designed to store), or could just record it in the description for the relevant Change entry. Either way, we should be able to see that these platform entries were deleted, and the relevant Sensors/States etc were moved over.

Actually, that raises another question: should we have an entry in Logs for each of the other objects that we have altered? We have `field` and `new_value` fields, so we could record a new value for the `Comment.platform_id` field for each comment that we change. It would be quite a lot of work, and lead to quite a lot of new Logs entries each time we do a platform merge, but it would give us a better 'paper trail' to work out what has been done (particularly if someone sometime needs to undo a merge...). What do you think about this @IanMayo?

IanMayo:
That's all good, thanks @robintw 
Yes - I think the verbose change is justified, to keep the audit trail.  The clients aren't expecting to do a lot of merging. The current duplicates need to be fixed, but better working practices should reduce the need for it in the longer term.

BarisSari:
Hello @robintw and @IanMayo. I added the necessary code to log the changed objects. As Robin suggested, I filled `new_values` field with the old value(merged platform_id s, or old sensor_id for States and Contacts). I also put a list of merged platform_ids in Change's reason. I hope we have now enough logs to trail the merge process.

robintw (changes requested):
Looks good, thanks.

Could you add a simple test to check that the Logs and Changes tables are updated correctly?

BarisSari:
Sure, I'm adding it

robintw (approved):
Great, thanks.

IanMayo (approved):

PR #757, opened 2021-02-09
Generic Merge

### LINKED ISSUE #754: 4.2 Merge (Generic)
## 🐞 Overview

Release one of Maintenance Tool just handled merging platforms.

Release two covers merging more types of data.  This task covers the backend processing necessary to do this.

## 🔗 Feature

#600 
## Merge processes
Merge is not required for all types of data.  This is how assorted types get merged:
* **Sensors** Move all measurements to "master"
* **Datafiles** Move all measurements to "master"

Note: for merging sensors, the sensors don't have to be for the same platform.  I don't think this affects the logic.

Once merged, the old data objects are deleted.

Beyond that, for all **Reference** tables, move linked objects to the new "master".

## Acceptance tests
* Read in some of our sample data, check we can merge sensors.  Check measurements have moved to the new sensor
* Read in some sample data, with platforms of different nations. Merge the nations to one. Check platforms updated.

## API
Receives:
* name of table
* list of IDs to merge
* ID for the new `master` entity

Processing:
* find tables that have this table as a foreign key
* find all instances for the IDs to merge
* change ID to the new `master` ID.

Returns:
* success/failure error code

BarisSari:
## 🧰 Issue
Fixes #754 

## 🔨Work carried out:
- [x] Allow merging Tasks and Tags
- [x] Allow merging reference table objects
- [x] Allow merging Sensors and Datafiles (by moving measurements)
- [x] Tests pass

I also fixed the Sphinx error, which was causing failure: https://github.com/sphinx-doc/sphinx/issues/8885 

## Confirmations

## 📝 Developer Notes:
Ian suggested writing a test that merges nations, but I thought it would be better to use the most used reference table: Privacies. That's why I tested merge methods with it.

While I was working to find out the referencing foreign key fields of a table, I found `sqlalchemy-utils`. And I realised that I'm trying to reinvent the wheel. My function was almost identical to one of its helper functions. 😄 So, I also included `sqlalchemy-utils` in this PR. It has many useful functions, especially foreign key helpers: https://sqlalchemy-utils.readthedocs.io/en/latest/foreign_key_helpers.html . 

You might ask why I didn't use `merge_references(from_obj, to_obj)` for all merge-operations. The reason is that we should handle dependent objects(Platform-Sensor, etc.) differently. If I tried to merge two platforms that have the same sensor name, it would throw a constraint error. So, I tried to use it as much as I can (for merging reference table objects), but I didn't use it for special cases.


IanMayo:
> So, I tried to use it as much as I can (for merging reference table objects), but I didn't use it for special cases.

Sounds perfect to me, @BarisSari  :-)

migrations/sqlite_versions/2021-02-10_a382a0f4b1be_correct_privacy_id_type_of_hostedby.py
```diff
@@ -0,0 +1,43 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "a382a0f4b1be"
+down_revision = "cb9a5f93f0aa"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
IanMayo, line 21: Forgive my being, dim, but. Should the above be "please do not adjust"?

BarisSari: No, Alembic says that we should check the autogenerated migration scripts. That's why it is written: "please adjust!". Let me remove it.

pepys_import/core/store/data_store.py
```diff
@@ -10,6 +10,7 @@
 from sqlalchemy.exc import ArgumentError, OperationalError
 from sqlalchemy.orm import sessionmaker, undefer
 from sqlalchemy.sql import func
+from sqlalchemy_utils import merge_references
```
IanMayo, line 13: Hey @BarisSari - `merge_references` is a great find.  That's doing a lot of the heavy lifting for this PR.

BarisSari: Yes, @IanMayo. It finds all dependent objects and updates them.

IanMayo:
I'm sure I made this comment before, @BarisSari  -  but now I can't find it.

I couldn't see this in the tests.  You're testing merging entries in the `Privacies` table. It would be a good idea to load some `State` entries from file, and check their privacy gets changed to the new `merged` value.  This will check that `merge_references` works in the way we're expecting it to.

BarisSari:
> I'm sure I made this comment before, @BarisSari - but now I can't find it.
> 
> I couldn't see this in the tests. You're testing merging entries in the `Privacies` table. It would be a good idea to load some `State` entries from file, and check their privacy gets changed to the new `merged` value. This will check that `merge_references` works in the way we're expecting it to.

Sure, @IanMayo. I'm extending the related test.

IanMayo (dismissed):
Great, thanks.  Will wait for @robintw  review.

robintw (changes requested):
Thanks for this @BarisSari.

The approach looks good, and I like the sqlalchemy-utils functions you're using.

I'm a little confused about a few bits though:
- Can `merge_generic` merge platforms? It doesn't look like it can at the moment - it'd be really good if all merges could just go through the `merge_generic` function, so we didn't have to do platforms separately. I can see why the `merge_platforms` function hasn't been deleted at the moment (as it would break the GUI), but hopefully we could delete it once I'd updated the GUI to use `merge_generic`.
- What about the other metadata tables that we might want to merge? I can see the code deals with Sensor and Datafile at the moment. What about Task, HostedBy, Participant etc. @IanMayo I assume I'm correct in saying these should be able to be merged too (I know we're not using them at the moment, but it'd be good if the code supported them for when we start using them shortly).

BarisSari:
Thanks for the review, @robintw. 

1. Yes, I didn't want to break the current flow. But, I'm going to update `merge_generic`. It makes sense to use it for merging Platforms as well.
2. I did it on purpose because Ian didn't mention those tables in the issue. According to his answer, I can update the method.

IanMayo:
@BarisSari - @robintw highlighted a couple of other tables that it would make sense to support merge:
* Tasks (in case two people add the same exercise, resulting in duplicates)
* Tags (for removing duplicates, or when granularity was too fine.  Analyst creates tags for `2020 Jan`, `2020 Feb`, ...  But then wishes to re-group into `2020 Q1`, `2020 Q2`, etc).

BarisSari:
> @BarisSari - @robintw highlighted a couple of other tables that it would make sense to support merge:
> 
> * Tasks (in case two people add the same exercise, resulting in duplicates)
> * Tags (for removing duplicates, or when granularity was too fine.  Analyst creates tags for `2020 Jan`, `2020 Feb`, ...  But then wishes to re-group into `2020 Q1`, `2020 Q2`, etc).

Okay, @IanMayo. I'm modifying the PR.

tests/test_data_store_merge_operations.py
```diff
@@ -215,7 +218,9 @@ def test_merge_platforms_with_different_sensor_names(self):
             assert len(states_before_merge) == 0
             assert len(contacts_before_merge) == 0
 
-            self.store.merge_platforms([platform_2.platform_id], platform.platform_id)
+            self.store.merge_platforms(
```
robintw, line 221: Most of the other calls to merge_platforms seem to have been switched to merge_generic, apart from this one and a couple of others.

robintw (changes requested):
Thanks @BarisSari, this looks good.

I see you've replaced most calls to `merge_platforms` with `merge_generic` - but there are a couple that have been missed in the tests. Could you switch those over too?

Also, given that we've now replaced all calls to `merge_platforms` - I think we should be ok to delete `merge_platforms` (I see you've replaced the call in the GUI too).

BarisSari:
Okay, @robintw. I'll update the tests. But actually, I'm still calling the `merge_platforms` inside of `merge_generic`: 
https://github.com/debrief/pepys-import/pull/757/files#diff-438b06426ef37127a7f6f793bc66ffc87ca4752f3f5f653d77fb1e9eb223e53bR1851

I just removed duplicated codes, and simplified the `merge_platform`. But I didn't delete it completely because `merge_generic` would be more complex.

robintw:
Ah yes of course @BarisSari - that makes sense. I'll approve now.

robintw (approved):
All looks good now, thanks for this Baris.

PR #758, opened 2021-02-11
Maintenance GUI - Phase 2

robintw:
## 🧰 Issue
#600 

## 🚀 Overview: 
Implements phase 2 of the maintenance GUI, as defined at https://docs.google.com/drawings/d/1WopPwQoUFGYnEZCQeglWsBalH7ye6ECWh2TyI0r7tfY/edit. This adds the ability to filter and merge other reference/metadata tables.

image

## 🤔 Reason: 
Ease maintenance of Pepys databases.

## 🔨Work carried out:

- [x] Add ability to select table
- [x] Generate relevant column data for FilterWidget automatically from selected table
- [x] Link to new `merge_generic` function to do merging on any table
- [x] Fix various other bugs and niggles in the interface (change colours on focus, add cancel button to merge dialog, fix various bugs)
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
The CodeCov integration is saying this PR is failing, because the coverage isn't good enough. It is very difficult to test the GUI fully from the tests, so we can't really get the coverage much higher. I have an approach for testing the GUI partially working (it at least checks that it opens and displays correctly), but can discuss spending more time on trying to get this fully working (probably in consultation with Jonathan) later, if you want me to. For the moment, I suggest just accepting the lower level of coverage.


IanMayo:
We can't logically merge Sensors from multiple Platforms.  

Hmm, maybe we can.  Maybe we merge Sensors within the same Platform.  Aah, but we'd have to ask the user which is the target Sensor for each platform.  That's cumbersome.

I think it's acceptable to popup error message when analyst tries to merge sensors from multiple platforms, indicating that it is only possible to merge Sensors from the same Platform.

IanMayo:
My performance may be suffering because I'm testing maintenance from a Windows VM.

It takes 6 seconds to enter Maintenance panel, then 5 seconds to switch to `Sensors`.  In the longer term it would be good to reduce this time.  But, if the first delay is spent doing the `load platforms table` processing, it would seem good to defer that step until after the maintenance panel is open, so we can display the `in progress` dialog.

pepys_admin/maintenance/gui.py
```diff
@@ -60,19 +57,17 @@ def __init__(self, data_store=None):
             # It won't cause problems for users, as they will access via Pepys Admin
             # where data store will be defined
             self.data_store = DataStore(
-                "", "", "", 0, ":memory:", db_type="sqlite", show_status=False, welcome_text=""
+                "", "", "", 0, "test_gui.db", db_type="sqlite", show_status=False, welcome_text=""
```
IanMayo, line 60: @robintw should we delete this file in the test `cleanup`

tests/gui_tests/test_gui_methods.py
```diff
@@ -18,22 +20,49 @@
 # for most of the tests without -s, and then the GUI tests with -s.
 
 
+def get_test_datastore():
+    if os.path.exists("test_gui.db"):
```
IanMayo, line 24: We use the "test_gui.db" string 4 times. It would be my preference to declare it once as a constant, then use that constant 4 times.

IanMayo:
> reduced codecov as %age of code

Yes, that's fine @robintw

pepys_admin/maintenance/gui.py
robintw: Yes - fixed in 19340c5d using a new pytest fixture (that is another way of doing a teardown, but doesn't require using a test class which would stop us using another feature we need there).

tests/gui_tests/test_gui_methods.py
robintw: Fixed in 19340c5d

IanMayo:
@robintw - previously we loaded the `Platforms` table as we entered the Maintenance view.  That made sense when `Merge Platforms` was the only action, but less so now we have more actions.

Anyway, loading `Platforms` was an opportunity to check the database tables were defined, and to show a suitable error message.

We don't do that now.  So, I got a long error thrown when I tried to merge `Nationalities` with an empty SQLite database.

image


This is unlikely to affect our actual users, but I think we should trap that instance.

IanMayo:
I'm playing with using `integer` values in the filters.

For `Privacies` we have the `level` field, which is an integer.   For numeric fields I think we should offer `<, <=, >, >=` operators.

Separate to that, if I select `Name` as the field of interest in `Nationalities` I am only offered `=` operator.

IanMayo:
One other observation, please

IanMayo:
Oops :-)

IanMayo:
Ok. May have something reproducible:
- enter maintenance mode
- select `Nationalities`
- filter for `Name`
- apply filter, check working
- filter for `Priority`
- no operator offered, cursor moves to top-left
- filter for `Name`
- operator for `Name` shown at top-left

screencast

robintw:
Thanks @IanMayo - lots of good stuff there that I'd missed.

I think they're all fixed now. I couldn't reproduce your reproducible example, which I assume means that I'd already fixed it with some of the other fixes I'd done - could you try it again from your end?

IanMayo (approved):
Working fine. No problems found.

BarisSari (approved):
It works without any problem. Thanks, @robintw.

PR #777, opened 2021-02-23
Maintenance GUI - Phase 4

robintw:
## 🧰 Issue
#600

## 🚀 Overview: 
Phase 4 of the Maintenance GUI - mainly adding edit capability.

**Screencast of edit functionality**
MaintenanceGUI_2021-02-26_b

**Screencast of view SQL functionality**
MaintenanceGUI_2021-02-26_a

## 🤔 Reason: 
Ease maintenance

## 🔨Work carried out:

- [x] Add `edit_items` method to `DataStore
- [x] Add EditDialog which provides GUI for editing items
- [x] Link up dialog and DataStore methods
- [x] Reinstate functionality for viewing SQL, with bug fixes and syntax highlighting
- [x] Tests pass

## Confirmations


robintw:
@IanMayo For this release (Phase 4) it says "View SQL" on the plan (https://docs.google.com/drawings/d/1WopPwQoUFGYnEZCQeglWsBalH7ye6ECWh2TyI0r7tfY/edit). Are you ok for it to be just viewing, and not editing the SQL? I can't remember what we decided when we discussed it before.

IanMayo:
Hello @robintw, view-only of SQL is fine for now. 

If you think view-only is more than a few hours of effort then I'm reasonably willing to let it slip to a future release.  But, my guess is it's quite straightforward.

robintw:
I think I've completed this now, but I'm aware there may still be some bugs or issues in the GUI that I haven't found. Please have a go at testing it and let me know if you find any problems.

IanMayo:
The screencasts look great @robintw 

As you'd predicted, `codecov` is a bit misleading, since it's hard to cover UI code in unit tests.

`codacy` has also complained a bit. I'm happy to leave it to your judgement about which warnings to respond to.

IanMayo:
@robintw  - is there a darker shade for the blue title in the edit form?
image

IanMayo:
I've entered a constraint error. I'd like to check it's correct.

I selected two sensors titled `GPS`. They're currently on different platforms.  I tried to edit them to associate them with a third platform.
image

I'll attach the SQLite file I was using.
[pepys_test.sqlite.zip](https://github.com/debrief/pepys-import/files/6051361/pepys_test.sqlite.zip)

I guess the uniqueness constraint is that we don't allow multiple sensors of the same name on one platform.  Is that correct?

IanMayo:
Last one. The `OK`/`Cancel` buttons look greyed out. They're not inactive. I think they should be in normal Black text.

robintw:
I think I've fixed all of those: I've got a new colour for the instruction text, black for the buttons, and have fixed another bug too:
image

As for the constraint error you had: yes, that's correct, the uniqueness criteria for Sensor is on `name` and `host` - so two sensors with the same name on the same host would give an error.

tests/test_data_store_edit.py
```diff
@@ -0,0 +1,133 @@
...
+
+        result = (
+            self.store.session.query(self.store.db_classes.Platform)
+            .filter(self.store.db_classes.Platform.identifier == "NEWIDENT")
+            .all()
+        )
+        assert len(result) == 1
+
+        assert result[0].name == self.platforms[0].name
+        assert result[0].nationality_name == self.platforms[0].nationality_name
```
BarisSari, line 38: This test and test below only show that name and nationality_name fields didn't change. I think it would be better to assert that the identifier was changed to "NEWIDENT".

pepys_import/core/store/data_store.py
```diff
@@ -2005,3 +2006,59 @@ def split_platform(self, platform_id, set_percentage=None) -> bool:
...
+        # Add a log entry for each field we've updated
+        # (We do all the updates in one SQL query above, for efficiency, but have to loop through the items
+        # and fields here to create the logs entries)
+        for current_id in ids:
+            for col_name, new_value in update_dict.items():
+                self.add_to_logs(
+                    table_object.__tablename__,
+                    row_id=current_id,
+                    field=col_name,
+                    new_value=str(new_value),
```
BarisSari, line 2062: I think we can log the `old_values` as we did for merge and split operations. Because these objects already have the `new_value` in their `col_name` field.

BarisSari (commented):
Looks great, @robintw. I just added 2 minor comments for testing and logging.

tests/test_data_store_edit.py
robintw: The change to NEWIDENT was implicitly tested by the fact that the query was filtering on the identifier, so wouldn't have returned any results if the identifier hadn't been updated properly - but I agree it's better to have it explicitly testing, so I've changed it.

robintw:
I've realised that the help text hasn't been updated since the first release of the GUI - under the actions heading it currently says:

```
# Fourth panel: Choose actions (F8)
Merge Platforms is the only available action for now.

The action applies to the platforms you selected in the preview
list if any. When merging, you will be prompted to select the
target platform in a popup. Select the platform and press Enter.
Once the platforms are merged, press Enter and the Preview List
will be updated.
```

@IanMayo or @clenoble - would you be able to draft some updated text that also explains splitting and editing?

IanMayo:
@robintw & @clenoble - my attempt at the new words.

```
# Fourth panel: Choose actions (F8)

1 - Merge.  When multiple rows from a table are selected, the user is invited to merge the e rows.  The merge action
starts with the user selecting the target row that that the data is to be merged into.  When merging platforms, the
sensors of the platforms to be merged are moved to the target platform.  If a sensor with that name already exists
then the measurements are moved to the existing sensor.
2 - Split platforms.  It the data-type is Platform, and just one row is selected, the user is able to split that platform
into multiple new platform instances.  The platforms are split according to the datafiles their measurements are
imported from.
3 - Edit values. If one or more rows are selected then Edit Values can be used to edit one  or more fields for those
data items.
```

@robintw  - I think I've introduced some more "selective logic" in there, where the Actions are only enabled for relevant circumstances.  I suggest we grey out Actions that aren't available, rather than add/remove them.

robintw:
@IanMayo Hmm, at the moment, if you try and run an action that can't be run in the current circumstances (eg. more than one item is selected, and you try and run Split Platforms) then you just get a dialog saying it can't be run. I will look into trying to grey out entries that aren't applicable, but it'll take a bit of work. How would you like moving up and down the list to work with greyed out items? Do you want them to be skipped when you use the up and down arrows? I'm trying to work out how to implement that at the moment: it should be possible, but I'll need to do a bit of playing around with it.

It's up to you how high a priority you put on this work: I'm happy for this to be merged and released without it, and the greying out to come in a later release.

IanMayo:
Aah yes, it could be quite a bit of work.  Let me check my wording.  It seems ok for the current behaviour. 

**Note:** in my suggested words I've deleted a stray standalone `e` in the first sentence of `merge`.

IanMayo:
Second **note:** - this PR is still marked as a Draft PR.

The analysts are working on Pepys tomorrow - it would be good if we could get a new distribution out today please @robintw

robintw:
Sorry, I must have forgotten to mark it as ready for review. As far as I'm concerned, this is ready for merging now - I've just fixed the help text to remove the extra `e`.

I'm working on allowing actions to be disabled - but that can wait for a future release.

IanMayo:
@robintw -  I was just creating an issue for the "context-sensitive actions", and suspect the current arrangement is the best.  If we disable actions then we don't have a way of informing the user of the circumstances when they are enabled.  In the current processing (popup dialog) we can provide guidance on what must be selected to to use that action.

That _seems_ better for users.  Let's see what @clenoble thinks.

robintw:
Ok, sure.

FWIW, I've implemented a version where actions that aren't relevant are greyed out, but can still be activated (by pressing enter or clicking), which shows the dialog explaining why they aren't able to be used. I was just about to create an issue myself to say that actually stopping the user moving to the actions that are disabled would be difficult, as we'd have nowhere to put the cursor if all actions were disabled (eg. when the user loads the app and hasn't even selected a table yet).

So, if you wanted (and @clenoble agreed), we could grey out the unavailable actions, but still leave them clickable so users can find out what they need to do to enable them? I'm not sure whether that would be helpful (a visual clue as to which actions are available) or confusing (things that are greyed out are still clickable). We could even add a brief message after the action name (we've got lots of horizontal space) saying something like "unavailable, click for more information" or something.

IanMayo:
Yes - that sounds like the best of both worlds.  "Click for more information" is a great solution.  Let's go for that.  If it's not too much effort I'm happy to include it in this PR.

robintw:
I'd prefer to test it a bit more thoroughly, and not rush to include it in this PR - as it involves a fair amount of changes to the combo box that is used for the actions list. I'll implement it, and we can get @clenoble feedback too.

IanMayo:
That's fine @robintw  - no rush.

IanMayo (approved):
Have tested editing a range of data types.

robintw:
All tests are now passing on this @IanMayo. There are a few Codacy issues that I can't seem to fix (it complains about docstrings when they start with a newline, and when they don't start with a newline!) - but I think it's ready to be merged, and then we can get a release out to the users today.

PR #784, opened 2021-03-01
Disable action entries in GUI when not applicable 

robintw:
## 🧰 Issue
#600  #785

## 🚀 Overview: 
This PR changes the list of actions on the bottom left of the display, to mark actions that are not currently available to be run. For example, merging entries requires multiple entries to be selected, so the merge action will be unavailable until multiple entries are selected. Similarly, the Split Platforms entry requires that a single entry from the Platforms table is selected, and so on.

Currently, the unavailable entries are greyed out, but are still able to be activated (by pressing Enter or clicking) - so that the user can still see the message box telling them why the entry is disabled. We also add a bit of text to the end of the entry telling the user to click to get more information (they can also press Enter, but I can't think of a nice way of phrasing that in the message - any ideas?)

I'd be interested in your feedback on this @clenoble, to see what you think from a UX perspective.

MaintenanceGUI_2021-03-01
You can see in the screenshot above that all the actions start off unavailable, and then gradually become available as the conditions match the requirements.

## 🤔 Reason: 
Make things clearer for the user.

## 🔨Work carried out:

- [x] Add new AdvancedComboBox widget that has this functionality (as it would complicate the standard ComboBox widget to have this functionality always available)
- [x] Use this AdvancedComboBox for the actions list, plus little functions to tell when an entry should be enabled or not
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
I'm a little concerned about the phrasing of the bit we add to the entry text - I'm not sure whether 'unavailable' is right, and whether we ought to mention that they can press Enter on it too? I'm also a little concerned that they all start off unavailable, and so users might think they've been disabled by an administrator or something. Any thoughts @IanMayo @clenoble?


IanMayo (dismissed):
Tested deployment in MS-Windows. Working perfectly.

IanMayo:
The Java software engineer in me would have taken a more heavyweight route, creating a list of actions:
- Actions class:
    - name(): string
    - isEnabled(table_type, selections): boolean
    - notEnabledMessage(): string
    - execute(table_type, selections): void
  
Each action would then extend/implement that type.

But, that may not be pythonic.  Maybe it would be of greater value as the list of actions grows.  With a list in place the UI code will largely just be cycling through them.

robintw:
Hmm yes, that is a possible way of doing it. I was aiming to keep it as simple as possible to begin with - and keep it separated from the UI as much as possible, with the UI being composed of fairly generic objects - in this case a ComboBox that can have disabled items. At the moment the UI doesn't know anything about what the entries are - it just has a list of entry strings and booleans for whether they're enabled or disabled. To be honest, I think the current way of doing it fits better with the architecture of the rest of the GUI, but I'd be happy to look into changing it if you'd prefer.

The issue with the other way of doing it would be that the `execute` method needs to access various other methods/attributes on the main MaintenanceGUI class - and if these actions were all encapsulated in separate classes then that would be harder to do (I guess these classes could have a reference to the main MaintenanceGUI instance, but that kinda breaks the encapsulation of having them in a separate class).

IanMayo:
_Maybe_ there is some merit in this `codachy` suggestion:
image

robintw:
Ah yes, that's a good idea: it can actually be done with a list comprehension, which is even better. Fixed now.

IanMayo:
> it can actually be done with a list comprehension

_One day_ I'll have the Python skills to understand what's happening here: 🤣 

image

IanMayo (approved):

clenoble:
@robintw @IanMayo Sorry I was so long to answer on this. I have been very busy. I think your solution works perfectly. 
I would only change the accompanying text by (unavailable - select for more information). "Select" can be clicking or pressing Enter.

PR #786, opened 2021-03-01
Maintenance GUI - Phase 5

### LINKED ISSUE #790: 'Show info' action in GUI?
I had an idea just now...would it be worth having a 'Show info', or 'Show full details' action in the GUI? It would allow you to select a single item in the preview table, and then show all of the fields for that item. It would save the user having to add and remove fields from the display manually until they've seen all the fields they want for that item.

In terms of implementation, it would be very easy: it would just be the left-hand side of the Edit Dialog, displayed by itself in a dialog.

(Yes, as mentioned above, the users could do this already by choosing to edit and then cancelling the dialog, but that's not obvious)

Thoughts @IanMayo?

robintw:
## 🧰 Issue
#600 
Fixes #790

## 🚀 Overview: 
This PR covers half of the Maintenance GUI - Phase 5. It adds 'Add', 'Delete' and 'View' actions, but does _not_ add filtering, displaying and editing of measurement tables. It also adds the ability to run actions by simply typing their number (1, 2, 3 etc) from either the actions pane or the preview table pane. That makes it easier to run an action: simply select the relevant items, and then press `4` (or whatever).

**Add action**
MaintenanceGUI_2021-03-02

**Delete action**
MaintenanceGUI_2021-03-03

## 🤔 Reason: 
Ease of user maintenance

## 🔨Work carried out:

- [x] Add multiple actions to GUI, with relevant dialogs and links to DataStore methods
- [x] Add keybindings to process single digit keys to run actions, and link CheckboxTable to be able to run these actions too
- [x] Tests pass

## Confirmations


robintw:
@IanMayo @clenoble - Here is a video showing the new Add functionality I've implemented. Note that required fields are shown with a `*` and you get an error if you don't fill in a required field. Feedback welcome.

MaintenanceGUI_2021-03-02

IanMayo:
Screencast is great, thanks @robintw .

robintw:
We're making good progress with the add and delete functionality, but I've got a number of questions about starting to include measurement tables (`State`, `Contact` etc) in the GUI. I've got the very basics working (you can view the States table, for example), but there are lots of things we need to decide about:

**Filtering measurement tables**
What sort of fields do we need to be able to filter by? For example, do we need to be able to:
- filter by location? If so, we'll need some way of specifying something like a bounding box, or a point and a radius or something - and we're going to start getting into complicated GUIs for specifying these, plus more complicated geospatial queries for querying the data. Definitely not impossible, but quite a lot of work - so we need to consider whether the user needs it.
- filter by measurement values - such as speed, bearing, elevation and so on. This would definitely be easier, but we'd need to work out what we were doing about units: do we force the user to query in the units the database uses internally (m/s, radians etc), or do we allow them to specify their units and do the conversions ourselves. Again, that requires the relevant GUI and backend code to deal with it.

It makes sense that we would want to be able to filter by the Platform or Sensor that the measurement was obtained from - but this actually gets a bit difficult. There can (and probably will) be multiple Platforms/Sensors etc with the same name - so just having a dropdown of name won't be very useful (if there are 5 items with the same name, which do you choose). What we really need to do is to provide a dropdown with all the distinguishing criteria used for that table in it - so a dropdown showing something like `HMS BLAH - 123 - United Kingdom` (name, identifier and nationality). _However_, if we do that then we're creating our own representation of data from the database and can't easily do things like `LIKE` queries on it (and `LIKE` queries would be weird anyway, as they would match in the nationality name as well).

The same problem appears when we are editing measurement tables: we need a way of choosing a new Sensor for a State object, or assigning one from scratch if we're adding rather than editing. In fact, this will get even more complex for some measurement tables, as we want to be able to search on the Platform that is linked to a State, but we can't let the user edit the Platform value directly, as it is indirectly referenced through the assigned Sensor (a State has a Sensor, and each Sensor has a Platform).

Actually, these issues also already apply to the Sensors table, which even though it is a metadata table, has a foreign key to the Platforms table, so suffers from the same problems.

What do you think @IanMayo? Do we have some use cases from the clients on what they actually want to do with the GUI when it comes to filtering and editing measurement tables? If not, I think these would be helpful, as it would save us spending a lot of effort implementing functionality that may not be used (although I can understand the desire to have all of the Add/Edit/etc functions work for measurement tables, for completeness sake).

robintw:
@IanMayo Once Baris's PR (#789) has been merged, the add and delete functionality will be complete in this PR. Do you think there is value in me removing the (very small) extra code that supports measurement tables, fixing the tests, and then merging this to get Add/Delete for metadata/reference tables released to the client soon?

We can then take our time working out the answers to the questions I posted yesterday evening (previous comment above this one) about how to deal with measurement tables.

IanMayo:
@robintw - sure, good idea.  Let's get the release out.

IanMayo:
> filter by location

Yes, let's defer this.  

> filter by measurement values

Yes, let's include this.  It could let them do some elementary analysis. I think we can be more restrictive in units, at first.  We could only take values in the existing units for an attribute.  We could defer being flexible with units.

Yes, there are some interesting issues there.  I think we can't avoid the multi-field drop-down entries. I'd hope that the type-to-filter practice that we use in `import data` would also work within the form.  I think in most cases the user will type the platform name, and if multiple nations have a platform with that name then the user will select the relevant one.

Maybe the sensor/platform issue is simpler. If we are editing a State object, then the Sensor field will contain values that represent the Sensor _and_ the Platform (sadly we would need the 3 platform unique identifiers).  Again, I would expect the user to trim the matches down using the platform name first, then click/select the relevant sensor from the reduced list.

** Filtering/editing measurements.** 
Once we have a graphing capability I look forward to them producing their own tim-variable graphs.  Editing measurements should be rare.  I think the "change" sensor processing may actually be a "merge" operation, though that would only transfer _all_ measurements from one to another.  Aah, another (future) operation could be tagging measurements.  For some type of analysis they may want to know periods when a platform is travelling within a set speed bracket. They'll use maintenance tools to find the data, then add a tag to the State rows (well, create TaggedItem entries).

For the bigger picture regarding measurements, I don't think they're thinking at that level yet.  They're focussing on the metadata, and just using `Debrief` to handle (view) measurements.  But, as they move away from spatial analysis, they will need new tools/capabilities.

IanMayo:
Thanks @robintw - aim to test it tomorrow morning, unless you really want feedback tonight?

robintw:
No @IanMayo - tomorrow is fine, I'm finishing for the day now anyway.

BarisSari (dismissed):
Looks good, thanks @robintw.

IanMayo (changes requested):
@Robin when the focus is on the checkbox table, and I press (for example) 5, then the digit 5 appears in the data source type checkbox.

screencast

robintw:
@IanMayo Did you press `F6` to move focus to the Preview List first? Thinking about it, I expect you just clicked on the preview table?

At the moment, I don't think clicking on the preview list actually moves focus there - clicks kinda operate outside of the focus model in prompt_toolkit. I can see a way that I can probably fix this though - I'll investigate shortly.

IanMayo:
Aah yes. If I tab to the Preview pane it keeps focus.  It's just when I mouse-click a row that it doesn't take focus.

IanMayo:
One other issue, please.  I've just looked in the `gui.log` file. It shows me the detail of the two rows I've just added:
image

The presence of data values in the log means our users can't send the log directly to us, if we do have some future Maintenance UI problem.  Would it be possible to switch the logging to give more detail about what's happening, rather than the values being entered?

robintw:
Those should all be fixed now. The logging information is removed, the focus is now set to the preview table when you click on it, and the text showing that an action is disabled has now been updated to Celine's new wording.

IanMayo:
This is really minor...

I'm editing a row in a table with only one field.  The "^" icon looks a bit like it's for a combo box control.

But. I now realise it's a one-line-high scrollbar control.

Is it possible to not show the scrollbar if there is only one row?

In fact, since the dialog is sized proportionate to the table, I haven't encountered an Edit Form where the scroll bars do anything.

Maybe we only show them when the logical panel is taller than the visible viewport?

IanMayo:
Oh, and would it be possible for `Escape` to cancel a view/add/edit form?

robintw:
Escape does cancel any of the dialogs - at least it does for me. Can you check again, and let me know exactly what you're pressing/clicking if you find it doesn't work?

I'll look into what we can do about scroll bars. They come automatically with the Scrollable Pane control that I am using, but you're right, I've yet to find a situation where they are actually needed. Probably the simplest thing to do is remove the scrollability entirely - as it isn't really needed. I don't think there is an easy way to turn the scrollbars in a ScrollablePane on and off, but we could definitely not bother with the scrollable pane if there is only one entry. In general thought, it would probably look tidier if we just removed the ScrollablePane alltogether.

IanMayo:
More investigation of the above:
* I have add or edit dialogs open
* I press Esc
* After a delay of a couple of seconds the dialog goes
* If I click on `Cancel` the dialog closes immediately

The time difference between the two close operations suggests some error may be being thrown in the background.

robintw:
Ah yes: there was a bug in the View dialog which caused it to crash when you press Escape. That is fixed now. I've also removed the ScrollablePane from the dialogs, as it isn't used at the moment - if we end up with a situation where we need it then we can add it back in future.

As for Escape taking a while to trigger: yes, I've noticed that myself. It seems worse on Windows, but still present on other platforms. I've no idea what's causing it - there is no sign of any error in the background. Shall I add it to my list of things to discuss with Jonathan?

IanMayo:
> bug in view dialog

Great, fixed.

> Delay in cancelling

Yes, putting it on the list for Jonathan would be good.

IanMayo (approved):
Great, well done.

BarisSari (approved):

PR #789, opened 2021-03-02
Maintenance - Delete Records

### LINKED ISSUE #788: Method to delete records from a table
## 🐞 Overview
As part of the Maintenance GUI, we need a 'delete' action that will delete the selected records from a specific table. For example, we might have selected multiple Platforms and want to delete them. Actually doing the delete is easy - just run a SQLAlchemy delete operation - but we need to work out what to do about the items that depend on them.

For example, when deleting a Platform, we may have Sensor, Comment, Participation etc objects that link to that Platform. What should we do about those objects? Similarly, when deleting a Nationality or PlatformType object we may have Platforms that depend on that object - and so on, for all the other tables.

I'm genuinely not sure what to do about this - I don't know how users would expect to deal with this situation, and how much they would understand the situation itself. @IanMayo @clenoble Your advice here would be great.

Once we've established how this should work, I would suggest @BarisSari works on the method, while I work on some of the other GUI stuff.

## 🔗 Feature
#600, specifically supporting #786.

## 🔢 Acceptance criteria
A new method on DataStore that deletes items given a list of IDs (or item objects)
Tests showing that the delete method works correctly given various database objects with various dependencies between them

BarisSari:
## 🧰 Issue
Fixes #788 

## 🔨Work carried out:
- [x] Tidy up maintenance methods. Use this order: Helper internal methods, Other helper methods, and Main maintenance methods, which the GUI calls
- [x] Add ondelete="cascade" to all foreign key relationships
- [x] Add migration scripts  
- [x] Write tests for `find_dependent_objects` method
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
I assumed we already support cascading delete. Apparently, we had this support for only a few foreign key relationships. That's why I added `ondelete="cascade"` parameter to all foreign key relationships.


pepys_import/core/store/data_store.py
```diff
@@ -2066,3 +2070,23 @@ def edit_items(self, items, edit_dict):
                     new_value=str(getattr(item, col_name)),
                     change_id=change_id,
                 )
+
```
IanMayo, line 2073: @BarisSari - would you mind documenting the parameters for this method please? In particular, is `id_list` the ids of the items that are to be deleted?

I see it's not being called, but I guess that's because it's going to be called from Robin's UI code.  Aaah, since it's to be used elsewhere, it's probably worth documenting (commenting) the structure that gets returned.

I was going to suggest that you cover the new method with tests, but now realise this PR is still in Draft.

I probably shouldn't even be adding these comments/observations yet.

Oops 🤣

BarisSari: Hello @IanMayo . Yes, that's the list of items that are to be deleted.

I'll document it and add tests. I just thought maybe I could wait until we get an answer from the client. That's why I only implemented a method that returns the dependent objects now. So, I'll add tests, don't worry. 😄

robintw:
@BarisSari Thinking about this from the UI perspective, we want to run the delete in two ways: first we want to be able to run a function and find out how many dependent items would be deleted, and then we want to run a function to actually do the delete. It might be easiest to implement them as a single function that takes arguments to say what to do (eg. `operation='count'` and `operation='delete'` or something, or even just `confirm=True` on the second call). Does that make sense?

BarisSari:
> @BarisSari Thinking about this from the UI perspective, we want to run the delete in two ways: first we want to be able to run a function and find out how many dependent items would be deleted, and then we want to run a function to actually do the delete. It might be easiest to implement them as a single function that takes arguments to say what to do (eg. `operation='count'` and `operation='delete'` or something, or even just `confirm=True` on the second call). Does that make sense?

Hello @robintw. I actually created two methods for the same purposes:  `find_dependent_objects` for finding out how many dependent items would be deleted and `delete_objects` (it was already there, I modified a bit) for deleting.  We can create one method with an additional parameter, but we already need a delete method for the merge tasks. So, there will be duplicated codes. So, I think having two different methods is better. What do you think?

I have realised that the latest methods of DataStore become more and more complex to track. If you agree, I can use this order to tidy them up: 

1. Helper internal methods
2. Other helper methods
3. Main maintenance methods, which the GUI calls

What do you think?

robintw:
Yeah, two methods is fine - it probably makes more sense.

As for tidying up - yes feel free to do a bit of tidying and re-arranging.

IanMayo (approved):
Tests look fine.

tests/test_data_store_delete_operations.py
```diff
@@ -0,0 +1,213 @@
...
+        with self.store.session_scope():
+            result = self.store.find_dependent_objects(constants.PRIVACY, [self.privacy_id])
+            assert "Privacies" in result and result["Privacies"] == 1  # self.privacy
+            assert "Datafiles" in result and result["Datafiles"] == 2  # self.file and self.file_2
+            assert "Platforms" in result and result["Platforms"] == 1  # self.platform
+            assert "Sensors" in result and result["Sensors"] == 1  # self.sensor
+            assert "States" in result and result["States"] == 2  # self.state and self.state_2
+
+            self.store.delete_objects(constants.PRIVACY, [self.privacy_id])
+            result = self.store.find_dependent_objects(constants.PRIVACY, [self.privacy_id])
```
robintw, line 172: Could you add a simple check here to ensure that the privacy object has actually been deleted, by querying for a privacy entry with that ID. I know it should appear in `find_dependent_objects` if it hasn't been deleted, but checking it explicitly would mean we wouldn't be caught out if there was a bug in `find_dependent_objects`.

pepys_import/core/store/data_store.py
```diff
@@ -2063,3 +2058,57 @@ def edit_items(self, items, edit_dict):
                     previous_value=str(getattr(item, col_name)),
                     change_id=change_id,
                 )
+
+    def find_dependent_objects(self, table_name: str, id_list: list) -> dict:
```
robintw, line 2062: This method takes a table_name parameter as a string, but the `delete_objects` method takes a table object instead. Could you make them match? Ideally, by making both functions take either table names or objects using the:

```
if isinstance(table_obj, str):
    table_obj = self._get_table_object(table_obj)
```

robintw (changes requested):
Looks good, just a few minor niggles.

pepys_import/core/store/data_store.py
BarisSari: I changed table_obj to table_name: https://github.com/debrief/pepys-import/pull/789/commits/c766544a7abc49a54b98e8d8a355684d8ae850d1

tests/test_data_store_delete_operations.py
BarisSari: Added: https://github.com/debrief/pepys-import/commit/c766544a7abc49a54b98e8d8a355684d8ae850d1

pepys_import/core/store/data_store.py
```diff
@@ -2096,17 +2096,16 @@ def find_dependent_objects(self, table_name: str, id_list: list) -> dict:
...
-        :param table_obj: A table object, or name of the table that IDs belong to
-        :type table_obj: SQLAlchemy Model or str
+        :param table_name: Name of the table that IDs belong to
+        :type table_name: str
         :param id_list: List of objects IDs
         :type id_list: list
         """
-        if isinstance(table_obj, str):
-            table_obj = self._get_table_object(table_obj)
+        table_obj = self._get_table_object(table_name)
```
robintw, line 2108: I think I might not have explained myself properly: I was hoping that both `delete_objects` and `find_dependent_objects` could take a parameter which could be _either_ a table name, or a table object. This change actually removes this from `delete_objects` - I was hoping the logic that was in `delete_objects` could be added to `find_dependent_objects` as well. Does that make sense?

robintw (changes requested):

pepys_import/core/store/data_store.py
BarisSari: Aah, my mistake. Yes, it makes sense. I changed the code: https://github.com/debrief/pepys-import/pull/789/commits/d8544fcfa31ff15ba25727b96d39f7f0d0c8129f

robintw: Great, thanks - I'll approve now.

robintw (approved):
All good now, thanks.

IanMayo:
@BarisSari  - the alembic code for cascading deletes is huge. Did you have to hand-craft all of that?

BarisSari:
> @BarisSari - the alembic code for cascading deletes is huge. Did you have to hand-craft all of that?

No, @IanMayo. Most of it was generated by Alembic's itself. I added `naming_convention` parameter to SQLite code, and that's it. There were some old foreign keys which don't have a name. That's why I needed to add this parameter. Most of the time, I don't touch the generated code :)

IanMayo:
(just waiting for tests to complete)

PR #793, opened 2021-03-04
Update to include reminder to use alembic revision

IanMayo:
Also provide link to guidance

Minor documentation change, to ensure we lead devs to alembic guidance.


robintw (dismissed):
Great

BarisSari (approved):

BarisSari:
I approved the PR but this might also help, @IanMayo: https://pepys-import.readthedocs.io/en/latest/database_migration.html#how-to-use-it-for-developers
Do you think it's worth to add this link as well?

IanMayo:
> I approved the PR but this might also help, @IanMayo: https://pepys-import.readthedocs.io/en/latest/database_migration.html#how-to-use-it-for-developers
> Do you think it's worth to add this link as well?

Hmm. Maybe we should move the wiki content into the `readthedocs`.

What do you think?

BarisSari:
> > I approved the PR but this might also help, @IanMayo: https://pepys-import.readthedocs.io/en/latest/database_migration.html#how-to-use-it-for-developers
> > Do you think it's worth to add this link as well?
> 
> Hmm. Maybe we should move the wiki content into the `readthedocs`.
> 
> What do you think?

I'm not sure. Because the wiki content is short and simple. But you might easily get lost in `readthedocs`. It's up to you 🙂

IanMayo:
Let's move it to read the docs, please

BarisSari:
I moved it, @IanMayo: https://pepys-import--793.org.readthedocs.build/en/793/database_migration.html#how-to-use-it-for-developers
 
I also updated the link in the PR template.

IanMayo:
image

This paragraph is a bit confusing.  What is it that we only do for Postgres?  We have transitions for both types of database.

I'm unsure what the use of `--sql` is.  It says we need to use it, then we can't use it.

BarisSari:
> image
> 
> This paragraph is a bit confusing. What is it that we only do for Postgres? We have transitions for both types of database.
> 
> I'm unsure what the use of `--sql` is. It says we need to use it, then we can't use it.

I put a horizontal line to separate this from the latest revision ID section. But it seems like it's not clear yet. 

The purpose of this paragraph is to offer another checking option to developers. And it says that it's possible to create raw SQL scripts from a migration file. But it only works for PostgreSQL. If you try to generate raw SQL from SQLite migration script, it fails.

PR #794, opened 2021-03-04
Maintenance GUI - Phase 5, Part 2

### LINKED ISSUE #791: Issues around measurement table filtering and editing
_Note: The text below is copied from comments on #786, the original PR for doing these changes. That PR was re-scoped to just Add/View/Delete functionality and not dealing with measurement tables, so I've copied the comments here so they're easier to access once that PR is merged._

**Robin:**
We're making good progress with the add and delete functionality, but I've got a number of questions about starting to include measurement tables (State, Contact etc) in the GUI. I've got the very basics working (you can view the States table, for example), but there are lots of things we need to decide about:

Filtering measurement tables
What sort of fields do we need to be able to filter by? For example, do we need to be able to:

filter by location? If so, we'll need some way of specifying something like a bounding box, or a point and a radius or something - and we're going to start getting into complicated GUIs for specifying these, plus more complicated geospatial queries for querying the data. Definitely not impossible, but quite a lot of work - so we need to consider whether the user needs it.
filter by measurement values - such as speed, bearing, elevation and so on. This would definitely be easier, but we'd need to work out what we were doing about units: do we force the user to query in the units the database uses internally (m/s, radians etc), or do we allow them to specify their units and do the conversions ourselves. Again, that requires the relevant GUI and backend code to deal with it.
It makes sense that we would want to be able to filter by the Platform or Sensor that the measurement was obtained from - but this actually gets a bit difficult. There can (and probably will) be multiple Platforms/Sensors etc with the same name - so just having a dropdown of name won't be very useful (if there are 5 items with the same name, which do you choose). What we really need to do is to provide a dropdown with all the distinguishing criteria used for that table in it - so a dropdown showing something like HMS BLAH - 123 - United Kingdom (name, identifier and nationality). However, if we do that then we're creating our own representation of data from the database and can't easily do things like LIKE queries on it (and LIKE queries would be weird anyway, as they would match in the nationality name as well).

The same problem appears when we are editing measurement tables: we need a way of choosing a new Sensor for a State object, or assigning one from scratch if we're adding rather than editing. In fact, this will get even more complex for some measurement tables, as we want to be able to search on the Platform that is linked to a State, but we can't let the user edit the Platform value directly, as it is indirectly referenced through the assigned Sensor (a State has a Sensor, and each Sensor has a Platform).

Actually, these issues also already apply to the Sensors table, which even though it is a metadata table, has a foreign key to the Platforms table, so suffers from the same problems.

What do you think @IanMayo? Do we have some use cases from the clients on what they actually want to do with the GUI when it comes to filtering and editing measurement tables? If not, I think these would be helpful, as it would save us spending a lot of effort implementing functionality that may not be used (although I can understand the desire to have all of the Add/Edit/etc functions work for measurement tables, for completeness sake).

**Ian:**
filter by location

Yes, let's defer this.

filter by measurement values

Yes, let's defer this until we're clear on it.

I do see the value - It could let them do some elementary analysis. I think we can be more restrictive in units, at first. We could only take values in the existing units for an attribute. We could defer being flexible with units.

Yes, there are some interesting issues there. I think we can't avoid the multi-field drop-down entries. I'd hope that the type-to-filter practice that we use in import data would also work within the form. I think in most cases the user will type the platform name, and if multiple nations have a platform with that name then the user will select the relevant one.

Maybe the sensor/platform issue is simpler. If we are editing a State object, then the Sensor field will contain values that represent the Sensor and the Platform (sadly we would need the 3 platform unique identifiers). Again, I would expect the user to trim the matches down using the platform name first, then click/select the relevant sensor from the reduced list.

** Filtering/editing measurements.**
Once we have a graphing capability I look forward to them producing their own tim-variable graphs. Editing measurements should be rare. I think the "change" sensor processing may actually be a "merge" operation, though that would only transfer all measurements from one to another. Aah, another (future) operation could be tagging measurements. For some type of analysis they may want to know periods when a platform is travelling within a set speed bracket. They'll use maintenance tools to find the data, then add a tag to the State rows (well, create TaggedItem entries).

For the bigger picture regarding measurements, I don't think they're thinking at that level yet. They're focussing on the metadata, and just using Debrief to handle (view) measurements. But, as they move away from spatial analysis, they will need new tools/capabilities.

robintw:
## 🧰 Issue
#600
Fixes #791 

## 🚀 Overview: 
This is the second part of Phase 5 of the Maintenance GUI, adding the ability to filter/view/edit measurement tables, alongside a few other things.
**Note:** This PR does not include the ability to filter on location or floating point measurement columns (such as speed or heading). Editing location is also not supported.

**Screencast of filtering and editing State entries**
MaintenanceGUI_2021-03-05

**Screenshot of initial help dialog**
This only appears the first time
image

## 🤔 Reason: 
Simplify user database maintenance

## 🔨Work carried out:
- [x] The ability to select measurement tables and filter them (note: filtering on location or floating point measurement fields is not supported yet)
- [x] The ability to edit, view, merge entries in measurement tables
- [x] Improvements to selecting complex foreign keyed objects when editing/adding - for example, when selecting a Platform for a new Sensor entry, the dropdown contains all the distinguishing information for the Platform
- [x] An initial help ('getting started') dialog that appears on first load. After it has appeared, it writes a file in the home directory (`~/.pepys_maintenance_help.txt`) which is used as a marker to tell the app not to show the dialog next time.
- [x] Fixed bug with MaskedInputWidget where a backspace from the first field would cause a crash
- [x] Improved display of EditDialog by increasing size and evening out the columns
- [x] Tests pass

## Confirmations


robintw:
@IanMayo How would you like complex foreign key relationships to be shown in the dropdowns. For example, at the moment, when editing a Sensor we have a list of entries that look like:

`SensorName / PlatformName / PlatformIdentifier / PlatformNationality`

So, something like:

`GPS / ADRI / 123 / United Kingdom`

Does that look about right to you? The same thing for the Platform dropdown would be the same but without the first element (just `ADRI / ...`). When the user types to filter it will filter on any of the text in that dropdown.

IanMayo:
> Does that look about right to you?

Yes, that looks right.

robintw:
@IanMayo @BarisSari I've marked this as Ready for Review and invited you to review. I think all the functionality is there, and I've done quite a bit of testing of the GUI, but it would be good for you to both have a go at trying to break it. I've changed how a number of things work, and I hope I haven't broken anything and not noticed!

IanMayo:
In my MS-Windows VM the aqua title is too light to be read clearly:
image

Users are no longer restricted to `Platforms` table:
image

I'm looking at `State` filters.  I think we need the platform (name, id, nationality) - to distinguish which sensor we're looking at, like in the `Edit Values` form:
image

robintw:
@IanMayo I'll fix the aqua text and the contents of the help dialog.

The sensor dropdown in the FilterWidget is a bit more complicated. So far, I've deliberately not switched to using 'complex' dropdowns (like `NAME / IDENTIFIER / NATIONALITY`) within the FilterWidget, as in some ways it causes as many problems as it solves - and I'm not sure of the way forward.

If we switch to using these complex dropdowns then we can't do `LIKE` queries based on them - as the `NAME / IDENTIFIER / NATIONALITY` string (or equivalent) isn't stored in the database - it's just generated by the Python code. So for these foreign keys, we would have to be limited to `=` or `!=` - which would work based on the ID value of the entry that was chosen in the dropdown. Would that be acceptable?

It would also limit our ability to search for specific sorts of things when they relate to foreign keys. At the moment, selecting the `sensor` field when querying States actually queries the `sensor_name` field, which could be used to find all States measured by a specific sensor name, regardless of the platform. I could see this potentially being useful (eg. selecting all measurements from a particular type of sensor, or make/brand or similar). Though actually, that would be more straightforward if we provided filtering by a `sensor type` field.

I can see two ways forward:

1. We switch to using these 'complex dropdowns' for all foreign keys, and therefore don't allow LIKE queries for these foreign keys.

2. We _add_ these complex dropdowns with the filter column called something like `sensor entry` (or possibly just `sensor`), and then add the other foreign keyed fields as `sensor name`, `sensor type name` etc. That would allow filtering on the whole entry, or on any constituent part of it. This would be the most flexible, but would probably come with a time penalty for gathering the required data when selecting a table (although a similar amount of time will be saved when using the edit/add functionality, as we wouldn't have to load the data again there), and could add significantly to the number of columns available to be selected by the user (although we could restrict to just the ones we think are interesting).

One issue which we would have is that some foreign keys are 'simple' (eg. nationality or platform type - they just have a single name) and we wouldn't want to limit the use of operations such as LIKE with these entries. I'm not sure how best to handle this - we could potentially check each association proxy to find out if it's foreign table is a reference table or not, and if it's a reference table then we do it the current 'simple' way, and if it's not then we do it the 'complex' way.

Once we've worked that out, either of these should be _reasonably_ easy to implement: the code for creating these complex dropdowns is already written to use with the edit functionality, it would just involve switching various other parts of the code to use this, and to deal with ID values instead of string values.

I don't think I'm explaining this particularly well, but we can talk about it more on Slack if necessary.

What do you think?

IanMayo:
Afternoon @robintw ,
I think we do need to include the extra fields. With just the sensor name we'll have lots of `GPS` entries, and now know which GPS it is.

(suspending answer while I flash up Windows VM to have a play).

IanMayo:
Sensor drop-down in current acugis database:
image

Hmm, `Type to filter` is effectively a `LIKE` statement.  So, I _think_ it's acceptable to only have `=`, `!=` operators.

Hmm, maybe not.  We have platform types like "Military - Aircraft" and "Military - Warship".  So analyst could select "Platform-Type" "LIKE" "Military" to find all military platforms.

That suggests we need both capabilities.  Aah, that meets your suggestion for a new field type. So, `Platform` would have all 3 fields (name, id,nationality) and only support =/!=m and `Platform Name` would just be the name, and would support `LIKE` operator.

IanMayo:
Yes, let's start with a minimal set and add others as they realise they require them.

That will control the level of clutter.

robintw:
@IanMayo I _think_ I have done all the changes necessary to do what we discussed (ie. provide both `platform` as a filter column, which gives a dropdown with all the details, and `platform name`, `platform identifier` etc - for all tables). I've tested it tonight, and I think it all works, but my brain is going to mush with all of the column data I've been looking at.

Could you have a go at testing this tomorrow? It'd be good if you could test all sorts of filtering of various sorts of tables, plus adding/editing of those tables (as we use the same column data dictionary as a backend to both operations). I will have limited availability tomorrow morning, hence telling you this now.

Note: I'm aware that the tests won't pass on this yet. I have a load of tests on the column data stuff that will need updating - but I wanted to check it actually worked in practice first before I updated all of the tests (they're failing because I changed the structure of column data, and I might have to change it again if things aren't working).

IanMayo (dismissed):
New displayed filter labels work fine, I like the free text search.

Conflict to be resolved before we can merge.

IanMayo:
Aah, have spotted something else.  If I'm showing filter drop-down, then change the field, the existing (now irrelevant) drop-down is still showing.

I can't click anywhere to collapse the (irrelevant) drop-down, though `esc` will collapse it.
image

IanMayo:
Afternoon @robintw  - I'm playing with maintenance tools to see if I can reproduce this afternoon's errors.

I can't so far. But I do have more feedback on editing Platforms.

I can't imagine a scenario when the user wants to edit platform-type or nationality by typing in the GUID.

Have we deliberately decided to show these fields, or are we using introspection to allow all table fields to be used in filter?

robintw:
@IanMayo I'll work on the first issue - I should be able to stop any other dropdowns opening once a dropdown is open.

As for the second: can you reliably reproduce it? I've done a few changes that should stop it happening, but I can't reproduce it on my machine.

robintw:
@IanMayo What do you mean by this:

> I can't imagine a scenario when the user wants to edit platform-type or nationality by typing in the GUID.
> Have we deliberately decided to show these fields, or are we using introspection to allow all table fields to be used in filter?

Where are you seeing the option to use a platform type or nationality GUID?

When looking at filters I don't see these:
image

and when looking at the edit dialog I don't see any option for GUIDs either:
image

IanMayo:
I get platform/nationality IDs in the `Select column` dropdown for `Platforms`:
image

Will download new version & try that.  Hopefully the newer builds will include the fix for #803 (database connection when first opening)

robintw:
Ahhh, I've just realised that it's a Postgres/SQLite issue, and you're running on Postgres. I'll fix it now.

(and yes, this PR's branch now includes the latest from `develop`, which includes the fix for #803)

IanMayo:
No longer have an issue with multiple inconsistent drop-down menus.

Is it only possible to drop out of a menu using `esc` key.  I can't find a mouse-click location that cancels a drop-down menu in the filter pane.

My expectation from "normal" UIs is that if I click anywhere other than the menu, it cancels the menu.  Or, if I click in the menu title.

IanMayo:
Yes - the database connection issue is sorted.
Yes - those columns are no longer offered.

robintw:
> Is it only possible to drop out of a menu using esc key. I can't find a mouse-click location that cancels a drop-down menu in the filter pane.
> My expectation from "normal" UIs is that if I click anywhere other than the menu, it cancels the menu. Or, if I click in the menu title.

I've now added the ability to cancel the menu by clicking on the menu title (ie. the bit you click to open the menu in the first place now closes the menu if you click it while it's open).

I don't think it's possible to have a 'click anywhere to close' operation in prompt_toolkit. All mouse handlers have to be attached to specific widgets, and there is no way to do this for just 'any empty space'.

Do you think a combination of either Esc or clicking on the title to close will be ok? We can point out the Esc option in the documentation.

IanMayo:
> All mouse handlers have to be attached to specific widgets, and there is no way to do this for just 'any empty space'.

But, if mouse events propagate up the UI component "tree", wouldn't there be a high level component (such as the whole app) that could handle it?

(it's quite likely that my above suspicions are mistaken ;-) )

robintw:
I don't think mouse handling does propagate up a tree in that sense, and mouse handlers have to be attached to individual pieces of text in the UI (slightly strangely, mouse handlers are actually part of the 'formatted text' specification), as opposed to key handlers which are attached to components and do some sort of tree-style handling.

I think it's best to create a separate issue for this, and I can investigate it later - and ask Jonathan if I'm still struggling and we really want it implemented. (There's a _possibility_ I could do something crazy with keyhandlers that also catch mouse events or something - but that will take a while to play with and then implement robustly).

IanMayo:
Sure, that's all fine @robintw

IanMayo:
@robintw non-UI test fail.  If we can get this fixed sometime tomorrow I can get the release to their IT in time for more testing on Thursday.

robintw:
Ah that's just a benchmark fail for an import - nothing of which was touched by this code. It seems to happen very occasionally - and I think Github Actions is being a bit slow today. I'll increase the threshold to try and make it fail less often, even if GH Actions is being a bit weird.

PR #803, opened 2021-03-08
Fix error when opening GUI

### LINKED ISSUE #802: Maintenance GUI needs "2. Status" before it will allow "8. Maintenance Tools"
I've entered Pepys-Admin to verify new maintenance tools.

If I try to go straight to maintenance tools, it tells me the database isn't initialised.

But if I run "2. Status" first, it's then possible to go into maintenance tools.

Video recording here: https://i.gyazo.com/191ede19fa4fb7e77d375c9292ea7cd6.mp4

(video link provided rather than GIF, since GIF was too small to be readable)

robintw:
## 🧰 Issue
Fixes #802 

## 🚀 Overview: 
This ensures that when we check the database when we load the GUI, we check it inside a `session_scope()` - otherwise we can't always access the database properly. This caused the bug in the attached issue, because the Status command created a session which could be used.

## 🤔 Reason: 
Fix bug

## 🔨Work carried out:

- [x] Change to check database within a `session_scope()`
- [x] Tests pass
## Confirmations

- [x] Any database schema changes are implemented via `alembic revision` [transitions](https://pepys-imxort.readthedocs.io/en/latest/database_migration.html#how-to-use-it-for-developers)

### DIFF
--- a/pepys_admin/maintenance/gui.py
+++ b/pepys_admin/maintenance/gui.py
@@ -81,13 +81,16 @@ class MaintenanceGUI:
 
         self.data_store.setup_table_type_mapping()
 
-        # try:
-        #     # This calls a simple function to check if the Privacies table has entries
-        #     # We don't actually care if it has entries, but it is a good simple query
-        #     # to run which checks if the database has been initialised
-        #     _ = self.data_store.is_empty()
-        # except Exception:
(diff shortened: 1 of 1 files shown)


IanMayo (approved):
Tested & verified as working.

PR #804, opened 2021-03-09
Allow relative time filtering

### LINKED ISSUE #797: Allow for relative time to be used as query filter in Pepys Maintenance
I'm starting to think through some graphical plots that could be generated in Pepys Maintenance, and potentially in the Status Board.

For the status board, the time filter could be one of:
- in past 24 hours
- in next 24 hours
- Yesterday
- Day before yesterday
- Today
- Tomorrow (used for tasks & participants)

Clearly these all just apply to fields of type "Time".

BarisSari:
## 🧰 Issue
Fixes #797 

## 🔨Work carried out:
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
I included a new library for testing new functions. It helps to mock DateTime objects: [freezegun](https://github.com/spulec/freezegun)

### DIFF
--- a/pepys_admin/maintenance/utils.py
+++ b/pepys_admin/maintenance/utils.py
@@ -1,6 +1,13 @@
 import re
+from datetime import datetime, timedelta
 
 import pint
+import sqlalchemy
+from sqlalchemy.dialects.postgresql import TIMESTAMP as PSQL_TIMESTAMP
+from sqlalchemy.dialects.sqlite import TIMESTAMP as SQLITE_TIMESTAMP
+
+from pepys_import.core.store.data_store import DataStore
+from pepys_import.utils.table_name_utils import table_name_to_class_name
 
 
 def get_display_names(fields, capitalized=False):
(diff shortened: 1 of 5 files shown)


pepys_admin/maintenance/utils.py
```diff
@@ -69,3 +76,72 @@ def get_str_for_field(value):
...
         return str(value)
+
+
+def create_time_filter_dict() -> dict:
+    """Creates a dictionary that has the filter strings as keys, and related datetimes as values."""
+    today = datetime.today()
+    today_as_date = today.date()
+    one_day_delta = timedelta(hours=24)
+    time_filter_dict = {
+        "In past 24 hours": (today - one_day_delta, today),
```
IanMayo, line 87: @BarisSari - could you store these strings a constants?  Then we can be sure the UI & the backend are expecting the same string.

BarisSari: Sure, @IanMayo 👍 I'm updating the code

BarisSari: Done: https://github.com/debrief/pepys-import/pull/804/commits/f0bca1867bc9fd93547cfab9ca22ca74520d2d42

robintw (changes requested):
Thanks @BarisSari, this looks good - I like how you've implemented the time stuff.

I think in the GUI, I'll be needing to use this alongside `filter_widget_output_to_query`, combining the filters from that function with the time-based filters. Therefore, it would be useful if the output of your function was a filter object, as opposed to the full results from a query object. For example, when using the output from `filter_widget_output_to_query`, I wrote some code like:

```
query_obj = self.data_store.session.query(self.current_table_object).filter(filter_query)
```
where `filter_query` was the result returned from the function.

If this function returns a filter too, then I should just be able to add another `.filter` on to the end of that.

Does that make sense? If we change it like this then it probably needs a new name - maybe `convert_relative_time_string_to_sqlalchemy_filter` or something?

BarisSari:
> Thanks @BarisSari, this looks good - I like how you've implemented the time stuff.
> 
> I think in the GUI, I'll be needing to use this alongside `filter_widget_output_to_query`, combining the filters from that function with the time-based filters. Therefore, it would be useful if the output of your function was a filter object, as opposed to the full results from a query object. For example, when using the output from `filter_widget_output_to_query`, I wrote some code like:
> 
> ```
> query_obj = self.data_store.session.query(self.current_table_object).filter(filter_query)
> ```
> 
> where `filter_query` was the result returned from the function.
> 
> If this function returns a filter too, then I should just be able to add another `.filter` on to the end of that.
> 
> Does that make sense? If we change it like this then it probably needs a new name - maybe `convert_relative_time_string_to_sqlalchemy_filter` or something?

Thanks, @robintw. I changed the code. It returns a `filter_query` now: https://github.com/debrief/pepys-import/pull/804/commits/d3301fa45b323d6a566dbcbb2dbe8851a4670410

robintw (approved):
Looks good, thanks Baris.

Issue #817, opened 2021-03-11
Maintainer tools freezes on Privacy table


IanMayo:
```
When I load up the Privacy table, with the intention of deleting the Private and Public entries, the whole 
window freezes up and I will have to restart Pepys Admin.
```

I'll try to reproduce it.

Note: I've just tried to reproduce this on the 16Mb sqlite database.   I was able to load the privacies table, but when I try to delete it has frozen.  _Maybe_ it hasn't frozen, but is taking a long time to sort out the dependent fields.

Aah, it has returned after a couple of minutes.
![image](https://user-images.githubusercontent.com/1108513/110828817-57ce3c80-828f-11eb-97f0-b6a678027cb1.png)

robintw:
I can't seem to reproduce this either. I suspect I may not have tested fully with the Privacies table, but it seems to work ok for me - it loads the table, I can then filter normally and edit successfully.

There is code in place that should catch any exceptions that occur during the creation of the column data for a table - which is what happens when you select a table from the dropdown on the top left. I'm intrigued that it actually hangs, rather than giving an exception of some sort.

Can you ask them whether the progress bar saying `Loading table data` appears, and roughly how far along the progress bar it gets before it crashes?

IanMayo:
Issue description updated, @robintw

robintw:
Ahh yes, freezing when they try and delete a privacy that is referenced in a lot of places makes more sense. I'll add a progress bar to the dependent objects calculation tomorrow - it'll be a bit tricky to work out how to increment the progress bar, as the time consuming thing is finding the objects, and so we don't know how many we have at that stage (ie. what the denominator would be for the proportion completed). I _think_ the best thing to do will be to have the progress bar go round and round (going 0 to 100, then back to 0 to 100 etc), as then at least it will be continually moving and showing that something is happening.

IanMayo:
I guess the denominator is the number of rows in that table.  The # of
row in the tables are probably:
* platforms: 50
* datafiles: 600
* states: 1.5m
* comments: 5k

So the denominator for the above would be ~ 1.505m, as we process
platforms numerator goes to 50, then 650, then 1.5m, then 1.505m (I
guess...)

On Thu, 11 Mar 2021 at 17:36, Robin Wilson ***@***.***> wrote:
>
[...]

PR #819, opened 2021-03-12
Add progress bar for dependent objects calculation - and speed it up

robintw:
## 🧰 Issue
Hopefully this will fix the problems the users were having in #817 - but they _said_ that it froze when they _selected_ the privacy table, whereas this fixes a free when you actually choose which ones to delete - so we'll have to get them to test it.

## 🚀 Overview: 
Added a progress bar while the dependent objects were being calculated. Also sped up the dependent objects calculation by around 3x (from 50s to 16s for my example), by caching some things that the sqlalchemy-utils function was calculating each time.

MaintenanceGUI_2021-03-12

## 🤔 Reason: 
Stops the GUI looking like it has frozen.

## 🔨Work carried out:
- [x] Added a progress bar dialog while the dependent objects calculation is carried out
- [x] Added caching of foreign key relations to the find_dependent_objects function
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
I've managed to find a way of doing a progress bar for this operation that doesn't go backwards, and still gives a reasonable idea of progress! We know how many objects we have in the list that we're going through (although we're adding to that list constantly), so we can use that as a sort-of denominator, with a big of extra logic so that the progress bar doesn't go backwards when we add a load of new objects to the list.

### DIFF
--- a/pepys_admin/maintenance/column_data.py
+++ b/pepys_admin/maintenance/column_data.py
@@ -187,7 +187,8 @@ def create_normal_column_data(col, data_store, table_object):
         # Skip all ID columns except the primary key
         return None, None
 
-    if details["type"] == "string":
+    # Skip getting values for the remarks column, as we don't need a dropdown for that
+    if details["type"] == "string" and details["system_name"] != "remarks":
         # Get values
 
         all_records = data_store.session.query(table_object).all()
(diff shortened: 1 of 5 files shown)


IanMayo (approved):
Tested on large database, working fine

Issue #828, opened 2021-03-18
Maintain tasks & participations


IanMayo:
## 🐞 Overview

Analysts need to add and maintain the tasks and participations that are used to indicate which platforms are tasking part in which tasks.

While we have a document format that sometimes can be used to populate the initial data, the nature of naval wargames means that we can expect a number of daily changes.  Pepys will also be used to record undertakings where the initial document isn't available.   The complexity of this task justifies the provision of task-based editing environment (form).

This form would be accessed via `(9). Maintain Tasks`.

## Design
![image](https://user-images.githubusercontent.com/1108513/111618252-0ddbde00-87dc-11eb-94b1-d01dec2ca5d6.png)

Original [here](https://www.figma.com/file/ln2z6r1h4OTZceH0xBtUbN/Tasks-%26-Participations?node-id=0%3A1) 

## 🔗 Feature

#601

robintw:
(Not urgent at all, I'm about to stop for the week)

What are the uniqueness criteria for a Task entry? Do they all have unique names? Or is it unique across `(name, parent)`, or do we have to consider dates as well?

IanMayo:
The tasks are unique for that parent.

It is quite likely that successive Joint Warrior exercises have serial called 05100.

robintw:
Current progress on tasks GUI shown in video below. Includes the TreeView widget I've written, plus loading the Tasks from the database, and adding a new entry to the TreeView (this doesn't yet add a new entry to the database, and you can't yet view information on the task).

Still lots more to do (including styling and lots of UI stuff), but I've made progress.

![TasksGUI_2021-03-22](https://user-images.githubusercontent.com/296686/111986277-a5576e80-8b05-11eb-9ade-44c1538288b3.gif)

IanMayo:
Great progress @Robin.
One comment - we have the user's endorsement to restict the data to 3 tiers.  So, we won't need the children of `ADEX 324`.

The presence of this constraint means we may be able to make other assumptions on the data, assumptions that let us change the UI according to level - such as:
- tasks in Tier one don't take a start/end date, or participants
- tasks in Tier two have start/end date and participants
- tasks in Tier three have start/end dates and a subset of the parent participants

robintw:
A few questions @IanMayo:

1. How is the force that a participant is part of represented in the database. I see there is a string `force` field in the `Participants` table - is it as simple as just setting that to `"Blue"` or `"Red"`? Can the force ever be anything apart from Blue or Red?
[...]

PR #833, opened 2021-03-19
Tasks/Participants editing GUI

robintw:
## 🧰 Issue
#828

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


IanMayo:
Morning :-)

On MS-Windows, I can't read the selected Task in the left-hand column:
image

IanMayo:
I'm connected to TracStore. 

When I expand the second task, I get this exception:
image

screencast

IanMayo:
We should offer **Confirm** on Delete.

Delete should cascade through child tasks & participations.

IanMayo:
I think we need to disable/enable the `Save task` button according to if name and privacy are provided.  TBH - the current greyed-out shade did look like disabled.  I think the '< >' markers denote _button_, so we prob don't need them in another shade.  So maybe this shade is disabled, and normal white is enabled.

Aah, for privacy, I did suggest by default providing the lowest (first) Privacy value.  These exercises (tasks) are typically published to the public - since they can introduce no-sail and no-fly zones.

On the current tracstor database I can't:
- expand `JOINT WARRIOR SERIES`
- add a child to `JOINT WARRIOR SERIES`

screencast

robintw:
I've got to go for childcare duty now, but I've completed a number of the issues you raised here. Specifically:

- The bugs you found with selecting/expanding/adding entries in the TreeView should be fixed
- I've changed the styling so that the TreeView entries are visible on Windows
- I've added a confirmation dialog on Delete
- Deletes now cascade through to children and participants
- The lowest privacy value is provided as a default

I haven't yet done:
- Disabling the Save button when the form isn't in a state where it can be saved (eg. entries are missing) - this will come with the validation of the form in general, which I haven't implemented yet
- Tests for the creation of Task entries, checking the cascade works properly and that children/participants are deleted, that a proper hierarchy of `TreeElements` are created given a hierarchy of Task elements in the database etc. I'll work on these next, as it will give me more confidence that all the other implementations have worked correctly.

As discussed, there will be a decrease in work on this next week while I'm at my conference (and also dealing with my son home self-isolating).

IanMayo:
That's all great, thanks @robintw .

robintw:
@IanMayo I've made a lot of progress with the Tasks GUI, and I think it's now ready for you to play with (and try and break!).

You'll need to check out this PR, and make sure you've run migrate on your database (or use a blank SQLite database with the `--db blah.sqlite` option to Pepys Admin). If you want a 'ready-made' database with some tasks already entered into it, then feel free to use this database: [test_tasks.db.zip](https://github.com/debrief/pepys-import/files/6316357/test_tasks.db.zip)

The UI looks like this at the moment:
image

I'm sure there will be some UI changes you want, and you'll probably find some bugs - so have a play and see what you think.

IanMayo:
First error. I opened the new page with an uninitialised database. It fell over because the `Serials` table was missing.

It all worked once I'd initialised the database, but a more graceful failure would be less cary.

IanMayo:
* [ ]  I have a newly initialised database - with no wargames or serials.  I don't have an "Add" button for "Tasks".
* [ ] Now that we've moved to a fixed set of tiers, the left-hand can be called "Wargames" and the right-hand pane "Serials".
* [ ] I'm still in this empty condition.  If I click `f3` to move to the right-hand pane I get an error.
* [ ] I don't see the new tables when in `maintenance tools`.  It's not urgent to have that, since we have the dedicated UI for editing Wargames/Serials/etc, but they may find it useful for some bulk tasks.
* [ ] I've switched to the PostGres database.   I've. tried to rename `Test Series` to `Jolly Warrior`.  Should it rename as I leave the text box, or does it need a `Save` button.

I've found the cause of some of the above.  I bet there are buttons, but I can't see them with black background:
image

robintw:
I'm just about to go on childcare duty - I'll look at some of these in more detail later this afternoon. However, briefly:

- I've pushed a change that sets the button text colour back to gray. I'd done a quick switch to black and hadn't tested it properly with a black background. I'll need to check how we dealt with it in the GUI to make it work on both a white background and a black background.
- Renaming (or any edits) need the save button clicking - obviously it'll help when you can see the save button!
- You're right that I need to deal with an uninitialised database - I'll fix that
- I'm not sure whether we can rename the LH pane to Wargames and the RH pane to Serials, because the right-hand pane is where the viewing/editing happens for _any_ object - it could be a Series, a Wargame or a Serial. What do you think?

I'll work on this more later this afternoon.

IanMayo:
I'm able to do valid test now I've reversed background color.  It's mostly working as we'd expect it to.

> title of right-hand panel
As a newbie, it is initially challenging to remember at what level we're working at.  But, I see the current item remains selected on the left-hand side.  Aah, I also see the first field includes the data type (series, wargame, serial)

* [ ] It would be useful to see if a serial participant has a restricted time period. It looks like we've got sufficient horizontal space to do that 
* [ ] I don't appear to be able to edit existing serial-participations
* [ ] It's a minor pain to enter all of the dates.  For serials, could we initialise the date-time field with the start date-time of the wargame?
* [ ] For a wargame there could be 30 participants, for a serial there could be 20.  I think we should grow the Wargame participants to 15 entries, and the serial participants to 15 for Blue and 5 for Red.  Hmm, if we move the participant lists to the right of `Blue/Red Force Participants` it would gain us a few lines of vertical space.
* [ ] Is there an opportunity to switch `<Add task>` and to `<Add wargame>`? (or whatever this level is).

robintw:
Working through these now @IanMayo.

Nearly all of the first set of issues you reported are now fixed, apart from the issues with editing the tables in the Maintenance GUI. That's going to take a bit of work - as I'm using some new SQLAlchemy features that the Maintenance GUI doesn't yet support. I'll prioritise getting the rest of the Tasks GUI finished, and then tackle that.

As for your second set of comments:

1. What do you mean by:

> Is there an opportunity to switch `<Add task>` and to `<Add wargame>`? (or whatever this level is).

I've searched the code, and can't find a button that says `Add task` - which button were you referring to?

2. Currently editing of participants isn't supported - you have to delete the participant and add a new one. I left editing because I wasn't sure how I should work: would you like an extra `Edit` button alongside the `Add` and `Delete` buttons?

I should be able to fix the other issues relatively easily.

Also: what are the timescales for using this looking like from the client's perspective? I said I'd aim to get it finished this week, but some SQLAlchemy issues meant it took longer than expected. I'm a bit limited in availability tomorrow (although I will be able to do some work), but don't want to rush to try and get a finished Tasks GUI and then find the clients won't be able to install a new version and try it until the middle of next week.

robintw:
@IanMayo I think the suggestion to move the participants list to be on the same line as the label for the list will conflict with the request to add the start/end periods to the list. Now that they're on the same line as the label, there is less space for the entry - and we're starting to get very close to the end of the line:

image

Which requirement do you think is more important?

(I'm also aware that the Windows terminal probably has a smaller width anyway, so it would be even more of a problem there)

robintw:
Sorry, another question. You said:

> For serials, could we initialise the date-time field with the start date-time of the wargame?

Do you want just the start time pre-populated, or the end time as well?

robintw:
>More specific buttons (`Add wargame` rather than `Add`, and `Save wargame` rather than `Save`

Done

>If we provided default values for start and end, at least the month/year values would be present already.

Done

>Trimming nationality and identifier

Done

>Editing

The problem with popping up an edit dialog when you select the entry in the list, is that you have to select the entry before you can delete it (using the delete button). I'll implement an `Edit` button for now, but we can revisit this if necessary.

I think I've now dealt with all of your comments apart from adding the ability to edit. I'm going to stop for today, and focus on editing either tomorrow (if I have time) or at the weekend.

IanMayo:
> click to edit

Aah, I see.  In that case, then yes - `<Edit>` is the perfect solution.

IanMayo:
1. I did have an error when creating a series for a newly created series in an new database, but I can't reproduce it.
2. We can reduce the amount of data to be entered, and save some screen space if we remove seconds for creation & view of serials & participation periods.

I've had quite a long play, creating and deleting series, wargames and serials, and can't find anything else to report.  No crashes or errors at all. 

Aah, one error.  It's probably because the new data isn't incorporated into the maintenance UI yet.  I wanted to change the nation of `DOLPHIN_FR` from `UK` to `France`.  I filtered the platforms, selected the platform to edit, and got this error when I clicked on `Edit values`.
image

IanMayo (approved):
This was a big old package of work. Thanks @robintw

PR #852, opened 2021-04-21
Changes to make Maintenance GUI work with new Tasks/Participants tables

robintw:
## 🧰 Issue
N/a

## 🔨Work carried out:

- [x] Add a 'skip_in_gui' attribute to relationships that we don't want to display in the GUI
- [x] Make sure all Task/Participant relationships use 'lazy=joined' to solve detached instance errors in GUI
- [x] Add some more association proxies to get things we want to be able to filter by in the GUI
- [x] Change default fields display for SerialParticipants to something that doesn't cause an error
- [x] Change placing of expunge call in run_query to expunge _after_ we've got the data into the table view
- [x] Processes ObjectAssociationProxies as well as ColumnAssociationProxies, and loop through when processing to get the final thing that the assoc proxy points to
- [x] Exclude fields that are lists (eg. a list of participants etc) from being edited in the Edit dialog
- [x] Tests pass

## Confirmations


robintw:
Hi @IanMayo. I've almost got this finished now - editing of all of the new tables now works in the Maintenance GUI. I've got a couple of questions:

1. Various attributes of the database objects can now contain lists of items - eg. a list of participants. These aren't actual fields in the database, they're clever SQLAlchemy attributes that transparently deal with relationships between tables. This works fine - it means we can filter on some useful things (eg. find all of the wargames that a particular platform participates in), and I've excluded these from the edit dialog (as we don't need to edit them directly). We can also view these fields in the preview table - and currently they look like this:
image
You can see they're shown as a standard Python list with square brackets and values. Should that be left how it is, or should we display it in a different format - eg. just the values with commas between them? or something else?

2. Currently the database itself doesn't stop someone from creating duplicate WargameParticipant entries - that is, two entries with the same `wargame_id` and the same `platform_id`. The Tasks GUI stops you from entering these (it only offers platforms that aren't already participating in the wargame as options to select from), but you can do it in the Maintenance GUI. Should we make this impossible in the database, as it makes no sense in real-life? We could do it with a unique constraint on `(wargame_id, platform_id)`.

IanMayo:
Hello @robintw ,
1. Could we display the quadgraph for each platform?  They're generally quite human-readable, and of predictable length.  I'll encourage the users to invest effort in maintaining them.
2. Hmm, no.  I only imagine "maintenance tasks" being used for searching for a participation, or maybe bulk deletion.

robintw:
@IanMayo

In response to 1) Actually, we have various fields that might display as lists - for example a list of `child wargames` of a Series entry, or a list of `participating wargames` for a Platform entry. The question was really just how these lists should be formatted - is it ok to format as a Python list `[value1, value2 etc]` as in the screenshot above, or should we just separate with commas (possibly looks nicer, but less explicit that it is a list).

IanMayo:
> @IanMayo
> 
> In response to 1) Actually, we have various fields that might display as lists - for example a list of `child wargames` of a Series entry, or a list of `participating wargames` for a Platform entry. The question was really just how these lists should be formatted - is it ok to format as a Python list `[value1, value2 etc]` as in the screenshot above, or should we just separate with commas (possibly looks nicer, but less explicit that it is a list).

I can't think of any instances when we support commas within Pepys fields.  So, just having commas will probably work fine - it will be clear they refer to a list, I think.

robintw:
Right, I think this is all ready now. This PR needs merging before #855 - and I'll now pull these changes into #855 to test that the issues you found in #855 are fixed by this PR. Once this has merged, you can merge #855.

IanMayo (approved):

tests/gui_tests/test_column_data.py
```diff
@@ -1374,6 +1557,8 @@ def test_edit_data_platform():
         },
         "privacy": {
             "foreign_table_type": TableTypes.REFERENCE,
+            "multiple_values_allowed": False,
```
IanMayo, line 1560: We need to lose one of these duplicated lines, please @robintw 

(though I expect it doesn't have runtime impact)

robintw: Fixed now.

IanMayo:
Codacy is moaning about the complexity of this function, but it looks fine to me:
image

robintw:
Yeah I saw that, and I don't think it's a problem. I've always found automated ways of assessing the complexity of a function give quite unreliable outputs that often don't align with my thoughts on it!

IanMayo (approved):

PR #876, opened 2021-05-04
Don't initialise the database on import unless it is completely empty

### LINKED ISSUE #874: Don't re-initialise database at start of import unless number of tables is zero
Currently, if `is_schema_created()` returns False, then `data_store.initialise()` is called at the beginning of an import run. This can be a problem because `is_schema_created()` returns False if the number of tables doesn't match the expected number of tables - not just if the number of tables is zero.

If the number of tables is zero (ie. database is completely empty, or brand new), then running initialise is the right thing to do - this allows you to run Pepys Import on a non-existent SQLite file and it will be created and initialised automatically. However, if there are some tables there already, we don't want to initialise - we should give the user a message and tell them to run the migration functionality.

The _only_ place that `is_schema_created` is used as a condition for calling `data_store.initialise()` is in `pepys_import\cli.py`.

robintw:
## 🧰 Issue
Fixes #874.

## 🚀 Overview: 
Changes the logic in the import process. Previously, if the number of tables in the database didn't match the number it expected to see, then it re-initialised the database. This could lead to problems - particularly if conflicting versions of Pepys were used by different users. Now, it will only initialise the database if there are zero tables currently in the database - otherwise it will give an error asking the user to run the database migration function.

## 🤔 Reason: 
Stops the database getting into confused states.

## 🔨Work carried out:

- [x] Change logic for initialising on import
- [x] Add some more tests
- [x] Tests pass

## Confirmations


tests/test_import_cli.py
```diff
@@ -407,3 +407,107 @@ def test_skip_validation(patched_prompt):
...
+        # so we drop the table and create a new table instead
+        conn.execute("DROP TABLE Geometries;")
+        conn.close()
+
+        temp_output = StringIO()
+        with redirect_stdout(temp_output):
+            process(path=DATA_PATH, archive=False, db="cli_import_test.db", resolver="default")
+        output = temp_output.getvalue()
+
+        print(output)
```
IanMayo, line 435: @robintw - is printing the output to console just needed during development?  I see we test it in code in subsequent lines, so suspect we don't need to do it here or at line 498.

robintw: I'm currently debugging a test failure that only seems to occur on the CI - hence adding the print statement. They will be deleted once I've sorted the failure.

IanMayo: Is it now ready to go @robintw ?

robintw: I've just pushed a commit to remove those print statements as it now works on CI. Should be ready to merge now.

IanMayo (approved):
Great, thanks.

PR #879, opened 2021-05-04
Add stored procedures for dashboard queries

### LINKED ISSUE #840: How to generate "Coverage" statistics
A major part of the status-board we're developing is the publishing of "coverage" for a platform.

image

This data is constructed as follows:
- find the 3rd-level tasks that happened yesterday
- for each task:
    - for each participant:
        - find the period for which they took part in the task (potentially only part of task)
        - find their default state data interval (new field in PlatformType). Something like 1 minute for ships, and 1 second for aircraft
        - examine the State measurements for the time period.  Identify any periods where there is an interval of more than 5 times the default data interval.  These form "gaps", the other periods form "coverage".  Return periods of coverage
        - also sum up the periods of coverage, and divide by the participation period - to give %age coverage
    - calculate an overall %age coverage for the task (mean of participations)
- calculate an overall %age coverage for the day (mean of tasks)

Working out gaps/coverage in the State data could be an enormous task.  But, once we've filtered the data down to one participant, for the period of participation in the serial, I think it becomes manageable.

At the start of the day, there will be no periods of data loaded.  The display will just show the recorded start/stop times for Platforms taking part.  As data gets imported into Pepys these periods of coverage will increase.   I guess we would update the UI every minute, or every few minutes.

But, once a block of states have been processed - to give the coverage, we don't need to process those states again.  So, that suggests we should cache the coverage periods - a table is the _usual_ place to do that.

Note: the participation period may change - if the analysts find out a platform actually took part in the whole task, not just the first half. When this happens I _don't think_ it would invalidate any previously calculated coverage, but may introduce a requirement for new coverage to be calculated.

So, questions:
1. _Where_ do we do the above processing?
    a. PostGres - this is near to the data, I think the code runs pretty well, it's easy to access results from SuperSet
    b. Python - as data is imported, the periods of coverage could be calculated, and stored in a new table.  But, we're calculating coverage for periods that potentially won't be considered (outside period of tasks)
    c. SuperSet/JS - _maybe_ it's possible to process some of this within Apache Superset. On one level it could be seen as a custom visualisation of State data - though we're not making the %age coverage visible elsewhere
    d. SuperSet/Python - alternatively we try to process the data in the SuperSet backend, if that's possible.
2. _When_ do we do the above processing?
    a. On import, in Python.  It's easy to do the processing at this point, though at a processing cost.  We can build it into the core processing, so it's hidden from custom import code.  Since we can't trigger re-processing if participations change, then we have to do it for all state data.  This strategy also works best for SQLite, which can't have stored procedures/functions. 
    b. On import, in SQL.  This method is robust to loading data from non-Python sources (of which none are forecast).  It's also quick, and doesn't involve extra data going over the wire - since it would be a "hook" on new state data.
    c. Live, on SuperSet query.  When SuperSet requests data for a Participant, check if there is any coverage data available. If there isn't, generate it & store it, return it.

Any thoughts/tips/suggestions/worries about the above are quite welcome.

robintw:
## 🧰 Issue
Fixes #840.

## 🚀 Overview: 
Adds stored procedures to Postgres databases for calculating coverage statistics to display in the dashboard. Includes code to create these procedures for newly-initialised databases, and to add them to existing databases through a migration.

## 🤔 Reason: 
Enable dashboard

## 🔨Work carried out:

- [x] Add stored procedure .sql files
- [x] Add code to execute these files on database initialisation
- [x] Add migration to add these procedures to an existing database
- [x] Tests pass

## Confirmations


IanMayo (approved):
Great. My first thought was that "we need to handle the situation where the function is already present on the server, like with tracstor".  Obviously @arunlal-v had already thought of that :-)

PR #880, opened 2021-05-04
Add coverage dashboard

### LINKED ISSUE #878: Integrate Flask timeline into Pepys-Admin
To help the devs get up & running as quickly as possible, the timeline was developed in a different repo (`Pepys-Timeline`) - and thus avoid/defer needing to learn about the pepys architecture.

But, delivering the timeline within Pepys-Admin is seen as the most convenient way of it being made available to analysts.

After this has happened, subsequent maintenance of the timeline will happen in `Pepys-Import`.

robintw:
## 🧰 Issue
Fixes #878 

## 🚀 Overview: 
Adds the Flask-based dashboard app from the [pepys-timeline](https://github.com/debrief/pepys-timeline) repo into this repo, and integrate with the Pepys Admin application.

Dashboard

## 🤔 Reason: 
Provide dashboard functionality for user

## 🔨Work carried out:

- [x] Copy code from pepys-timeline repo into this repo
- [x] Update config code for pepys-timeline app to use the main Pepys Import config file
- [x] Add function to run Flask app from within Pepys Admin
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
There are multiple ways to run Flask apps from within a Python command-line app. You can use the built-in Flask server, but this prints out warnings to the console that it should not be used in production - and that might scare the users. Therefore, I've switched to using the Twisted server, which is 'production-ready'. I haven't yet tested this on Windows though, so there might be some installation issues.

Currently, the only way to close the server is to close the Command Prompt window (or press Ctrl-C). Ideally we'd run the server in a separate process so that we could display a nice prompt within Pepys Admin asking the user whether they want to exit the server, and then we can take them straight back to the Pepys prompt.


krsnik93:
@robintw looks good.

Regarding production-ready web servers, I have not used Twisted previously. I think it might be an overkill for our use case, as it's not just a WSGI server, but a full-blown networking framework. Usually when going to prod and replacing the built in Flask server, I would use Gunicorn: https://github.com/benoitc/gunicorn.

If Twisted was already being used in `pepys-import`, I think it would be an appropriate choice in order to avoid having to add another dependency. But it seems it was added in this PR, so I think we should consider switching to something lighter.

IanMayo:
Great, thanks @robintw.

The 3rd party dependencies make this a huge PR - I'd like to get it merged so that future changes are more clear to identify/review/navigate.

One item of feedback.  I think we need to give a bit more guidance in the pepys-admin window once the server is up & running.  We don't need to expose all the underlying technology, but this will be a power-user.

I think something like:
```
The Pepys timeline dashboard process is now running on: localhost:5000.  A browser window should have opened displaying the timeline.
Keep this window open for the server to continue running. The server process can be terminated by closing this window.
```

robintw:
I agree that Twisted is a large dependency @krsnik93 - but it was the best I could find at the time. Gunicorn doesn't work on Windows (our clients will be running Windows), and various other servers were tricky to install.

However, I've just found a server called [waitress](https://docs.pylonsproject.org/projects/waitress/en/latest/usage.html) that I've seen recommended in various places. It seems lightweight and is marketed as a cross-platform alternative to Gunicorn - so I'll switch to that.

IanMayo (approved):
Waitress working fine under MS-Windows

Issue #882, opened 2021-04-29
Indication of time since last update / time to next update


IanMayo:
I get impatient looking at the chart, waiting for it to update.

If there was some readout of time elapsed  it would manage my expectations.

A countdown timer like this would be fine:
https://codepen.io/zebateira/pen/VvqJwm

![timer](https://i.gyazo.com/ed30cce6c8c44bffa9105f3a2505750f.gif)

IanMayo:
Closed in #903

PR #903, opened 2021-05-06
[wip] add countdown timer

krsnik93:
## 🧰 Issue

#882 

## 🚀 Overview: 

Adds a countdown timer into the toolbar, to the right of the dates. Timer is initialized at the `frequency_secs` returned from the database and then counts down until the next data fetch. Consists of a number in the middle surrounded by a circular progress bar.

## 🤔 Reason: 

So that dashboard user can be more aware of when the data fetches are happening.
## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

countdown-timer-demo

## Confirmations

## 📝 Developer Notes:

Not looking to merge immediately, would like feedback regarding positioning, colors etc.

I added a new dependency for this task: https://github.com/albert-gonzalez/easytimer.js.

I think we should try to switch to React for UI at some point.


pepys_timeline/static/js/src/pepys.js
```diff
@@ -6,7 +6,56 @@ const DATE_FORMATS = {
...
+const onTimerSecondsUpdated = (event) => {
+  const { detail } = event;
+  const { timer: t } = detail;
+  const { seconds } = t.getTimeValues();
+  countdownNumberEl.textContent = seconds;
+  updateCountdownProgress(seconds);
+}
+const onTimerTargetAchieved = () => {
+  fetchSerialsMeta();
+  timer.reset();
```
IanMayo, line 37: I _think_ the correct logic is to reset the timer _then_ fetch the data.

Otherwise the update cycle is one minute _plus_ the time to retrieve the data.

(happy to discuss, good chance I have that wrong 🤣  )

IanMayo (changes requested):
They're quite minor, but could you fix these JS warnings please?
image

pepys_timeline/static/js/src/pepys.js
krsnik93: I think JS fetch is asynchronous, so it will not wait for the whole round trip time to reset the timer, but maybe we could flip the two lines just to be completely precise.

krsnik93:
> They're quite minor, but could you fix these JS warnings please?
> image

Strange error as the `setInterval` line was removed in this PR. I will take a look into it later.

IanMayo:
@krsnik93  - did you consider a timer component that included its own UI? 

I'm happy with your decision, but I'd like to understand your thought process.

krsnik93:
> @krsnik93 - did you consider a timer component that included its own UI?
> 
> I'm happy with your decision, but I'd like to understand your thought process.

That was my original attempt, however it became difficult to keep the two timers (UI one for the countdown and the "backend" one for fetching serial data) synchronized, so I was looking for a way to simplify and unify.

krsnik93:
> They're quite minor, but could you fix these JS warnings please?
> image

@IanMayo I cannot replicate this warning, which browser are you using? Seems like we might still have some cache issues.

IanMayo:
they are from the static analysis tool.

IanMayo (dismissed):
Nice work @krsnik93  - sorry if those linter fixes were a real pain. I hadn't appreciated how many it would throw up.
Thanks for persevering.

robintw (dismissed):
Looks good, thanks.

One very minor point: you don't need to add the `interactive_review` tag for PRs that are only affecting the web-based UI (ie. the timeline). The `interactive_review` tag starts an interactive demo of the command-line UI, which you aren't currently working on. I'm aware that the instructions in the PR could do with some clarification on this - I'll fix that shortly.

IanMayo:
Sorry @krsnik93 - in order to get a release out today, I tried to resolve your conflicts with develop.  But, I broke it - probably in the CSS.  I'd appreciate you fixing the problem(s) I've generated 😭 

Here's the commit that caused the problem:
https://github.com/debrief/pepys-import/pull/903/commits/6f732811c278ec6466a78ff011b8fc4e9ff77f7b

(I'll make a release without the countdown timer)

krsnik93:
@IanMayo I think it's fixed now. I made the colors lighter again as we moved the countdown timer from the lighter toolbar to the darker navbar.

image

IanMayo (approved):
Great, thanks.

PR #891, opened 2021-05-05
Add Arunlal's original test functions from the pepys-timeline repo

robintw:
## 🧰 Issue
#890

## 🚀 Overview: 
Copies across @arunlal-v's original SQL test functions from the pepys-timeline repo.

Some fairly minor alterations have been done to enable the tests to be run using pytest, and to fix formatting issues so we don't get black/flake8 errors on the CI.

## 🤔 Reason: 
Tests are good :)

## 🔨Work carried out:

- [x] Copy tests across
- [x] Modify tests to work with pytest
- [x] Reformat file so that black/flake8 are happy
- [x] Tests pass

## Confirmations

- [x] I have chosen reviewers for my PR.
- [x] I have chosen an appropriate label for the PR, adding `interactive_review` if reviewers will need to see UI
- [x] I have extended/updated the documentation in `\docs` folder
- [x] Any database content changes (Create, Edit, Delete) are recorded in the Log/Changes tables
- [x] Any database schema changes are implemented via `alembic revision` [transitions](https://pepys-import.readthedocs.io/en/latest/database_migration.html#how-to-use-it-for-developers)
- [x] I have completed the mandatory sections of this document.
- [x] I have deleted any unused sections.

PR #906, opened 2021-05-08
Add dashboard sql tests enhancement

### LINKED ISSUE #905: Bug in dashboard_stats query
## 🐞 Description

act_with_same_part_and_gap_end CTE in dashboard_stats query is having a wrong join clause (line 315 in the file) which causes the query to send two distinct and adjacent coverage sections, instead of one consolidated coverage section. i.e returns ["C","C"] instead of ["C"]

## ⚙️ Current behavior

Returns ["C","C"] instead of ["C"]

## ⚙️ Expected behavior

Should return ["C"]

arunlal-v:
## 🧰 Issue

fixes #905

## 🔨Work carried out:

- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_import/database/postgres_stored_procedures/dashboard_stats.sql
+++ b/pepys_import/database/postgres_stored_procedures/dashboard_stats.sql
@@ -312,7 +312,7 @@ act_with_same_part_and_gap_end as (
 				and cg.end_time =  s.time
 			inner join
 		sensors_involved si
-				on si.serial_participant_start = cg.start_time
+				on si.serial_participant_end = cg.end_time
 				and si.platform_id = cg.platform_id
 				and si.serial_id = cg.serial_id
 				and si.ser_idx = cg.ser_idx
--- a/tests/timeline_tests/test_dashboard_sql_functions.py
+++ b/tests/timeline_tests/test_dashboard_sql_functions.py
@@ -197,6 +197,7 @@
 import json
 import os
 import unittest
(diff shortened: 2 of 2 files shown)


arunlal-v:
>Could you fill in a few more details in the body of the PR (eg. the Overview and Reason sections), and emphasise that this PR includes a bug-fix for the SQL query itself.

Done. 

>From my reading of the issue detailing the bug, it seems like this bug won't actually affect the resulting display shown to the user - it will just mean there are two consecutive 'coverage' sections, rather than one big coverage section. Is that correct? If so, it means the bugfix is less urgent to deploy to the users.

Yep, that's right. The fix is not urgent. This is an edge case.

robintw:
Thanks @arunlal-v - that's great.

I see that this PR is going to merge into the `add-dashboard-sql-tests` branch - which is the branch of another PR. In general, we prefer to have PRs going directly into the `develop` branch - so I think it'd be good to either replicate these changes in the `add-dashboard-sql-tests` branch, or alternative change this PR to be merging into `develop` (and make sure @IanMayo merges the `add-dashboard-sql-tests` PR first). I don't know if it's possible to change which branch a PR is going to merge into - it may require closing the PR and creating a new one.

robintw:
Ok, we can do it that way around.

There is already a PR to merge the other branch to develop - see #891.

If we're going to merge this into `add-dashboard-sql-tests`, could you fix the merge conflict?

PR #915, opened 2021-05-14
[wip] Timeline diagnostics

### LINKED ISSUE #897: Timeline diagnostics
Connectivity for the deployed instance should be fairly solid, and not subject to change.

But, it would help their diagnostics if we did some reporting via the timeline UI, just in case something does go wrong.  Diagnostics could also be useful during development, when things may be a bit more flaky.

I suggest that the `timeline` web-page report these problems:
1. backend not running (REST queries unresponsive)
2. backend running, but database name, role, password or port not set
3. backend running, able to connect, but other database error
4. backend running, able to connect, no serials for specified date period

I _think_ that 1 & 4 can be done purely in JS.  I think 2 & 3 require some level of interaction with Python.  Maybe the `config` endpoint could return some error messages in response to 2 & 3.

krsnik93:
## 🧰 Issue

Fixes #897

## 🚀 Overview: 

Added error message display for different failure scenarios as outline in #897.

## 🤔 Reason: 

To give more detailed feedback to the user regarding the state of the app.

## 🔨Work carried out:

- [x] Tests pass

## 🖥️ Screenshot

image

## Confirmations

### DIFF
--- a/pepys_timeline/db.py
+++ b/pepys_timeline/db.py
@@ -34,10 +34,10 @@ def get_query_result(query, vars_=None):
                 with conn.cursor(cursor_factory=RealDictCursor) as curs:
                     curs.execute(query, vars_)
                     result = curs.fetchall()
-            except psycopg2.Error:
-                raise DatabaseQueryError('Error querying database.')
-    except psycopg2.Error:
-        raise DatabaseConnectionError('Error connecting to database.')
+            except psycopg2.Error as e:
+                raise DatabaseQueryError('Error querying database.') from e
+    except psycopg2.Error as e:
+        raise DatabaseConnectionError('Error connecting to database.') from e
     return result
 
 
--- a/pepys_timeline/db.py
+++ b/pepys_timeline/db.py
@@ -43,9 +43,9 @@ def get_query_result(query, vars_=None):
 
 def get_config_options():
     res = get_query_result(CONFIG_OPTIONS_QUERY)
-    refresh_row = [
-        row for row in res if row['name'] == 'TimelineRefreshSecs'][0]
-    refresh_row['value'] = int(refresh_row['value'])
+    for row in res:
+        if row['name'] == 'TimelineRefreshSecs':
+            row['value'] = int(row['value'])
     return res
 
 
--- a/pepys_timeline/db.py
+++ b/pepys_timeline/db.py
@@ -42,7 +42,11 @@ def get_query_result(query, vars_=None):
 
 
 def get_config_options():
-    return get_query_result(CONFIG_OPTIONS_QUERY)
+    res = get_query_result(CONFIG_OPTIONS_QUERY)
+    refresh_row = [
+        row for row in res if row['name'] == 'TimelineRefreshSecs'][0]
+    refresh_row['value'] = int(refresh_row['value'])
+    return res
 
 
 def get_dashboard_metadata(from_date: str, to_date: str):
--- a/pepys_timeline/db.py
+++ b/pepys_timeline/db.py
@@ -10,6 +10,9 @@ from pepys_timeline.queries import (
     DASHBOARD_METADATA_QUERY,
     DASHBOARD_STATS_QUERY
 )
+from pepys_timeline.exceptions import (
+    DatabaseConnectionError, DatabaseQueryError
+)
 
 
 def get_db_conn_kwargs():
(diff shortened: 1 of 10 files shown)


krsnik93:
@IanMayo do you see any way to configure Codacy through its GUI? It seems to throw up new lint issues randomly... I created a config file for ESLint but to no avail.

krsnik93:
@IanMayo I think Codacy is looking at stale file versions. This is what it complains about in the latest run:
image

If I go into the file, the line number does not correspond to what's there currently:
image

I have run ESLint locally and got it to pass. I am not sure why Codacy is behaving differently.

IanMayo:
sorry it's being a pain. If it's not bringing value we should probably drop it.
Job for Monday...

IanMayo:
@krsnik93 - right now I'm unable to deliberately break the database & see the errors.

Could you please give @robintw & me a summary of the ways in which you've tested the handlers?

Which types of failure were you able to test?

krsnik93:
> @krsnik93 - right now I'm unable to deliberately break the database & see the errors.
> 
> Could you please give @robintw & me a summary of the ways in which you've tested the handlers?
> 
> Which types of failure were you able to test?

These are the ones that I tested out:

1. to test out database connection, change one of the params in `get_db_conn_kwargs` in `pepys_timeline.db`, for example port to 1234
2. to test out query error, change one of the queries, for examples `pepys_timeline.queries.CONFIG_OPTIONS_QUERY` to 'blabla'
3. To test out 0 serials returned, just set the date range to yesterday

I have not been able to simulate the scenario when the flask app is completely down. When I was checking it yesterday, frontend simply did not do the network calls when the Flask app is down. this is something to drill down into next.

krsnik93:
Regarding Codacy, I don't think we necessarily have to drop it. We should definitely have some type of static code analysis going, whether with Codacy, Git Hooks or something else is less important. I do think there is some sort of a problem currently with Codacy.

IanMayo:
Thanks @krsnik93 the test for flask being down is a useful one, since the Python window may have been killed (or PepysAdmin closed) but the browser may still be open.
I guess we do this in the error handler for the 'get/post' call from JS.

krsnik93:
image

@IanMayo Please let me know if the message is ok as is or if you would prefer a different one.

krsnik93:
Way to test this error would be to do `python pepys_admin/cli.py` and then once the browser loads, kill the python process with:
`ps ax | grep python` and `sudo kill -9 <python_process_id>` and then wait for the next fetch.

krsnik93:
There was a bug because the refresh frequency from the database was a string. I wonder if it would be better to hold config options in Python seeing that the different params will have varied data types and so it's difficult to keep them in a single relational table.

IanMayo:
> image
> 
> @IanMayo Please let me know if the message is ok as is or if you would prefer a different one.

I'd like it in user-speak, please. I personally aren't sure which error this is, so I can't give you the words. Could you have a go at wording it in user-speak please?

IanMayo:
> There was a bug because the refresh frequency from the database was a string. I wonder if it would be better to hold config options in Python seeing that the different params will have varied data types and so it's difficult to keep them in a single relational table.

There isn't a central place to store this type of option in an easily editable form.  I guess we have to assume all data from the `ConfigOptions` table are strings.

krsnik93:
image

Please let me know if it still sounds too technical

pepys_timeline/db.py
```diff
@@ -23,30 +26,38 @@ def get_db_conn_kwargs():
...
-            config_options = curs.fetchall()
-    return config_options
+    try:
+        with psycopg2.connect(**db_conn_kwargs) as conn:
+            try:
+                with conn.cursor(cursor_factory=RealDictCursor) as curs:
+                    curs.execute(query, vars_)
+                    result = curs.fetchall()
+            except psycopg2.Error:
+                raise DatabaseQueryError('Error querying database.')
```
IanMayo, line 38: @krsnik93 is the actual error logged anywhere?
If it isn't easily accessible then I suggest it gets included in the message, so the users can pass it on to us.

IanMayo (changes requested):
(have indicated that there may be a need for more detailed diagnostics going to the rendered output)

pepys_timeline/db.py
```diff
@@ -23,30 +26,38 @@ def get_db_conn_kwargs():
...
+    except psycopg2.Error:
+        raise DatabaseConnectionError('Error connecting to database.')
+    return result
+
+
+def get_config_options():
+    res = get_query_result(CONFIG_OPTIONS_QUERY)
+    refresh_row = [
```
robintw, line 46: This code looks a bit clumsy - it looks like you're editing `refresh_row` and then returning `res`, and it doesn't seem clear that editing `refresh_row` edits the value in `res`. I would edit `res` directly instead.

robintw (changes requested):
Looks good, apart from a minor Python issue (the code there probably works, but is a bit confusing).

pepys_timeline/db.py
krsnik93: @IanMayo Very good point! Great moment to add logging to the app. I have set it so that it logs to stdout, which means it will be easily accesible in Heroku logs. If we ever wish to log to a file, the change will be simple.

pepys_timeline/db.py
krsnik93: @robintw good point, I refactored this.

pepys_timeline/db.py
IanMayo: Thanks @krsnik93 

Yes - I recognise the value of Heroku logs.

But, my users may not know how to access stdout logging.  If any of the connection/database errors happen the impact is catastrophic - it just won't work.  So, in my opinion it's fine to show the errors there in the browser-page. Then the users can most quickly/effectively pass the error onto me (or their tech support), and we can get it fixed.  

So, can we show the errors in the browser page please?

krsnik93: @IanMayo Gotcha, sorry for the earlier confusion. I have made a change to show the whole exception traceback in the error field in the GUI.

krsnik93: image

IanMayo: Thanks. Minor improvement, please @krsnik93 

Could we have `Error connecting to the database` as `h3` (or something like that) then the traceback as `normal` (or something like that) please?

This will inform the users what has happened in general terms, then offer more detail for follow-up.

krsnik93: @IanMayo Good point. Let me know if the paragraph does not look visible enough.

image

robintw: Could you translate the newlines `\n` in the Python string into newlines (`<br>` or something) in the HTML, so that the traceback displays properly? Thanks!

IanMayo: Aah, this _may_ introduce the need for scrollbars in the new `div`.

krsnik93: Good point, sorry about that.

image

@IanMayo It does overflow on my screen and shows a scrollbar, would this be an issue? We could try reducing font or clipping or maybe shortening the message...

IanMayo: Thanks @krsnik93  the scroll-bar is acceptable.

IanMayo (approved):

krsnik93:
@IanMayo have you made any changes to Codacy? It's passing now...

IanMayo:
> @IanMayo have you made any changes to Codacy? It's passing now...

I've updated the branch, but I doubt that's the issue. Maybe they have sorted their issue.

PR #918, opened 2021-05-18
Remove pagination restrictions if pagination parameters are not provided

### LINKED ISSUE #914: Remove default page size in stored functions
The stored functions have a default page size of 100 records.

So, if a page size isn't provided, only 100 records are returned.

I _think_ this is counter-intuitive.  If a page size isn't provided, then I think all records should be returned, without paging.

I've just run this import, which should import `104121` records. It returned with `imported 200 items`.  I guess that's 100 per tick-box:
image

Paging the results could certainly help provide a responsive UI in the future, but we need to handle basic querying first.

I guess this will need the collaboration of both @arunlal-v  and @robintw 

Note: Issue #913 may also introduce the need to modify the stored procedure, so we should probably combine both updates into one database migration.

arunlal-v:
Remove pagination restrictions if pagination parameters are not provided.
Fixes #914 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


robintw:
Migration added - I think this is ready to go now.

(Note: the Create Deployment CI task will fail due to the branch name having a / in it)

IanMayo (approved):
Great, thanks Robin.

Issue #923, opened 2021-07-13
Potential migration issue


IanMayo:
The Debrief dialog used to import Pepys data was failing, so I opened Pepys-Admin to check if the PostGres database was running.

Pepys-Admin opened, and displayed the valid connection details.  When I tried to use `View data` to examine the data, it reported `Database tables not found. Do you need to initialise?` (or something like that).

A few minutes later I opened Pepys-Admin again, again it displayed the valid connection details. But, this time when I tried to view data, it said `Please run database migration to bring tables up to date`.  When I tried to run the migration, it reported:  `Exception details: Requested revision 00116b58f22e overlaps with other requested revisions ff4215cb3d30`.

I used PGAdmin to opened the `alembic_versions` table, and it contained the above two rows.

Note: the migration script log also included `len(table_names)=42`.

I see the two above migration ids in the `migrations` folder, though they are from 3 and 7 months ago.

robintw:
Does this need dealing with now @IanMayo, or can it wait until we get the contract sorted? I'm still recovering from Covid, but I might be able to do a bit on it soon if needed.

IanMayo:
Have replied in Slack.

robintw:
For reference, the two migrations listed above are:

`00116b58f22e` from 1st Dec 2020, which adds the HelpTexts table
`ff4215cb3d30` from 26th April 2021, which adds a unique constraint for the Serials table

These are both listed as the current version in the `alembic_version` table, and therefore the migration procedure doesn't know what to do.

The only migrations which come _after_ `ff42...` are:
- Adding the `ConfigOptions` table
- Adding stored procedures for dashboard
- Adding stored procedures for Debrief (potentially a reason why the Debrief import isn't working @IanMayo?)
- Fixing pagination in the Debrief procedures

My suspicion at this stage is that `ff42...` is the _correct_ version that should be in the `alembic_version` table, and the `0011...` version was added by an out-of-date Pepys version that someone ran. Therefore, the way to fix this is to remove the row containing `0011...` and then run the migration again.

However, before doing that I want to check that the database is actually at the `ff42` version. To do this we can check that:
- A constraint called `uq_Serial_serial_number_wargame_id` exists on the `Serials` table.
- The `ConfigOptions` does _not_ exist

So, @IanMayo - do you think we can get them to check that using pgAdmin?

[...]

PR #940, opened 2021-08-12
Initial fix/prevention measure to stop older version of Pepys modifying current databases.

### LINKED ISSUE #939: Prevention of older versions of Pepys interacting with newer version of the database.
## 🐞 Overview

We have experienced issues where old versions of Pepys connect to a newer version of the database, and this causes various problems (including the old version of Pepys messing up the internal version of the database). Therefore, we should add code to Pepys that stops it connecting to the database if the database version (as stored in the alembic_versiontable) is a version that is newer than any that that version of Pepys knows about (ie. the current database version is not a 'known version' for this version of Pepys).

SamWilliams-dev:
## 🧰 Issue

Fixes #939 
Supports #923

## 🚀 Overview: 

Error handling and checks to ensure that future database migrations cannot happen if the user is using an older version of Pepys. If found to be using an older version of Pepys without any reference to the current version of the database, then the initialisation of the Data_Store will fail and the user will be advised to update to the latest version of Pepys.

## 🤔 Reason: 

Prevents users from being able to use an older version of Pepys to update or modify a database using a more up-to-date revision of the database. If the database that Pepys is trying to access is not recognised from a list of known PostGres and SQLite revisions then Pepys will no longer allow modifications and will instead exit.

- [x] Tests pass

## Confirmations


pepys_import/core/store/data_store.py
```diff
@@ -44,6 +46,22 @@
 DEFAULT_DATA_PATH = os.path.join(PEPYS_IMPORT_DIRECTORY, "database", "default_data")
 USER = getuser()  # Login name of the current user
 
+# Python Set Object - is this item in this set
+
+# Constant Lists of revisions - list all files in the directory
```
IanMayo, line 51: on first impressions it looks like we have `postgres_versions` and `sqlite_versions` directory names the wrong way around @SamWilliams-dev

pepys_import/core/store/data_store.py
```diff
@@ -44,6 +45,23 @@
 DEFAULT_DATA_PATH = os.path.join(PEPYS_IMPORT_DIRECTORY, "database", "default_data")
 USER = getuser()  # Login name of the current user
 
+# Constant Lists of revisions - list all files in the directory
+SQLITE_REVISIONS_FOLDER = os.path.join(MIGRATIONS_DIRECTORY, "sqlite_versions")
+SQLITE_REVISION_LIST = os.listdir(SQLITE_REVISIONS_FOLDER)
+SQLITE_REVISION_IDS = []
+if len(SQLITE_REVISION_LIST) > 0:
```
robintw, line 52: No need for this in Python - if there is nothing in the list then the for loop below just won't run at all
(Same applies fo the Postgres equivalent on L60)

pepys_import/core/store/data_store.py
```diff
@@ -44,6 +45,23 @@
...
+    for sqlite_revision in SQLITE_REVISION_LIST:
```
robintw, line 53: This loop is fine, but could probably be done more 'pythonicly' with a list comprehension. Have a google to learn about them - as you'll find them elsewhere in the codebase - and then have a go at rewriting as a list comprehension

pepys_import/core/store/data_store.py
```diff
@@ -44,6 +45,23 @@
 DEFAULT_DATA_PATH = os.path.join(PEPYS_IMPORT_DIRECTORY, "database", "default_data")
 USER = getuser()  # Login name of the current user
 
+# Constant Lists of revisions - list all files in the directory
```
robintw, line 48: Could you add a more detailed comment here - saying that this is a list of revisions known to this version of Pepys, and by putting it in the top of the file (rather than in a function), it is computed once at file load, thus making the DataStore constructor more efficient

pepys_import/core/store/data_store.py
```diff
@@ -89,8 +107,12 @@ def __init__(
             driver, db_username, db_password, db_host, db_port, db_name
         )
         try:
+            # Create engine to be used by both branches of if statement
```
robintw, line 110: I think this comment is out of date

tests/test_datastore_utils.py
```diff
@@ -98,6 +100,84 @@ def test_create_alembic_version_multiple_rows():
...
+    ds = DataStore("", "", "", 0, ":memory:", db_type="sqlite")
+    revisions = []
+
+    with ds.engine.connect():
+        with pytest.raises(SystemExit) as exit_exception_e:
+            temp_output = StringIO()
+            with redirect_stdout(temp_output):
+                ds.check_migration_version(revisions)
+        output = temp_output.getvalue()
+        assert "ERROR: Expected list of known revisions is 0." in output
```
robintw, line 132: Thinking about it, this error message isn't quite right. How about `...is empty` rather than `...is 0`.

robintw: (I'm aware that this will need changing in the main function too, as well as here)

tests/test_datastore_utils.py
```diff
@@ -98,6 +100,84 @@ def test_create_alembic_version_multiple_rows():
...
+        assert exit_exception_e.value.code == 1
+
+
+def test_check_migration_version_not_in_revisions():
+    # Call the function with no revisions in the list - function should call sys.exit(1)
```
robintw, line 137: Outdated comment

robintw (dismissed):
Here are some comments on the PR. In general it's doing the right thing, but there are a few loose ends to tie up, a few comments that are out of date, and a couple of non-Pythonic things.

IanMayo:
@SamWilliams-dev - the static code analysis tool warned about this issue in your PR:
image

Could you have a look in the codebase and see if/how we avoid SQL injection issues elsewhere?

IanMayo:
One last thing @SamWilliams-dev - code-cov is reporting a couple of lines that aren't covered by tests. They're actually just one branch of logic, so shouldn't be too hard to cover with an extra test:
image

IanMayo:
Oh dear, I may be about to look silly @SamWilliams-dev .

image

In the above code we're retrieving the list of alembic versions. There should only be one - the _current_ version.  But, then we look at the contents of the first entry, and check it is one character long.  But the alembic versions are really long guids.  Why should we want to check it's one character long?

IanMayo:
Sorry @SamWilliams-dev - static checker found a fault in the new list-related type checking:
image

.pre-commit-config.yaml
```diff
@@ -3,7 +3,7 @@ repos:
     rev: 20.8b1
     hooks:
       - id: black
-        language_version: python3.7
+        language_version: python3
   - repo: https://github.com/asottile/seed-isort-config
```
IanMayo, line 7: @SamWilliams-dev - do you know why the python version constraint was relaxed?

IanMayo: Ok, @SamWilliams-dev  replied to indicate he needed to relax the constraint since he's running 3.9.x on his dev machine.

3.9 is the currently supported version of Python - so there is merit in adopting it. Since we deploy python in the `Pepys` distributable we could upgrade with minimal impact to users.

IanMayo (approved):
PR approved, subject to us solving the python version in git-hook issue elsewhere.

PR #950, opened 2021-08-13
Store Pepys version and build date (if available) in the Changes.reason field when running import

### LINKED ISSUE #949: Report Build version in extraction log
When we import a file, we log it in the `Changes` log.

In this report, we should include the version and build-date of the Pepys instance in use.  This will allow us to reconstruct the import operation, if necessary.

[Request from CF]

robintw:
## 🧰 Issue
Fixes #949 

## 🚀 Overview: 
Extends the import process to store the version of Pepys and the build date (if available - it's only available for releases) in the `reason` field of the `Changes` entry created for the import. The `reason` will now look like:

```
Importing 'rep_test1.rep' using Pepys 0.0.25, built on 2021-01-01 17:00:00
```

## 🤔 Reason: 
Gives a better audit trail to determine which files were imported with which versions.

## 🔨Work carried out:

- [x] Altered FileProcessor to change reason
- [x] Added tests
- [x] Tests pass

## Confirmations


tests/test_importers.py
```diff
@@ -1033,5 +1035,38 @@ def test_skip_file(self, resolver_prompt, menu_prompt, file_processor_prompt):
             assert len(self.store.session.query(self.store.db_classes.Comment).all()) == comments
 
 
+def test_changes_table_entry_contains_version():
```
mew-nsc, line 1038: Test for version looks good Robin. Would it be possible to have another test to check `__build_timestamp__` being applied?

mew-nsc (approved):

tests/test_importers.py
robintw: Unfortunately that would be fairly difficult - as the build timestamp is only set when we do a full release. That means that when we're running the tests on the development version, `__build_timestamp__` will always be None.

IanMayo: Aah, as it happens I'm going to make a new release today -  so I can offer a manual test to verify it.

IanMayo (approved):

tests/test_importers.py
mew-nsc: No problem Robin. If you're happy to test as part of your release Ian, that would be great.

PR #956, opened 2021-08-16
Treat warnings as errors in all pytest runs on CI

### LINKED ISSUE #955: Treat warnings as errors when running tests on CI
## 🐞 Overview
At the moment, warnings are printed out when running pytest, but don't cause an error - so would allow the CI to pass. We don't want any warnings, so we should use the appropriate options to pytest to convert warnings to errors, and therefore make the tests fail if there are any warnings.

This is particularly useful for me when testing the new SQLAlchemy version, as I need to check for deprecation warnings, but it is something we should do anyway as it will be useful elsewhere.

robintw:
## 🧰 Issue
Fixes #955 

## 🚀 Overview: 
Adjusts the CI configuration to treat all warnings during pytest runs as errors.

## 🤔 Reason: 
Makes sure we will be aware of warnings.

## 🔨Work carried out:

- [x] Update CI configuration
- [x] Tests pass

## Confirmations

### DIFF
--- a/.github/workflows/test_pepys.yml
+++ b/.github/workflows/test_pepys.yml
@@ -42,14 +42,14 @@ jobs:
         env:
           PEPYS_SPATIALITE_PATH: /usr/lib/x86_64-linux-gnu/mod_spatialite.so
         run: |
-          coverage3 run -m pytest -v --color=yes --ignore=tests/gui_tests tests/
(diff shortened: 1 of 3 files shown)


IanMayo (approved):
Nice improvement, thanks.

PR #973, opened 2021-09-01
Disabling old versions of Pepys from connection to database

### LINKED ISSUE #964: Disable old version of Pepys from connecting to database
## 🐞 Overview
An old version of Pepys somewhere on the client network is causing significant problems for us and the clients.

We want to disable this old version, without being able to modify its code in any way.

One way to do this would be to change the code that parses the environment variable into the path for the Pepys config file. At the moment, the environment variable is set like this: 

```PEPYS_CONFIG_FILE=/Users/robin/Documents/IanMayo/pepys_config.ini```

We could change it to something like:

```PEPYS_CONFIG_FILE=PEPYS/Users/robin/Documents/IanMayo/pepys_config.ini```

and then alter the code that reads the variable to strip the `PEPYS` off the beginning. The old code wouldn't know to do this, and would therefore try to read a file with that name, and give an error like:

```
Pepys config file not found at location: 'PEPYS/Users/robin/Documents/IanMayo/pepys_config.ini'.
```

(The `PEPYS` string could be any string that we could stick at the front of the path to make it invalid.)

Part of me thinks this is a bit of a 'yucky' solution, but it would definitely work, and would stop the old version of Pepys in its tracks.

To make this sort of problem easier to solve in future, we could create a table in the database called something like `InvalidClientVersions`, which Pepys checks on startup, and if it finds its version number listed in there then it stops and gives an error. This gives us a 'kill switch' that would allow us to disable specific versions of Pepys by simply editing the database (either in pgAdmin, or through an interface we could provide in Pepys Admin).

What do you think @IanMayo?

### PR DESCRIPTION
**Pull request by SamWilliams-dev**
## 🧰 Issue

Fixes #964 

## 🚀 Overview: 

Changes to `default_config.ini` and `config.py` variables to now be _database_ rather than _db_. 
Older versions should no longer connect as variable names will not match and will instead throw an error.

## 🤔 Reason: 

Preventative measure to stop older versions of Pepys from interacting with current databases.

- [x] Tests pass

## Confirmations

### DIFF
--- a/config.py
+++ b/config.py
@@ -31,7 +31,7 @@ if not config.has_section("database"):
 if not config.has_option("database", "database_type"):
     # Config file is likely an older version
     print(
-        "Config file contains incorrect variable names. Please update your configuration to the correct format."
+        "Config file contains variable names used in legacy versions of Pepys that provided insufficient support for database migration. \n Please contact your system administrator to discuss renaming db_xxx to database_xxx."
     )
     sys.exit(1)
 
--- a/config.py
+++ b/config.py
@@ -28,12 +28,12 @@ if not config.has_section("database"):
     sys.exit(1)
 
 # Fetch database section
-DB_USERNAME = config.get("database", "db_username", fallback="")
-DB_PASSWORD = config.get("database", "db_password", fallback="")
-DB_HOST = config.get("database", "db_host", fallback="")
-DB_PORT = config.getint("database", "db_port", fallback=0)
-DB_NAME = config.get("database", "db_name")
-DB_TYPE = config.get("database", "db_type")
+DB_USERNAME = config.get("database", "database_username", fallback="")
+DB_PASSWORD = config.get("database", "database_password", fallback="")
+DB_HOST = config.get("database", "database_host", fallback="")
+DB_PORT = config.getint("database", "database_port", fallback=0)
+DB_NAME = config.get("database", "database_name")
+DB_TYPE = config.get("database", "database_type")
 
 # Process username and password if necessary
 if DB_USERNAME.startswith("_") and DB_USERNAME.endswith("_"):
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Comment by SamWilliams-dev** [comment]
@IanMayo 
For the static code analysis, I'm not entirely sure why it's flagging it as an error. The function follows the same structure as all of the other tests as well as passing in (self) as a parameter which by the explanation of the docs you linked, should've been enough to not flag it. I'm not sure that adding @classmethod or @staticmethod would be the correct approach as none of the other tests do and I don't think the test should be static. Do you have any advice?

File: config.py [hunk 1]
```diff
@@ -27,13 +27,21 @@
     print(f"Couldn't find 'database' section in '{CONFIG_FILE_PATH}'!")
     sys.exit(1)
 
+# Error if database type is not found
+if not config.has_option("database", "database_type"):
+    # Config file is likely an older version
+    print(
+        "Config file contains incorrect variable names. Please update your configuration to the correct format."
```
robintw, line 34: @IanMayo Do we want a more specific message here that warns the user that they may be using an out-dated version of Pepys and to talk to their system administrator?

robintw (commented):
Looks good Sam. Just waiting for @IanMayo's feedback on the error message, then I'd say this is ready to be merged.

config.py
IanMayo: Yes, this is a _special_ error that I'm happy to provide additional guidance to.

Let's go for:
`Config file contains variable names used in legacy versions of Pepys that provided insufficient support for database migration.  Please contact your system administrator to discuss renaming `db_xxx` to `database_xxx`.

robintw (approved):
Looks good now, thanks Sam.

IanMayo (approved):
Tested by inspection.

Performance benchmark failing very slightly, but not concerned about that.

Issue #927, opened 2021-08-11
Investigate technology to automate running System Test Plan


IanMayo:
The project is going to benefit from a System Test Plan (STP) #930  - a sequence of physical interaction of the built application (potentially a release candidate) that very end-to-end performance.

Best practice will include modifying the test plan as code gets modified.

Writing/maintaining the STP is seen (_by Ian_) as a costly (but beneficial) undertaking. We'll get the most benefit of the investment if we can (optionally) automate running the STP.  Since `pepys-import` and `pepys-admin` are command-line apps, it shouldn't be too complex to reproduce the keyboard/mouse interactions. Ideally we'd be doing it an operating-system agnostic fashion that allows windows/mac devs to run the test, but also for GitHub to run it as part of every PR.

Note: there's an `delay cost` in this issue.  If we identify the technology early in the development of the STP then the plan can be written accordingly.  The longer the delay, the more work involved in porting the steps.

IanMayo:
Greater benefits from earlier adoption, if we are going automate it.

robintw:
I've been thinking about the various parts of the system that we want to test, and how we could automate them. Notes below:

## Import and Admin UIs
We want to test that we can run through the import/admin CLIs with specific interactions (press `2` here, select this option etc) and that the task completes successfully. This is partly tested already in the end-to-end CLI tests, which provide certain input to the UI and check that the import completes. There is also a couple of 'automaton' tests that try providing all sorts of valid and invalid input to the UI and check that it doesn't crash. However, all of these tests use the Pepys code to automate the tests - that is, they patch various functions and get them to return specific values, to simulate the user typing them in. There is currently no test that is _independent_ of the Pepys code itself.

Through using the `pexpect` module, we can 'drive' the Pepys UIs from a completely separate program, without using any of the Pepys API - it works exactly as if the user were typing the input themselves. A proof-of-concept of this sort of test is in #976.

_Note:_ This will not work properly for the Maintenance/Tasks GUIs - or at least, it will only work in a very superficial manner. I'll investigate this more.

## Installation
The installation scripts for Pepys are tested as part of the 'create deployment' CI job - this creates a deployable release, then installs it, and then runs a simple import. If any of these fail then the job fails.

## Windows UI - shortcuts etc
Shortcuts created in the Start Menu and the Send To menu are not currently tested. I am investigating possible ways of automating these tests, including AutoHotKey, various Python GUI automation frameworks, and so on.

## Maintenance/Tasks GUI
[...]

PR #976, opened 2021-09-07
Add an end-to-end system test using pexpect

robintw:
## 🧰 Issue
Contributes to fixing #927

## 🚀 Overview: 
Adds a test for doing an end-to-end test of the import CLI using the `pexpect` module (based on the Unix `expect` program). This is a way of running a program (in this case, the import CLI) and sending keypresses to it, and 'expecting' certain output from it. By combining these 'sends' and 'expects', we can automate the interaction with the UI. For example, code like this:

```
child.expect_exact("> ")
child.sendline("2")
```

expects a prompt from the program, and responds with `2` (followed by Enter), and 

```
child.expect_exact("Files got processed: 1 times")
```
expects to see that output from the program.

Currently this does not run on Windows, so there is code to skip this test on Windows.

## 🤔 Reason: 
Better testing

## 🔨Work carried out:

- [x] Add pexpect-based test
- [x] Tests pass

## Confirmations


robintw:
@IanMayo You weren't in standup when I discussed this earlier. Do you think this is a sensible approach? If so, I'll go ahead with trying to get it working on Windows.

IanMayo:
> @IanMayo You weren't in standup when I discussed this earlier. Do you think this is a sensible approach? If so, I'll go ahead with trying to get it working on Windows.

Ooh, it's a shame I missed the discussion.

This does tick the box of running Pepys _from the outside_, , but I _imagined_ something that was closer to plain English.  Then _roughly_ the same content could be used manually by a person, or automatically as part of CI.

I also think we need to consider where we invest effort in the System Tests.  If the greatest value comes from testing Pepys Admin (it's heavy GUI nature means it has less unit testing), but `pexpect` can't help us there, then I think we should consider another strategy.

Maybe we should discuss it with @mew-nsc & @SamWilliams-dev  either in the next standup, or a specific call.

IanMayo:
I think it may be worth discussing this at the next standup.  We do an import in the System Test.  I have a minor worry that this may be a bit fragile.  Hmm, but if it means the full import cycle is covered in every set of unit tests, that could be a positive benefit.

robintw:
Yes we can chat in standup. My personal view is that it's worth including at the moment, and if it proves to be really fragile then it can be removed in future. It does do a full independent end-to-end check which we don't have elsewhere - so I think it could be useful.

IanMayo (approved):
Yes - this is a valuable end-to-end test that would bring value to our automated unit tests by being run so regularly/frequently.

PR #977, opened 2021-09-07
CSV Exporter

### LINKED ISSUE #826: Export to CSV format
## 🐞 Overview

Analysts have a range of tools available, and may wish to apply analysis techniques to Pepys data.

One easy way of supporting this is via export to CSV.

We can do this by letting the analyst choose:
- data source
- columns to be present in the export
- rows to export (or select-all)

Then:
- Select `Export to CSV`
- Provide path\file (default is `~\Documents\yymmdd-hhmmss.csv`
- Put field names into header (with units when relevant)
- one row per row of data
- split location across two cols (degrees, 5 decimal places)

## 🔗 Feature

#600

SamWilliams-dev:
## 🧰 Issue

Fixes #826

## 🚀 Overview: 
Adds a new action to the Maintenance GUI, to export the selected entries to a CSV. Implemented as a backend method on the datastore to do the export, plus a GUI dialog to configure and run the export.

## 🤔 Reason: 
The user requires the ability to export a list of chosen entities to a CSV file.

## 🔨Work carried out:

- [x] Wrote backend method, including ability to add extra columns (eg. to separate location into lat and lon columns), add units in headers etc
- [x] Added GUI dialog box
- [x] Connected GUI and backend
- [x] Added extra tests
- [x] Tests pass

## Confirmations


robintw:
@IanMayo A few questions about how exactly you'd like the CSV export to be formatted.

Currently, an example of what this produces is this:
image

1. Are you happy with units in brackets in the column names?
2. Do you want column names to use the 'real' column names with _'s in them, or replace those with spaces?
3. In situations when a field doesn't have a value - eg. the whole course column here - do you want the field just left empty (so it is `,,` in the CSV), or filled with something like `NULL` or `None`?

IanMayo:
> @IanMayo A few questions about how exactly you'd like the CSV export to be formatted.
> 
> 1. Are you happy with units in brackets in the column names?
Yes, great.

> 2. Do you want column names to use the 'real' column names with _'s in them, or replace those with spaces?
I don't think there's a _right_ answer. I'm ok with the underscore

> 3. In situations when a field doesn't have a value - eg. the whole course column here - do you want the field just left empty (so it is `,,` in the CSV), or filled with something like `NULL` or `None`?
Do you have a preference?  I think I'm ok with them being empty.

robintw:
Great, I'm fine with all of that.

robintw:
Here's a quick screencast showing exporting from the GUI, and looking at a few of the CSV files afterwards.
ExportToCSV_Demo


Here is a screenshot of part of the States CSV, rendered nicely:
image

IanMayo:
Thanks @robintw - I don't need to see it, but could you do a test of exporting some really high volume dataset (such as `States in 2020`) and reporting if it works correctly?  It's just so we're aware of any bulk data issues such as memory or something timing out.

robintw:
@IanMayo Thanks for reminding me of that. I've added a proper progress bar now and tested it on a large export. An export of ~100,000 rows takes about 2mins. It was a lot slower than that, but I did a bit of profiling and speeded it up significantly.

The only tests that are failing at the moment are the benchmarking tests: this doesn't affect the import code, so it must just be that Github Actions is slow at the moment. It can be merged with those benchmarks failing.

robintw:
Codacy is showing a few issues, but I think they can all be ignored. Some of them are actually required (some functions in prompt_toolkit require certain parameters, even if they aren't used), and some we can't do much about (like our function being fairly complex, or taking a lot of parameters).

IanMayo:
I agree - the Codacy errors aren't very helpful.  I think the `else after return` construct you've used is clearer than omitting the `else`.

PR #984, opened 2021-09-13
Fix Codacy errors by giving Prospector configuration

### LINKED ISSUE #974: Deal with Codacy/pylint warnings about no-self-use in tests
## 🐞 Overview
The Codacy static analysis tool is giving lots of warnings about methods that could be functions. This occurs when you have a method defined like this:

```
class Blah:
    def method_here(self):
         # Code here
         # which never accesses self
```

The analysis warns that your `method_here` could actually be a standalone function, as it never actually accesses anything on the class. In general this is a pretty good warning, but it is being triggered a lot by some of our old-style tests in the Pepys codebase.

The old-style way of writing Python tests (using the `unittest` framework, which is based on JUnit from Java) is to write a test class which inherits from `unittest.TestCase` and then have all the tests as methods within that class. If this is a test that doesn't require any sort of setup or teardown, then the methods inside the test class will never access `self` - and thus will trigger this warning. The new-style way of writing Python tests is to write them as standalone functions, and only use test classes when you need particular setup or teardown (like creating/deleting a database at the start/end of the test).

The cleanest way to deal with this would be to switch all tests that don't _require_ a class to be standalone functions, but that will be a lot of boring work across the codebase and probably isn't a high priority.

Another approach would be to disable this warning. The underlying tool that the warning comes from is pylint, so you can use a pylint config file to turn off the `no-self-use` warning. However, this would disable them everywhere. We could put comments in the test functions to disable them for that test, but that would require littering the codebase with comments. A better option would be to put comments at the top of all of the test files that use old-style tests to disable the warning for that file - something like `# pylint: disable=no-self-use` at the top of the file should work.

The other alternative is to do nothing and just ignore the warnings if they come from a test file.

What do you think is best? I think I'd lean towards putting comments at the top of all of the old-style test files.

### LINKED ISSUE #978: Configure Codacy errors sensibly
## 🐞 Overview
We currently have some conflicting errors being raised in Codacy, through the prospector tool that they use. [This](https://stackoverflow.com/questions/45990301/pep257-d212-and-d213-conflicts) describes the issue, and we should be able to create a config file containing one of the [default profiles](https://github.com/PyCQA/prospector/tree/master/prospector/profiles/profiles) to fix it.

robintw:
## 🧰 Issue
Fixes #978
Fixes #974

## 🚀 Overview: 
Configures the Prospector tool (as used by Codacy) to remove unhelpful checks which lead to conflicting errors. By default it seems Codacy does not use the 'default' Prospector config, so we have copied it to a `.prospector.yml` file.

## 🤔 Reason: 
Get rid of conflicting and unhelpful errors.

## 🔨Work carried out:

- [x] Added prospector config
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
This is hard to test as Codacy only gives warnings for code lines/files that have changed. I suggest we assume this works and merge it, and then test it as part of the next set of PRs that we do (eg. the Export to CSV PR).

### DIFF
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -1,0 +1,5 @@
+strictness: medium
+doc-warnings: false
+test-warnings: false
+autodetect: true
+member-warnings: false


IanMayo:
@robintw - are we adopting this file to configure how our local tests are run (to match Codacy online), or to change Codacy to match how our local tests are run?

robintw:
Codacy runs some tests that we don't run locally - for example, it uses pylint which we don't run locally (we run various other things including flake8). This PR just forces Codacy to use the 'default' prospector config, which turns off a load of unhelpful warnings, and disables the conflicting warnings that we got before about docstrings.

I don't think we need to run all of these things locally - I think using Codacy as a rough guide that we can choose to ignore sometimes is fine.

IanMayo (approved):

PR #985, opened 2021-09-13
Add headers and file process counter during import

### LINKED ISSUE #669: Progress counter through series of files.
I'm loading a folder of data-files, using the default resolver.

It would be useful to see how far through the process we are.  Here is a full processing of a datafile (with defaults resolver):
image

When we're running cli resolver, there will be the step when a user processes the next data-file.  So in interactive mode, _maybe_ we should highlight the title block when we start a new data-file.

Aahh, hold on. I should re-consider this once we've got the context sensitive help in - since that's where these title blocks get introduced.

robintw:
## 🧰 Issue
Fixes #669 

## 🚀 Overview: 
Added a header every time a new file is processed, along with a counter showing the number of the file in the list of files being processed.

image

image

## 🤔 Reason: 
Give the user more status feedback

## 🔨Work carried out:
- [x] Switch to creating list of files first, and then iterating through it - so we know how long it is
- [x] Print header with counter each time a file is about to be imported
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
Sometimes the number of files to be processed will be very large, if there are lots of files in the folder that aren't recognised by Pepys. In this case, the counter will only increase for files that Pepys actually processes - so it could finish after processing file 8 of 100 - because the other 92 are ones Pepys doesn't recognise. Although this might be a bit weird from the user's perspective, it's the best we can do without major rearchitecting.


robintw:
@IanMayo I'd welcome your confirmation that the way I've done this is ok - particularly the bits in the Developers Notes about how we might get a slightly confusing output shown.

IanMayo:
Sure @robintw - that explanation is fine.  Maybe also include it in a comment near where the message is written to the console - to help out our future selves.

IanMayo:
Aah, I've just thought of a fix for the potential confusion over "Importing file 3 of 123".  Would it be more transparent if we used "Processing file 3 of 123"?  I guess "Processing" applies to files that we do and don't import.  _Maybe_ "Inspecting" would be an alternative.

robintw:
Ah excellent idea @IanMayo - I'll change it to 'Processing'

robintw:
Updated now @IanMayo. For some reason a separate commit isn't showing in the PR, but if you check the 'file changes' tab it shows it.

IanMayo (approved):

PR #986, opened 2021-09-13
Fix View Data CLI bugs and add more tests

### LINKED ISSUE #963: View Data CLI cannot view specific tables
## Version
develop

## 🐞 Description
The View Data CLI interface is unable to view a number of tables, including `Serials` and `Wargames` - and potentially others.

## ⚙️ Current behavior
The View Data CLI gives the following error when trying to view `Serials` or `Wargames`:

```
Attribute error! Please check the error message: 'Wargame' object has no attribute 'participants_platform_name'
```

## ⚙️ Expected behavior
Rather than an error message, a view of the table data is displayed.

## 🔢 Steps to Reproduce
1. Run Pepys Admin
2. Select View Data
3. Choose `Serials` or `Wargames`

## Notes on how to fix
This is caused by the View Data CLI assuming a certain naming pattern for association proxies of `"{descriptor.target_collection}_{descriptor.value_attr}"`. Ideally we could extract the name automatically, but I'm pretty sure this wasn't possible when I investigated it. Instead we should:
- Rename the association proxies so they match the expected name (and rename their use everywhere else in the code too)
- Alter the code in view_data_cli.py to check whether the attribute exists once we've built the name - by adding some code around line 134
- Add some error-handling to the data extraction code around line 147, so that if an attribute doesn't exist or can't be read, the code doesn't crash (but prints a message saying `XYZ attribute could not be accessed`
- Add more testing of the functions for viewing data - specifically, we should call `do_view_table` with all possible tables as options, and check that each of these leads to a successful display (without any of the `XYZ attribute could not be accessed` messages)

robintw:
## 🧰 Issue
Fixes #963 

## 🚀 Overview: 
Fixed bug where viewing some tables in the View Data CLI led to an attribute error. Refactored a bit, and added more tests to ensure that all tables can be viewed.

## 🤔 Reason: 
Fix bug and improve testing

## 🔨Work carried out:

- [x] Fix View Data CLI to get association proxies correctly, and thus not generate attribute errors.
- [x] Fix names of some association proxies to better reflect normal naming convention
- [x] Refactor function to view table to extract code that generates text
- [x] Add test to check that text can be generated for each table
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
The View Data CLI code that dealt with association proxies could be replaced with a call to one of the functions written for the GUI - the GUI didn't exist when this CLI code was first written, and so we were 'doing it the hard way'. This way is much more robust, and shouldn't lead to errors. We also add tests to ensure that all tables can be viewed without errors.

### DIFF
--- a/pepys_admin/view_data_cli.py
+++ b/pepys_admin/view_data_cli.py
@@ -139,6 +139,8 @@ class ViewDataShell(BaseShell):
                 headers.append(column.key)
         # Get the association proxy attributes for this table
         associated_attributes, _ = get_assoc_proxy_names_and_objects(table_cls)
+        if "privacy_name" in associated_attributes:
+            associated_attributes.remove("privacy_name")
 
         header_attributes = [getattr(table_cls, header_name) for header_name in headers]
         # Fetch first rows up to MAX_ROWS_DISPLAYED, create a table from these rows
--- a/pepys_admin/view_data_cli.py
+++ b/pepys_admin/view_data_cli.py
@@ -7,12 +7,12 @@ from prompt_toolkit.lexers import PygmentsLexer
 from pygments.lexers.sql import SqlLexer
 from sqlalchemy import inspect
 from sqlalchemy.exc import InvalidRequestError, OperationalError, ProgrammingError
-from sqlalchemy.ext.associationproxy import AssociationProxy
 from sqlalchemy.orm import RelationshipProperty, class_mapper, load_only
 from sqlalchemy.sql.expression import text
 from tabulate import tabulate
 
 from pepys_admin.base_cli import BaseShell
+from pepys_admin.maintenance.column_data import get_assoc_proxy_names_and_objects
 from pepys_admin.utils import get_default_export_folder
 from pepys_import.core.store import constants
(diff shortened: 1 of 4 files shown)


pepys_admin/view_data_cli.py
```diff
@@ -87,13 +90,21 @@ def do_view_table(self):
...
-        headers = list()
-        associated_attributes = list()
         table_names = self._get_table_names()
         message = "Select a table >"
         selected_table = iterfzf(table_names, prompt=message)
         if selected_table is None:
             return
+
+        text = self.generate_view_table_text(selected_table)
+        print(text)
```
IanMayo, line 100: @robintw - is the above a debug line?

robintw: Not really - it's in there to ensure that if we get a failure of the test, we actually know which table it failed on. Ideally we'd use a 'parameterised' pytest but we can't do that as we need to generate the list of tables based on access to the DataStore object.

It could be removed, but then whenever the test failed we'd have no idea what table it failed on without adding this line back in.

IanMayo: Sure, ok.

IanMayo (approved):

PR #987, opened 2021-09-13
Add extra test for check_migration_version

### LINKED ISSUE #966: Further tests required for the check_migration_version function of data_store.py
## 🐞 Overview

The `check_migration_version()` function within `data_store.y` has a part of the logic that is currently not covered by any of the tests and therefore flags up within the code coverage. This part of logic is expected to receive the alembic version and check that the format is a tuple list and that the received data is of the correct, expected length. If so, a check is in place to ensure that the first element of the tuple has a length of one but as of writing this, there is no test to ensure this works or whether the code here will ever trigger. 

If the received alembic version is not in a format that we expect, we cannot do a comparison with our known revisions and therefore Pepys should exit out of the program as detailed below. Currently the first if statement can be triggered if only one tuple is received but the second never triggers on any received data, as if the first element of the tuple always has a length of 1.

All other tests regarding this function pass.

This issue was identified in PR #940 , [here](https://github.com/debrief/pepys-import/pull/940#issuecomment-902539701).
              
```
  if len(table_contents) == 1 and type(table_contents) == list: 
         if len(table_contents[0]) != 1:
                # Content has been found but it is not the correct length
                print(
                    "ERROR: Retrieved version contents from database is incorrect length. \n"
                    "Cannot correctly compare with currently known migrations. Please check with your administrator."
                )
                sys.exit(1)
```

robintw:
## 🧰 Issue
Fixes #966 

## 🚀 Overview: 
Added extra test for check_migration_version to improve coverage.

## 🤔 Reason: 
Improve test coverage

## 🔨Work carried out:

- [x] Add test
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
The only bit of the function that wasn't tested was dealing with an incorrect number of columns returned from the database. This is unlikely to happen in real-life, but can be simulated by creating the alembic_version table with more than one column. That's what this test does.

### DIFF
--- a/tests/test_datastore_utils.py
+++ b/tests/test_datastore_utils.py
@@ -200,6 +200,23 @@ def test_check_migration_version_incorrect_length_null_value():
     assert exit_exception_e.value.code == 1
 
 
+def test_check_migration_version_incorrect_n_cols():
+    ds = DataStore("", "", "", 0, ":memory:", db_type="sqlite")
+    revisions = ["version_id", "test_version_id"]
+
+    with ds.engine.begin() as connection:
+        connection.execute(text("CREATE TABLE alembic_version (col1, col2, col3);"))
+        connection.execute(text("INSERT INTO alembic_version VALUES ('', '', '');"))
+
+    with pytest.raises(SystemExit) as exit_exception_e:
+        temp_output = StringIO()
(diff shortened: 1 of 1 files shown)


IanMayo (approved):
Thanks @robintw

PR #999, opened 2021-09-15
DO NOT MERGE - failing build on branch name

IanMayo:
This is an example of a build failing because of a slash in the branch name.

Once we have merged the fix for the issue, this branch should successfully _Deploy_.

### DIFF
--- a/tests/timeline_tests/test_dashboard_sql_functions.py
+++ b/tests/timeline_tests/test_dashboard_sql_functions.py
@@ -1,5 +1,4 @@
 # pylint: disable=duplicate-string-formatting-argument
-# pylint: disable=hardcoded_sql_expressions
 """
     The following describes:
 
@@ -327,7 +326,8 @@ def get_data(listPlatformRanges):
 
 
 def get_query(funcName):
-    return "select * from pepys.dashboard_" + funcName + "(%s, %s)"
+    # security checks suppressed on next line since it isn't processing user data
+    return "select * from pepys.dashboard_" + funcName + "(%s, %s)"   # nosec
 
 
 def populate_data(cursor, TIMELIST):
--- a/tests/timeline_tests/test_dashboard_sql_functions.py
+++ b/tests/timeline_tests/test_dashboard_sql_functions.py
@@ -1,3 +1,4 @@
+# pylint: disable=duplicate-string-formatting-argument
 """
     The following describes:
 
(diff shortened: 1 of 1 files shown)


IanMayo:
Close. The branch name is breaking our deployment code

IanMayo:
@robintw - this branch demonstrates an issue in our `create_deployment.ps1` code. If the branch contains a `/` symbol that gets used in the filename mangling it.

I'll try to remember not to use `/` in a branch name in the future, but it is a working practice I use on other projects to create effective separate folders for `hotfix` and `feature` branches.

I _think_ we just need to do a char-replace in `$output_filename` just after we create it

tests/timeline_tests/test_dashboard_sql_functions.py
```diff
@@ -354,7 +404,10 @@ def get_test_case_data(start, end):
         get_data([fij]),
         '["C","G"]',
     )
-
+def get_test_case_data2(start, end):
```
robintw, line 407: I'm not sure what the purpose of this function is - I think the place it is used, it could just be replaced with the tuple `(start, end)`.

tests/timeline_tests/test_dashboard_sql_functions.py
```diff
@@ -367,12 +420,21 @@ def validateStartTimes(rows, rangeTypes, startTimes):
...
             print(row, ranget, endt)
             return False
     return True
 
+def validateForIncludeInTimeline(rows):
+    for row in rows:
+        # pylint: disable=unused-variable
+        (recordType, includeInTimeline, serialid, pid, pname, exercise, ptname, start, end, gap, ftname, ftcolor) = row
+        if recordType == "SERIALS":
+            return includeInTimeline
```
robintw, line 435: Is this checking exactly what you want it to be checking? As soon as it finds a value of `True` for `includeInTimeline`, it will return True, and the `assert` (used where it is called) will pass. Is that what you want?

robintw (commented):
These tests seemed a bit confusing to me, but then these tests were originally a bit different and don't quite fit the testing style of the rest of the code.

I've made a couple of comments, but I think it might be worth adding an issue to refactor these tests at some point.

tests/timeline_tests/test_dashboard_sql_functions.py
IanMayo: Unnecessary method removed.

tests/timeline_tests/test_dashboard_sql_functions.py
IanMayo: There is just a single row returned by the sample test data - so dropping out on the first cycle is intended.

IanMayo:
@Robin - I had to drop this PR. It wouldn't build because of a `/` in the branch name.

So, the _actual_ code for this PR is in #1002 

I've kept this PR open as a _demonstrator_ for the above build issue, so we can see when fix for #1001 is valid.

IanMayo:
Build issue resolved:
image

PR #1004, opened 2021-09-16
Record extractions to database

### LINKED ISSUE #263: Extend extraction highlighter to record extractions in database
Currently the extraction highlighter code (the `record` methods on `Line` and `Token`) record the usages internally to produce a highlighted HTML output. We should extend this so that they can also record extractions in the database.

robintw:
## 🧰 Issue
Fixes #263 

## 🚀 Overview: 
Adds ability to record extractions to the database. Now the options are to not record extractions at all, to record them just to the highlighted HTML file, or to record them to the highlighted HTML file and the database. 

## 🤔 Reason: 
More traceability for the user

## 🔨Work carried out:

- [x] Update schema of extractions table
- [x] Update HighlightedFile, Token and Line classes to be able to record data into dicts
- [x] Update DatafileMixin to be able to flush extractions to a dict, and then commit them to the database
- [x] Add tests
- [x] Tests pass

## Confirmations


robintw:
This is now ready @IanMayo, and all Codacy warnings can be ignored (three of them are for tests using the old unittest style, which require tests to be methods of a class, and one of them is for a parameter which has to be passed as part of patching with pytest).

pepys_import/file/importer.py
```diff
@@ -22,13 +22,20 @@ def __init__(self, name, validation_level, short_name, datafile_type, default_pr
         self.errors = None
         self.error_type = None
 
-        self.do_recording = True
+        # By default all importers will record extractions to a highlighted
+        # HTML file, but not record them to the database
+        self.highlighting_level = "html"
```
IanMayo, line 27: @robintw - maybe I'm a crusty old Java programmer who is far from `pythonic`, but...

I would always declare these strings as constants, and refer to the constant in the code (to avoid [magic strings](https://deviq.com/antipatterns/magic-strings).  IMHO that avoids typos, and easily lets us find (for example) all of the places where a particular highlighting mode is used.

robintw: I thought you might say that! In python code people tend to use strings rather than constants for these sorts of things, but I can switch to using constants if you'd prefer - I'm happy either way.

IanMayo: Do you have a position on this as we move forward @mew-nsc ? Ours will be the new team-culture from the end of the week ;-)

IanMayo:
@robintw - should we be able to view `extractions` in the `admin-maintenance` page?  Since this data isn't included in any business processes (other than recording) it may be useful to provide (slightly) easier access to it - including for my checking the PR when I only have access to codespaces 🤣

robintw:
@IanMayo You can now! I keep forgetting that we have a hardcoded list of metadata tables that we can view in the GUI, to stop the user viewing tables that aren't used yet. I've added it to that list, and fixed a couple of things so it displays properly.

pepys_import/file/importer.py
mew-nsc: "Named constant?" or "Would this be better as an enum?" are some of my most common code review comments... It looks like an enum might make sense here, but I'm conscious that the Python convention is to use strings - I'm guessing we should probably stick with the language conventions?

robintw: We do use named constants elsewhere (eg. for validation levels - which is a similar setting that is configured for an importer) - so we could use them here.
I'm wondering whether switching to use constants for this PR might be a good task for @mew-nsc, as it would give him a good excuse to look through all the code in a reasonable amount of detail and see how it all fits together. I'm happy to do it myself, but just wondered if that might be useful.

mew-nsc: Agree that it would be useful to give me a better understanding of the extractions. Will take a look.

mew-nsc: Pushed the change (went for an enum based approach). Please shout if you have any concerns

robintw:
Looks good @mew-nsc, thanks

IanMayo:
Hello @robintw  & @mew-nsc 

GH is reporting a conflict, though it seems relatively benign.  Ok if I fix the text, and merge this PR?

robintw:
Yeah that's fine @IanMayo

IanMayo (dismissed):
Tested by modifying REP importer, and viewing results data in `pepys-admin`.

IanMayo:
It looks like `black` wants some newlines.  Have tried to fix.

image

IanMayo:
Closing & re-opening to try to wake up tests.

IanMayo (approved):
Tested & working.

PR #1016, opened 2021-09-17
Fix error when viewing Platform entry in GUI

robintw:
## 🧰 Issue
N/A

## 🚀 Overview: 
Fix bug where trying to view an entry from the Platforms table in the Maintenance GUI led to an error: `Exception Parent instance <Platform at 0x112b86c40> is not bound to a Session; lazy load operation of attribute 'participations' cannot proceed (Background on this error at: https://sqlalche.me/e/14/bhk3)`

## 🤔 Reason: 
Fix bug

## 🔨Work carried out:
- [x] Set up lazy-loading via joining on the backreference from WargameParticipants to Platforms, so that the `participants` attribute on the `Platform` class can be read without needing the object to still be in the session
- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_admin/maintenance/column_data.py
+++ b/pepys_admin/maintenance/column_data.py
@@ -404,6 +404,12 @@ def convert_column_data_to_edit_data(column_data, set_percentage=None):
             # Don't allow to edit ID columns
             continue
 
+        if key == "wargame participations":
+            # Don't allow to edit wargame participations, as joining the data
+            # initially takes a long time, and slows everything down
+            # This can be edited in the Tasks GUI
+            continue
+
         if value["sqlalchemy_type"] == "relationship" and value["second_level"]:
             continue
 
--- a/pepys_admin/maintenance/gui.py
+++ b/pepys_admin/maintenance/gui.py
@@ -31,7 +31,7 @@ from prompt_toolkit.lexers.pygments import PygmentsLexer
 from prompt_toolkit.styles import Style
 from prompt_toolkit.widgets.base import Border, Label
 from pygments.lexers.sql import SqlLexer
-from sqlalchemy.orm import undefer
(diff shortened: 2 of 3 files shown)


IanMayo:
@robintw - this PR is repeatedly performance failing test in CI.  I'd normally ignore it, but it's too slow by quite a large percentage:
image

Could this be an instance where the performance checking is drawing our attention to a potential issue?

(or, maybe some change causes the test to take longer - and we should accept that).

robintw:
I think you're right @IanMayo - this is causing a slow-down because of the way I've done this fix. Let me look into whether there is a better way of doing it - it may be that we don't show this extra bit of information in the GUI (the list of wargames that the platform is participating in), but still allow filtering by it, which would reduce the time penalty but still allow the useful filtering aspect.

robintw:
This should fix the slowdown - and I also learnt a new feature of SQLAlchemy! I will create a separate issue for some other speedups that could be implemented using the same technique.

IanMayo (approved):

pepys_admin/maintenance/column_data.py
```diff
@@ -404,6 +404,12 @@ def convert_column_data_to_edit_data(column_data, set_percentage=None):
             # Don't allow to edit ID columns
             continue
 
+        if key == "wargame participations":
+            # Don't allow to edit wargame participations, as joining the data
```
IanMayo, line 408: @robintw - do we need to explain to analysts that we don't allow this table to be edited, but they can make changes via the `Maintain Tasks` tab?

robintw: We could do, but to be honest I doubt they'd try and edit that manually... The 'obvious' place to change things would be in the Tasks GUI.

PR #1033, opened 2021-09-24
Significantly speedup calculation of table summaries

### LINKED ISSUE #1030: Try to remove two delays in import process
When we run `Pepys-Import` on the target network there are two noticeable delays:
1. When the command line window first opens, and the `bat` file shows `Successfully set paths for Pepys` there is a delay before the app opens. I was able to reproduce this when accessing Pepys on the parent Mac from a MS-Windows VM.  I _guess_ this is similar to accessing files across a network drive.
2. Near the end of the import process there is another delay.  Will visit to capture more details on this.

screencast

robintw:
## 🧰 Issue
Fixes #1030 

## 🚀 Overview: 
Speed up calculation of table summaries at the beginning and end of an import by not retrieving the whole object from the database, but just retrieving the relevant fields. Tests show speedups of around 2x for metadata summaries, and 6.5x for measurement summaries - although I imagine these speed-ups will be higher for the users.

## 🤔 Reason: 
Speed up imports.

## 🔨Work carried out:

- [x] Just query the created_date field rather than the whole object
- [x] Tests pass

## Confirmations

- [x] Any database schema changes are implemented via `alembic revision` [transitions](https://pepys-imxort.readthedocs.io/en/latest/database_migration.html#how-to-use-it-for-developers)

### DIFF
--- a/pepys_import/core/store/table_summary.py
+++ b/pepys_import/core/store/table_summary.py
@@ -1,4 +1,3 @@
-from sqlalchemy.orm import undefer
 from tabulate import tabulate
 
 from pepys_import.core.store import constants
@@ -27,10 +26,7 @@ class TableSummary:
     def table_summary(self):
         number_of_rows = self.session.query(self.table).count()
         last_row = (
-            self.session.query(self.table)
-            .options(
-                undefer("*")
-            )  # Fetch all attributes to enforce to failing if there is any mismatch
+            self.session.query(self.table.created_date)
             .order_by(self.table.created_date.desc())
             .first()
         )
--- a/tests/test_import_cli.py
+++ b/tests/test_import_cli.py
@@ -163,34 +163,6 @@ class TestImportWithWrongTypeDBFieldPostgres(unittest.TestCase):
         except AttributeError:
             return
 
-    @patch("pepys_import.utils.error_handling.custom_print_formatted_text", side_effect=side_effect)
-    def test_import_with_wrong_type_db_field(self, patched_print):
-        conn = pg8000.connect(user="postgres", password="postgres", database="test", port=55527)
-        cursor = conn.cursor()
-        # Alter table to change heading column to be a timestamp
-        cursor.execute(
-            'ALTER TABLE pepys."States" ALTER COLUMN heading SET DATA TYPE character varying(150);'
-        )
-
-        conn.commit()
-        conn.close()
-
-        temp_output = StringIO()
-        with redirect_stdout(temp_output):
-            db_config = {
-                "name": "test",
-                "host": "localhost",
-                "username": "postgres",
-                "password": "postgres",
-                "port": 55527,
-                "type": "postgres",
-            }
-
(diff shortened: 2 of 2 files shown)


robintw:
@IanMayo This all works fine, and is significantly faster. However, loading the whole object from the database had the side-effect of giving us an error if there was a mismatch between the database objects definitions and the actual database schema. We have other things in place now to detect a mismatch in terms of a missing field or a misnamed field, but we won't easily catch a field type change elsewhere. If there is a field type change then it would almost certainly lead to an error at some point, but the benefit of the checking occurring here was that it would be right at the start of an import run - and so we'd catch it before we did anything else to the database.

We have tests that check that we get errors if there are differences between the table objects and the database schema, and they are all passing except the one that checks for a difference in type. I think a difference in type is likely to be fairly unlikely - particularly given that we've done most of the major schema changes now, we're far better at using database migrations too - so I think we can delete this test and continue to merge this PR. In terms of a cost-benefit analysis, I think significantly reducing delays for the users probably beats this extra check on the database structure.

IanMayo:
>  I think significantly reducing delays for the users probably beats this extra check on the database structure.

That seems ok for me. 

Does the omission of this test justify adding something to the PR check-list?  Or, should we just add a unit test that does the fetch?  Aaah, we're testing this specific instance of the database - so it wouldn't test the client database.  Maybe there is some merit in testing `tracstor` though, since that close models the client database.

robintw:
I think any change of type that wasn't reflected in the database or dealt with via a migration would almost certainly trigger an error somewhere in the manual System Test Plan, so I think it's ok to just remove this test. I'll add a commit to do that.

IanMayo:
I agree that the failure would arise in the System Test Plan. But, the point that we learn about the issue is while we're (I'm) preparing a release.  The value of including the test in the unit tests is that the dev will encounter the issue in the PR that introduces the problem - both allowing early resolution and giving the dev a clue as to where the error came from (the current PR).

robintw:
I don't quite see how we can check this in the unit tests? The databases created in the unit tests are created from the code in the repo that is being tested - so wouldn't detect a change in the code that hadn't been updated in the database. Testing against TracStor in the unit tests raises a number of issues - like how do we securely store the credentials for TracStor on both the CI and local dev machines, we would then require internet access to run all the unit tests, and we would have to keep TracStor in a state that can be used by the unit tests. The latter is the most difficult, I think, as we could easily get out of step with migrations on TracStor as migrations would have to be applied as part of the unit tests and then downgraded. I think any unit tests should use transient databases rather than something that hangs around like TracStor.

For what it's worth, I don't think we've ever run into this issue before, and I doubt we're likely to - we don't tend to go around changing the type of fields, and if we do then we'll add a migration (and other unit tests would almost certainly fail).

IanMayo:
That's completely fine Robin. thanks for explaining so clearly. I appreciate that testing against `tracstor` is fragile.

Please feel free to remove the relevant test.

IanMayo (approved):

PR #1039, opened 2021-09-24
Speed up column data creation

### LINKED ISSUE #1036: Performance delay loading Platforms in Maintenance UI
If this is unavoidable, then c'est la view, but...

I entered Maintenance UI of `pepys-admin` multiple times while doing benchmarking.  To get rid of recent test data I had to separately delete the datafile and the platform that's ok.  When I selected `datafiles` the page opened instantly - even though there are lots of them.  But, when I selected `platforms` there was a 20 second delay.

I don't know if we can optimise what's happening there - I guess it's pulling data out of other tables to de-normalise the data for easier working.  

But, if we could remove the delay it would improve the user experience.

robintw:
## 🧰 Issue
Fixes #1036 

## 🚀 Overview: 
This changes the column data creation code to only request the relevant columns inside the for loop, rather than requesting the entire object from the database (including all of the denormalised fields). This speeds things up significantly - in some tests the speedup was more than 25x!

In real-world tests, when loading the Platforms table in the Maintenance GUI, when connected to TracStor (with around 500 Platforms to load), the time spent with the progress dialog during the load reduces from ~9s to ~1s.

## 🤔 Reason: 
Speed

## 🔨Work carried out:

- [x] Only extract relevant field from database each time around the loop
- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_admin/maintenance/column_data.py
+++ b/pepys_admin/maintenance/column_data.py
@@ -221,6 +221,9 @@ def create_normal_column_data(col, data_store, table_object):
     # Skip getting values for the remarks column, as we don't need a dropdown for that
     if details["type"] == "string" and details["system_name"] != "remarks":
         # Get values
+        # Here we query for just the specific column name (details['system_name']) so that
+        # the generated SQL is just selecting that column, rather than selecting all the columns
+        # and doing all the joins to get the denormalised data
         all_records = data_store.session.query(getattr(table_object, details["system_name"])).all()
         values = [str_if_not_none(record[0]) for record in all_records]
         details["values"] = sorted(remove_duplicates_and_nones(values))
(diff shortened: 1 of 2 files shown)


pepys_admin/maintenance/column_data.py
```diff
@@ -220,11 +221,8 @@ def create_normal_column_data(col, data_store, table_object):
     # Skip getting values for the remarks column, as we don't need a dropdown for that
     if details["type"] == "string" and details["system_name"] != "remarks":
         # Get values
-
-        all_records = data_store.session.query(table_object).all()
-        values = [
-            str_if_not_none(getattr(record, details["system_name"])) for record in all_records
-        ]
+        all_records = data_store.session.query(getattr(table_object, details["system_name"])).all()
```
IanMayo, line 224: Just to help out our future-selves could you explain in more detail what is happening above?  I understand the PR description, but don't quite see how the above line addresses the change.  It may be worth inserting a comment on the previous line.

robintw: Done

IanMayo: Aah - I now understand where the performance increase came from :-)

IanMayo (dismissed):

robintw:
The test failure is just a benchmark failure, and this PR doesn't affect the code that is being benchmarked - so it can just be ignored and merged (I assume GH Actions is having a bad day)

IanMayo (approved):

PR #1040, opened 2021-09-27
Change Postgres tests to properly fail if testing server can't be started

### LINKED ISSUE #951: Make Postgres tests skip properly if Postgres setup fails
## 🐞 Overview
Nearly all the tests that use Postgres have some code like this in the class `setUp` function:

```
except RuntimeError:
      print("PostgreSQL database couldn't be created! Test is skipping.")
      return
```
and
```
except Exception:
    print("Database schema and data population failed! Test is skipping.")
```

Unfortunately, the print statement is a lie... (!) The test doesn't skip, it just prints that message, and the test then probably fails later on for some other reason.

We should alter this so that the test actually skips, or fails, rather than just printing a message that will rarely (if ever) be seen.

This should be fixable with some clever find/replace.

robintw:
## 🧰 Issue
Fixes #951 

## 🚀 Overview: 
Changes all usages of the testing Postgres server to give a proper error if the testing server can't be started. Previously there was a printed message saying that the test was skipping - but the test wasn't actually skipped. Instead, this PR makes the test fail with a custom exception explaining the problem - and because it is a custom exception you can look back in the exception tree and see what exception was actually raised by the testing.Postgresql module.

## 🤔 Reason: 
Give sensible errors in the test suite

## 🔨Work carried out:

- [x] Big find/replace operation to change exception handling in Postgres tests
- [x] Tests pass

## Confirmations

### DIFF
--- a/tests/test_admin_cli.py
+++ b/tests/test_admin_cli.py
@@ -807,8 +807,7 @@ class TestAdminCLIWithMissingDBFieldPostgres(unittest.TestCase):
                 port=55527,
             )
         except RuntimeError:
-            print("PostgreSQL database couldn't be created! Test is skipping.")
-            return
+            raise Exception("Testing Postgres server could not be started/accessed")
         try:
             self.store = DataStore(
                 db_name="test",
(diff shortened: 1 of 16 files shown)


IanMayo (dismissed):

IanMayo:
A failing test in this PR @robintw

IanMayo (approved):

PR #1043, opened 2021-09-28
Set up pre-commit on Codespaces

robintw:
## 🧰 Issue
N/A

## 🚀 Overview: 
Installs the pre-commit hooks during Codespaces set up, which means that black/isort/flake8 etc will be run on commit.

## 🤔 Reason: 
Stops code being committed that doesn't meet our formatting rules, and would therefore fail on CI.

## 🔨Work carried out:

- [x] Update Codespaces config
- [x] Tests pass

## Confirmations


robintw:
@IanMayo Could you test this on Codespaces sometime? It should set up the pre-commit hooks. To test it, make a modification to a file with obviously awful formatting, and run git commit - it should fail and tell you that modifications have been made to the files. Normally you'd just add the files again (to get the modifications) and commit again - but of course you don't have to do this final commit for this test.

IanMayo:
@robintw - isn't there some merit it in configuring it as a general pre-push hook?

Then it would be run when we try to push from any dev console.

robintw:
What do you mean?

Pre-commit works before anything is committed, so has to be set up on each developer's machine and has to be set up each time you clone the repository. I don't think there's another way to do it.

robintw:
Ah @IanMayo - I've realised you're suggesting a pre-_push_ hook. I've done some googling, and while those do exist, they don't seem to be used to do code-formatting etc. Presumably because at push time they'd have to actually add an extra commit to do any code formatting changes, and that would clog up the history with loads of extra formatting commits.

Pre-commit hooks are very widely used, and the tool we use is specifically designed to deal with them - so I'd suggest staying with that. They do need setting up for each clone of the repo, but that's specified in our docs.

(Actually, we should check that @mew-nsc has them set up - as he may not have)

IanMayo:
Hello @robintw - I opened the branch in code-spaces. 

There was no `pre-commit` file in `.git/hooks`.   I ran `pre-commit install` and it replied to say a file was created in `.git/hooks`.  I looked at the file, and its contents are like this:
image

Note: from the `codespaces` terminal I was able to run `python -m pepys_admin.cli` - which suggests the startup line is being run.  Well at least the `pip3` component of it.

I was expecting something to happen with the `.pre-commit-config.yaml` file in the root folder.  But, this pre-commit looks unrelated to it.

robintw:
@IanMayo The content of that file looks right for a pre-commit hook - it looks like it's calling the pre-commit module to do the right stuff (you can see the text of the hook does mention the config file etc)

I'm slightly confused about why it wasn't there already when the pre-commit install command should have been run at startup. I'm not really sure what we can do about that - maybe installing pre-commit during the codespaces startup doesn't work because the right bits of git haven't been set up at that point (though that would surprise me).

IanMayo:
@robintw - I can imagine your frustration at not being able to run/debug this yourself 🤣 

Aah yes - I now see the call to the `yaml` file.  Let's not invest too much effort in this - `codespaces` remains a strange environment.  Let's invest more effort when/if they become more widely available.

PR #1049, opened 2021-09-29
Speed up querying in maintenance GUI

robintw:
## 🧰 Issue
N/A

## 🚀 Overview: 
Speeds up the querying of data in the Maintenance GUI, particularly for tables like State that reference a few tables directly (Datafile and Sensor), but then those tables reference loads of other tables (like DatafileType, SensorType, Platform, PlatformType etc).

## 🤔 Reason: 
Speed

## 🔨Work carried out:

- [x] Changed joinedload query configuration
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
In a previous PR we changed one of our SQLAlchemy relationships to not join by default. I then added a `joinedload("*")` call to the query runner in the Maintenance GUI, to ensure that it loaded all of the relationships for the object. However, it turns out that `*` does a fully recursive joined load, which uses joins to load all of the relationships for that object, and all of the relationships for all of the other objects down the tree (eg. not just the State object, but the Sensor object, the Platform object, the PlatformType object etc). This PR changes it to use a configuration where it only does a joinedload on the relationships that are on the top object, and lets the others stay as their default. This stops the recursive nature of the load, and takes the SQL down from 450 lines (!) to something far more manageable.

This is a slightly different bit of code to the one that was tested at the client's - because there were some disadvantages of the simple way we modified it for the client (it could lead to running 100 extra queries in some circumstances). It's hard to test the speed implications of this, as the queries are always very fast for us - but it doesn't seem to make them any slower. It might be worth @IanMayo taking this release in as a beta release and testing on their system - or getting the users to test it themselves.

Also, our automated tests for the GUI are very limited, so it would be good if @IanMayo could have a play around in the GUI with this PR and check that nothing breaks... I've done some manual testing, but it's always good to have someone else have a go.


robintw:
Ok, good about the States. I will make a few notes about what could be the cause of the delays for Platforms and Series. It seems like the delays there are definitely coming from the column data generation.

robintw:
Ok, a few notes on how to possibly improve the speed:

When we create the column data used in the GUI, we query the database to get values that we can make available in dropdowns etc for use when filtering. These values can be just fields of the database, or they can be relationship values, or association proxies. At the moment, when loading a relationship, we load the full denormalised data object, even though it might not be needed. The only fields that are actually needed to create the column data for a relationship are the fields defined in `_default_dropdown_fields` on the model. If those are all normal fields then we don't need to load the full object at all, we can just load those fields (with a query like `session.query(Series.field1, Series.field2).all()`) - however, some of these fields may actually be association proxies, which requires loading the full object (you can't just stick them in the list of things you query like in the previous example - really frustratingly).

For something like the Series, we don't have any association proxies listed in the `_default_dropdown_fields` - so we can save a lot of time by just loading the specific fields. We can also save this time by switching to lazy loading (using the `.options(lazyload("*"))` call on the query), which will not load any of the denormalised data by default, and only load it if we need it. However, if we do this for something that _does_ need one of these association proxy fields, then a _separate query_ will be run for _each_ object returned from the query, when we try and access the association_proxy field. For example, for `WargameParticipants`, we would get a quick load with a simple SQL query when we run the initial query, but then inside the loop for extracting the data from each entry, we'd run a separate SQL query each time we access `entry.wargame_name` (or any other association proxy).

Thus we're in the situation that a simple change to using lazy loading makes some queries much faster, and some queries much slower.

The proper way to solve this would be to add some code that checks what the `_default_dropdown_fields` are, and either:
- If they're all standard fields, just runs a query for those fields
- If they include association proxies, then run a query for the full object, but configure lazyloading for all fields by default (`lazyload("*")`) and then joined loading for the specific fields that will be needed. That isn't quite as simple as it seems, as you can have multiple levels of association proxies - eg. one that proxies to one object and then proxies to a field on a subobject of that - so you can get into very complicated stuff quite quickly. I'm sure a nice function can be written to do it nicely and efficiently, but I haven't had time to focus on it

All of the above also applies to retrieving values directly from association proxies into the column data - and most of the same options and limitations apply.

(As an aside, I looked into `selectinload` which I thought would solve all my problems...but it seems to be even slower - so I don't really know what is going on there).

robintw:
I've removed the debug logging, and I think this should now be merged @IanMayo. It definitely speeds up loading of States, so that's a definite improvement.

My comment above should probably be copied to an issue, to be tackled in a separate PR.

IanMayo (approved):
UI tested onsite, performance improvements witnessed.

Issue #821, opened 2021-03-12
Add more indexes to the database


robintw:
Some of the recent performance issues that the users have experienced may be helped by adding more indexes to the database. We'd need to investigate which ones are most useful to add, as there is a cost for each index (both in storage space, and in extra time when adding a record).

#820 would really help this, with having a large database to test on.

IanMayo:
I think the logical route to doing this is to find an operation that's running too slowly (a bottleneck), then invite @arunlal-v to investigate optimising the performance to remove that bottleneck via indexes.

arunlal-v:
I'm not quite sure on all the Postgis functions that we use in Debrief/Pepys-Import and I shared a [generic recommendation](https://github.com/debrief/pepys-import/issues/745#issuecomment-783614721) for Postgis related bottlenecks. For other bottlenecks, we should be able to analyze the execution plan and come up with indexes and/or rewritten queries.

IanMayo:
I think I can close this @robintw .  There was a performance problem in loading data into Debrief.  We resolved that through a logic change (it was mistakenly trying to load loads and loads of track data).

robintw:
Sounds good - I'm happy with that. We can always investigate later.

PR #1051, opened 2021-09-29
Add database indexes on all time columns

robintw:
## 🧰 Issue
#821 

## 🚀 Overview: 
Adds database indexes for all `time` columns - on tables like `States`, `Comments` etc. This includes `start` and `end` columns on tables like `Activations`.

## 🤔 Reason: 
Speed up database queries

## 🔨Work carried out:

- [x] Update table definitions
- [x] Add migrations
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
@IanMayo This migration may take a little while to run, as it will be going through the table adding each entry to the index. With 3.3m states this could take a little while. We should be aware of this when we go in to do the upgrade - and should definitely ensure no-one else is connected to the database at the time.


IanMayo (approved):
Logical step forward, thanks.

Issue #1022, opened 2021-09-21
Link 16 data - agree implementation


IanMayo:
##  :trophy: Business value
Quick access to high volumes of track data

## :alarm_clock: Business priority

Expect to be of value in JW

We will typically require these items:
* [ ] :name_badge: Unique name for file format `Link 16 v1` and `Link 16 v2`
* [ ] :floppy_disk: Obtain sample data: [V1.csv](https://github.com/debrief/pepys-import/files/7202142/Sample.Data.GEV.Link.16.V1.csv) [V2.csv](https://github.com/debrief/pepys-import/files/7202141/Sample.Data.GEV.LInk.16.V2.csv). Note: the location values have been replaced by random numbers - that's why they don't look like degrees.
* [ ] :notebook_with_decorative_cover: Exploitation plan (custom template)
* [ ] :construction_worker: Implementation. Fortuitously, the analysts have provided us with parsers: [Parsers.zip](https://github.com/debrief/pepys-import/files/7202150/Parsers.zip)
* [ ] :star: Testing & Acceptance
## Implementation notes
| Field | V1 Position | V2 Position | Units | Comments |
|--|--|--|--|--|
| Name | 2 | 4 | N/A | This is the "STN" field, which seems to be a number |
| Latitude | 8 | 13 | Decimal Degrees? | |
| Longitude | 7 | 14 | Decimal Degrees? | Note that Lat/Lon are swapped between the two formats|
| Course | 11 | 16 | Degrees | |
| Speed | 12 | 17 | Feet per second |  Units taken from the parsers which convert from feet to metres on speed |
| Depth | 10 | 15 | Feet | Unlike the parser (which converts from altitude to depth) we don't need to turn -ve|

* The base timestamp is embedded in the filename
* Each line has MMSS.MS values that are always relative (or a few minutes after) the time in the filename, if the value of minutes ever goes down, then increment the hours (this includes generating a file from the title)

mew-nsc:
Looking through the two CSV files that we have and the parsers that MWC use to convert these to REP format, both formats have the same key track information (just at different positions). 

| Field | V1 Position | V2 Position | Units | Comments |
|--|--|--|--|--|
| Name | 2 | 4 | N/A | This is the "STN" field, which seems to be a number |
| Latitude | 8 | 13 | Decimal Degrees? | |
| Longitude | 7 | 14 | Decimal Degrees? | Note that Lat/Lon are swapped between the two formats|
[...]

PR #1053, opened 2021-10-01
Link-16 Importer

mew-nsc:
## 🧰 Issue
#1022 

## 🚀 Overview: 
Adds an importer for the two Link-16 sourced data formats we've been given. 

## 🤔 Reason: 
Quick access to high volumes of track data

## 🔨Work carried out:
- [x] Initial parser created
- [x] Support for version 1 of the file format
- [x] Support for version 2 of the file format
- [x] Tests pass for version 1 of the file format
- [x] Tests pass for version 2 of the file format
- [x] Addressing issues raised from testing against real data
- [x] Questions around the format clarified and answers implemented

## Confirmations


importers/link_16_importer.py
```diff
@@ -0,0 +1,199 @@
...
+    def can_load_this_file(self, file_contents):
+        return True
+
+    def _load_this_file(self, data_store, path, file_object, datafile, change_id):
+        # Read the base timestamp from the filename
+        filename, ext = os.path.splitext(os.path.basename(path))
+        datetime_string = self.extract_timestamp(filename)
+        self.base_timestamp = self.timestamp_to_datetime(datetime_string)
+        self.current_hour = 0
+        self.previous_time = timedelta(hours=0, minutes=0, seconds=0)
```
IanMayo, line 57: hello @mew-nsc - I recognise the role of `previous_time`, but I can't see what `timedelta` is doing here, with the zeroes for h, m, s.

importers/link_16_importer.py
```diff
@@ -0,0 +1,199 @@
...
+        time_token = tokens[1]
+        line_time = timedelta(
+            hours=0,
+            minutes=int(time_token.text[:2]),
+            seconds=float(time_token.text[3:]),
+        )
+
+        if line_time < self.previous_time:
+            self.current_hour += 1
+
```
IanMayo: @mew-nsc - I was going to ask for a Slack call to for you to talk me through the logic in lines 90 & 91, but an explanation in comments would be a more persistent way of capturing what's happening. Would you mind explaining what's happening?

importers/link_16_importer.py
mew-nsc: This is setting the initial `previous_time` to a value that won't trigger the `current_hour` change on the first line in the file (which will always be 0hrs)

importers/link_16_importer.py
```diff
@@ -0,0 +1,199 @@
...
+            lon_degrees_token = tokens[14]
+            heading_token = tokens[16]
+            speed_token = tokens[17]
+            altitude_token = tokens[15]
+
+        # -- TODO - Confirm that this is correct for this data
+        platform = self.get_cached_platform(
+            data_store, platform_name=name_token.text, change_id=change_id
+        )
+        # TODO - confirm the actual sensor type
```
IanMayo, line 123: this may be getting philosophical, but I "think" the sensor type is GPS.  Platforms are reporting their own position via Link-16, and since they're airborne the sensor used to determine their position is GPS.

importers/link_16_importer.py
mew-nsc: I've added some comments in the code around the time handling that hopefully makes things clearer. Please shout if it still doesn't make sense

IanMayo: Ok. I've got it.  The `timedelta` docs say that the default value for all fields is zero - so I guess `timedelta()` would be sufficient.

importers/link_16_importer.py
mew-nsc: That makes sense. Worth confirming, but you've convinced me 😃

importers/link_16_importer.py
```diff
@@ -0,0 +1,199 @@
...
+    def timestamp_to_datetime(timestamp_string):
+        """Converts a Link-16 formatted timestamp into a Python datetime
+        :param timestamp_string: The Link16 file timestamp
+        :type timestamp_string: String
+        :return: a datetime if conversion successful or False if unsuccessful
+        :rtype: datetime | bool
+        """
+        timestamp_format = "%d-%m-%YT%H-%M-%S"
+        try:
+            res = datetime.strptime(timestamp_string, timestamp_format)
```
IanMayo, line 199: while I have DST on my mind, is it worth explicitly specifying that this will be a GMT time?

importers/link_16_importer.py
```diff
@@ -0,0 +1,199 @@
...
+
+    def can_load_this_file(self, file_contents):
+        return True
+
+    def _load_this_file(self, data_store, path, file_object, datafile, change_id):
+        # Read the base timestamp from the filename
+        filename, ext = os.path.splitext(os.path.basename(path))
+        datetime_string = self.extract_timestamp(filename)
+        self.base_timestamp = self.timestamp_to_datetime(datetime_string)
+        self.current_hour = 0
```
IanMayo, line 56: @mew-nsc - I _think_ we need to put the hours from the filename into this field.  Sample data showed that sometimes there was a delay of up to 10 minutes between the time in the filename and the first reading.

If the filename is `09-58-16`, and the first reading is `03:12.23`, then we need to increment the hours to generate a time of `10:03:12`.

tests/test_load_link16.py
```diff
@@ -0,0 +1,162 @@
...
+                self.store.session.query(self.store.db_classes.State)
+                # Elevation == 4222 ft
+                .filter(func.round(self.store.db_classes.State.elevation, 1) == 1286.9)
+                .order_by(self.store.db_classes.State.time)
+                .all()
+            )
+            assert len(results) == 3
+            assert results[0].time == datetime(2021, 5, 9, 1, 46, 38, 100000)
+            assert results[1].time == datetime(2021, 5, 9, 2, 40, 47, 400000)
+            assert results[2].time == datetime(2021, 5, 9, 3, 16, 35, 800000)
```
IanMayo, line 94: @mew-nsc in the above lines, it would be useful to also test that we have parsed location, course & speed too.  

Aaah, and maybe check the name for one of the new platform entries too - I guess it should be the `STN` value.

IanMayo:
@mew-nsc - have just looked at the highlighting report:
image

It looks like you need to capture the recording/extraction of time and track-name (STN).

importers/link_16_importer.py
mew-nsc: It would probably be good to clarify the behaviour of this - the existing parsers take a value for time in rather than reading it from the filename. So I'd taken the offset by the start time from there - so for the case you've given, the result would be 9:58:16 + 00:03:12.23 = 10:01:28

Running the file through the REP parsers with those values (as `start_time` = 095816) gives an output in the REP file of 10:01:51 - Not quite the number I was expecting but delving further it looks to be caused by a bug in the parser code around handling of the milliseconds.

IanMayo:
I've just noted that to support the new "Record extractions" capability we need to "flush" the extractions, like here:
https://github.com/debrief/pepys-import/pull/1004/files#diff-555f89be5cfc941ffffd7e0d9213d1318542555e9b8a79e7c7abd62172c40ef8R152

importers/link_16_importer.py
IanMayo: Aah,  so that's why it's a duration in the code.  I'm pleased we've explored it further.

From looking at the data it's not a duration. It's time.   So, when the time of the file is 12-36-32, the first line will be something like 37:12.01. It's a minute/seconds equal to, or shortly after the time value in the filename.

importers/link_16_importer.py
mew-nsc: I've noted this in the doc-comment. Couldn't see an obvious way to indicate this with the strptime function, but happy to add if I've missed a parameter

importers/link_16_importer.py
```diff
@@ -0,0 +1,217 @@
...
+    @staticmethod
+    def extract_timestamp(filename):
+        """Extracts a Link-16 timestamp from an unmodified filename
+        :param filename: The Link-16 filename that contains a timestamp
+        :type filename: String
+        :return: The timestamp extracted from the filename
+        :rtype: String
+        """
+        base_filename, ext = os.path.splitext(os.path.basename(filename))
+        base_filename, ext = os.path.splitext(base_filename)
```
robintw, line 197: This looks a bit unusual, but I'm guessing it's because there are two extensions on the end of the file? Like `blah.dat.txt`? If so, it's probably worth a comment explaining it.

importers/link_16_importer.py
```diff
@@ -0,0 +1,217 @@
...
+        # Timestamp could be anywhere in the filename, so find it
+        extracted_datetime = re.search(r"(\d+-\d+-\d+T\d+-\d+-\d+)", base_filename)
+        print(extracted_datetime.group(1))
```
robintw, line 200: In the final version, we shouldn't be printing anything (I'm aware this is probably just for debugging, but it's easy to leave it in)

importers/link_16_importer.py
```diff
@@ -0,0 +1,217 @@
...
+        return extracted_datetime.group(1)
```
robintw, line 201: We don't seem to have any error handling around the regex - we should check if there is a 1st group to return before we return. We could just wrap the call to this `extract_timestamp` function in a `try/except`

tests/test_load_link16.py
```diff
@@ -0,0 +1,340 @@
...
+            # Correct elevations
+            assert len(results) == 3
+            # Timestamp checks
+            assert results[0].time == datetime(2021, 5, 9, 1, 46, 38, 100000)
+            assert results[1].time == datetime(2021, 5, 9, 2, 40, 47, 400000)
+            assert results[2].time == datetime(2021, 5, 9, 3, 16, 35, 800000)
+
+            ureg = UnitRegistry()
+            # Location
+            assert results[0].location.latitude == 0.534945836
```
robintw, line 126: You may find some issues with rounding of these float values on different platforms - I've had issues before with failing on CI and not on a local machine, because of float rounding issues. To guard against this, I've always used `round(blah) == 0.534945` or something.

robintw (changes requested):
In general this looks really good @mew-nsc. You've obviously understood how to write an importer, and this has all the correct elements. Well done!

I've made a few minor comments about a bit of error handling and a few other aspects. None of them are particularly major.

I haven't looked into the details of the file format itself before doing this review, so I can't say whether this is an accurate representation of the format - but the Pepys-side of things is fine.

importers/link_16_importer.py
IanMayo: Aah, I know.  We can check this through inspection, since we're currently in DST.  If you print a `timestamp_string` and the resulting object to the command line - you can check they are the same value.  If they are - then we can be confident it is forcing it to UTC.

IanMayo:
Some feedback on the beta-test of this PR in comments here:
https://github.com/debrief/pepys-import/issues/1054#issuecomment-932276154

tests/sample_data/track_files/Link16/V2_GEV_16-05-2021T00-00-00.raw-SLOTS_JMSG.csv
```diff
@@ -0,0 +1,9 @@
+Xmt/Rcv,SlotTime,Set,FrameIndex,STN,RI,Packing,PG,Jx.y,MLI,ADDI,RC,TN,Lat,Long,Alt,Crs,Spd,Id,|,1,2,3,4,5,6,7,8
```
IanMayo, line 1: @mew-nsc - if you switch the filename to have non-zero time (such as `04-05-06`) we can test that the times are taken as minutes after the filename hours - rather than a delta since the filename timestamp.

tests/sample_data/track_files/Link16/V2_GEV_16-05-2021T00-00-00.raw-SLOTS_JMSG.csv
```diff
@@ -0,0 +1,9 @@
...
+SomeStr,59:31.6,B,550,2,0,SomeStr,6,J2.2,3,0,0,1832,0.953371727,0.458539269,25124,5,0.972404565,Friend or Foe,|,Some Hash,Some Hash,,,,,,
+SomeStr,18:45.2,B,560,669,1,SomeStr,6,J3.3,6,0,0,1580,0.762199436,0.296645626,76355,26,0.832872258,Friend or Foe,|,Some Hash,Some Hash,,,,,,
```
IanMayo, line 3: Oh, and if you reduce the STN column to just use a couple of STN values, it will be much quicker to run import test-runs - since we're creating fewer platforms.

tests/sample_data/track_files/Link16/V2_GEV_16-05-2021T00-00-00.raw-SLOTS_JMSG.csv
mew-nsc: Latest commit should have fixed this. Please can you take a look to make sure that the changes make sense.

IanMayo:
@mew-nsc - the `Codacy` warnings _may_ just be pointing to a mistake.  Could you double-check the required number of parameters, and see if the unused var(s) can be deleted please?

image

mew-nsc:
`_load_this_file` and `_load_this_line` are inherited from the base class (well, technically `_load_this_line` isn't because `_load_this_file` is...) and all the parameters are used. 

I've replaced the unused `ext` with `_` - which may make it clearer that it's a throwaway variable to codacy

importers/link_16_importer.py
mew-nsc: Discussion with MWC confirmed sensor type to be GPS

IanMayo:
> `_load_this_file` and `_load_this_line` are inherited from the base class (well, technically `_load_this_line` isn't because `_load_this_file` is...) and all the parameters are used.

Aah, I thought "too many parameters" was because it should be matching some existing interface. I've just googled and it's against a default (and arbitrary) limit of 5 parameters.  We certainly don't need to heed that warning.

IanMayo (approved):

PR #1080, opened 2021-11-11
Show warning/error when the database isn't at the latest revision

### LINKED ISSUE #1079: Warn if database requires migration
Pepys-Admin (and hopefully Pepys-Import) exit if they connect to a database that is newer than expected.  This prevents analysts from mistakenly using legacy instance of Pepys.

However,Pepys-Admin shouldn't exit if it encounters a database that is _older_ than expected - since the expected situation if there is a pending `migration` to be applied.

But, I believe Pepys should show a clear warning if it connects to a database that is older than expected.

It would probably be reasonable for `Pepys-Import` to exit if the database is older.  If we imagine a schema change to handle an element contained in a new file format, then the importer will fail if run against an un-migrated database.

robintw:
## 🧰 Issue
Fixes #1079 

## 🚀 Overview: 
Shows a warning (for Pepys Admin) or an error (for Pepys Import) if the database isn't at the latest migration.

**Pepys Admin:**
image
**Pepys Import:**
image

## 🤔 Reason: 
Make it easier for the users to know when a migration needs to be run.

## 🔨Work carried out:

- [x] Add check against latest_revisions.json file during the DataStore `__init__` method
- [x] Make check into error (not warning) when running Pepys Import 
- [x] Tests pass

## Confirmations


pepys_admin/utils.py
```diff
@@ -187,23 +187,30 @@ def database_at_latest_revision(db_path):
         )
         return False
 
+    versions = read_latest_revisions_file()
+    if len(versions) == 0:
+        return
```
IanMayo, line 192: @robintw - the other `return` statements all include a boolean. Should we return `False` here?

IanMayo (approved):
Will accept for now, and remember to verify once we are due to next have a database migration.

PR #1085, opened 2021-11-15
Fix GUI crash when filtering by UUID on Postgres

### LINKED ISSUE #1084: Crash in Pepys-Admin
I'm trying to find a platform using platform ID using `Maintain data`

I selected `Platforms`,  then the `Platform_ID` field, and the `Like` operator. As I start to enter the start of the platform id `e957` I get this error:
image

robintw:
## 🧰 Issue
Fixes #1084 

## 🚀 Overview: 
Fixes crash when trying to search by a field with a UUID type in the Maintenance GUI.

## 🤔 Reason: 
Stop Pepys crashing

## 🔨Work carried out:

- [x] Ensure that both SQLite _and_ Postgres UUID types are checked for before casting to a string when adding a string filter on a UUID type
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
When doing a `LIKE` query on a UUID field, we need to convert it to a string first. In the code that generates the SQLAlchemy filter conditions, this was only happening for SQLite UUID types (which are basically just a string anyway), but not for Postgres UUID types. This fixes that.


pepys_admin/maintenance/widgets/filter_widget_utils.py
```diff
@@ -122,7 +123,7 @@ def handle_three_variables(class_obj, output, idx, query_list, data_store):
 
     if ops == "LIKE":
         try:
-            if isinstance(col.type, UUIDType):
+            if isinstance(col.type, UUIDType) or isinstance(col.type, UUID):
```
IanMayo, line 126: Could you explain what the issue was here, please @robintw ?

robintw: I mentioned it briefly in the Developer Notes in the PR description, but basically this if statement is checking whether the column is a UUID. If it is, then it casts it to string before trying to do a `LIKE` operation on it. However, we were only checking whether it was a `UUIDType` (which is the type used for UUIDs on SQLite), and not checking against `UUID` (which is the type used for UUIDs on Postgres). Therefore, it failed on Postgres but worked on SQLite.

IanMayo (approved):
Tested. Filtering by platform_id works

pepys_admin/maintenance/widgets/filter_widget_utils.py
IanMayo: Aah, got you - thanks.

IanMayo (approved):
Tested, working fine

PR #1086, opened 2021-11-15
1071 wecdis importer

### LINKED ISSUE #1071: Importer for WECDIS
##  :trophy: Business value
Native Pepys importer for WECDIS format data

## :alarm_clock: Business priority
Higher priority format based on customer need

We will typically require these items:
(Use `convert to issue` if it is to be delegated/assigned)

* [x] :name_badge: Unique name for file format (include in issue title)
* [ ] :floppy_disk: Obtain sample data (attach if allowable)
* [ ] :notebook_with_decorative_cover: Exploitation plan (custom template)
* [ ] :construction_worker: Implementation
* [ ] :star: Testing & Acceptance

## 💾 Sample
Sample data provided has only one file example, but does include a parser to REP format
- TBC 

### Identifying file as wecdis
- `.log` extension
- date stamp on first line
- VNM tag contains the platform name

mew-nsc:
## 🧰 Issue
Fixes #1071 

## 🚀 Overview: 
Created an importer to import WECDIS format data

## 🤔 Reason: 
To get WECDIS formatted data into Pepys

## 🔨Work carried out:
- [x] Read DZA lines for time
- [x] Read CPOS lines for ownship position tracks
- [x] Read VNM for ownship name
- [x] Read CONTACT for contacts
- [x] Read TMA for contacts
- [x] Read TTM lines for contacts
- [x] Check other lines for useful information that isn't contained in contact/position lines
- [x] Tests pass

## Confirmations


importers/wecdis_importer.py
```diff
@@ -0,0 +1,344 @@
...
+        return suffix.upper() == ".LOG" or suffix.upper() == ".TXT"
+
+    def can_load_this_filename(self, filename):
+        return True
+
+    def can_load_this_header(self, header):
+        return "$POSL" in header
+
+    def can_load_this_file(self, file_contents):
+        # Need to differentiate from general NMEA - so check charts/version available
```
IanMayo, line 44: @mew-nsc - is it worth limiting how many rows of data we merge before we check for chart contents.  I'm pretty sure analysts get WECDIS logs that are 100Mb long.  I guess we can expect the charts to be declared in the first 100 rows.  I can verify when I see the collated samples, next time I'm in.

mew-nsc: Good point Ian. I've made that change so that I don't get caught out while testing. Would be good to know whether VER/Chart tags are always present in WECDIS files, and whether they always appear early on.

importers/wecdis_importer.py
```diff
@@ -0,0 +1,493 @@
...
+        if len(dza_tokens) < 4:
+            self.errors.append(
+                {self.error_type: f"Not enough parts in line {line_number}. No timestamp provided."}
+            )
+            return
+        if dza_tokens[1].text != "DZA":
+            raise TypeError(f"Expected a DZA line, given {dza_tokens[1].text}")
+        date = dza_tokens[2].text
+        time = dza_tokens[3].text
+        # TODO - confirm where the timezone is determined (presumably not always GMT)
```
robintw, line 123: I see this TODO about time zone is still in here: any progress on this?

importers/wecdis_importer.py
```diff
@@ -0,0 +1,493 @@
...
+        contact_id_token = tokens[5]
+        speed_token = tokens[6]
+        date_token = tokens[9]
+        time_token = tokens[10]
+        lat_token = tokens[15]
+        lat_hem_token = tokens[16]
+        lon_token = tokens[17]
+        lon_hem_token = tokens[18]
+        # TODO - figure out which one is the range
+        # TODO - figure out which sensor the contact comes from
```
robintw, line 154: Some more TODOs here?

importers/wecdis_importer.py
```diff
@@ -0,0 +1,493 @@
...
+        lat_hem_token = tokens[4]
+        lon_token = tokens[5]
+        lon_hem_token = tokens[6]
+        sensor_name = f"{source_token.text}_{sensor_token.text}"
+        combine_tokens(source_token, sensor_token).record(self.name, "sensor", sensor_name)
+
+        platform = self.get_cached_platform(data_store, self.platform_name, change_id=change_id)
+        sensor = self.get_cached_sensor(
+            data_store=data_store,
+            sensor_name=f"{sensor_name}",
```
robintw, line 219: This f-string isn't needed - it can just be `sensor_name=sensor_name`

tests/test_load_wecdis.py
```diff
@@ -0,0 +1,637 @@
...
+
+            # there must be one datafile afterwards
+            datafiles = self.store.session.query(self.store.db_classes.Datafile).all()
+            self.assertEqual(len(datafiles), 1)
+
+            stored_states = self.store.session.query(self.store.db_classes.State).all()
+
+            ureg = UnitRegistry()
+            # 1 - subsurface
+            assert stored_states[0].elevation.to(ureg.meter).magnitude == -123
```
robintw, line 153: This works fine, but possibly an easier way of doing it would to be do:

```
assert stored_states[0].elevation == -123 * ureg.meter
```

(In general, multiplying by a unit object assigns that unit)

robintw (commented):
This looks good, thanks.

There are still a number of TODO entries in the file - I haven't commented on all of them, as you can just search to find them. Some of these TODO items look quite important (eg. extra fields to import, working out units etc) - are we still waiting for resolutions for them all?

Also, is it intended that this importer will use the new 'Unknown Platform' functionality? I think the implementation of that is finished, I just need to write the tests now - so if you want to merge that branch in here and start using it 'in anger' to properly test it, then that'd be fine.

tests/test_load_wecdis.py
IanMayo: Yes, @mew-nsc - I'd like to start dog-fooding the `unknown platform` handling when I also test the WECDIS import against their data.

Are comfortable merging @robintw 's branch into this one @mew-nsc ?

IanMayo:
Hello @mew-nsc - I've just run through an import of WECDIS test data to generate highlighted report.

I wasn't sure where the `TTM-18_a` and `TTM-18_b` sensor names came from:
 screencast
 
Could you explain what is happening there please?

IanMayo:
aah - one other observation. Is there really not a comma between the track number (`M02`) and the speed (`12kts`)?
image

importers/wecdis_importer.py
```diff
@@ -0,0 +1,493 @@
...
+        bearing_valid, bearing = convert_absolute_angle(
+            bearing_token.text, line_number, self.errors, self.error_type
+        )
+        if bearing_valid:
+            contact.bearing = bearing
+            bearing_token.record(self.name, "bearing", bearing)
+
+        range_valid, contact_range = convert_distance(
+            range_token.text,
+            unit_registry.kilometers,  # TODO - confirm whether units are km, kyds or NM
```
IanMayo, line 434: this is certainly quite a biggie

mew-nsc: No units are mentioned in MWC's parser, how does Debrief handle the REP data that's generated? Can the analysts update the units in Debrief after the REP file is loaded?

tests/test_load_wecdis.py
mew-nsc: I'm happy with that. Does it make sense to change the target of the existing PR for the unknown platform handling to this branch, or to just do a straight merge across?

mew-nsc:
> @mew-nsc You said that you'd found some issues with memory usage when importing a large file. Do you have any large files that I could test this with and do some memory profiling? Or if not, is it possible to make a synthetic large file?
> 
> I've done a bit of memory profiling with the largest file we currently have (NMEA_Trial.log for the NMEA importer), and it seems like a fair amount of the memory usage is coming from the highlighting code, which has to store all of the individual token characters and character locations.
> 
> Can you try processing the large file with all highlighting turned off for that importer? This can be done by calling `self.set_highlighting_level(HighlightLevel.NONE)` in the `__init__` of the importer.
> 
> Turning off the highlighter for the NMEA importer reduces memory usage for an import of NMEA_Trial.log from 679Mb to 120Mb!
> 
> The other main source of memory usage will be the SQLAlchemy objects - below is something I wrote about this before I realised that it was probably the highlighter that was causing the most problem:
> 
> I get the impression that this file format will be importing a _lot_ of different platforms/sensors, and a _lot_ of individual contacts. I can see that we might run into memory issues, as we'll be holding a lot of SQLAlchemy objects in memory before we decide whether to commit them to the database after we've finished processing that file. There should be some ways around this, but it could be a bit complicated: we don't want to start committing measurements half-way through importing the file, as we don't want the measurements committed if there turns out to be an error later in the file. Two initial thoughts would be to see if we could either a) store dictionaries of column/value pairs, rather than full SQLAlchemy objects (which has the disadvantage of losing a lot of the nice SQLAlchemy stuff we have, and may well break validation etc) or b) trying to write these objects out to a file on disk when the in-memory store gets too large, and then re-load from disk when finally committing to the database

Thanks @robintw I'll check with the highlighter turned off when I get access to the classified data tomorrow. If that doesn't make much difference, I'll generate a large file for profiling. 

The Wecdis data file that we've got so far is all for the same platform, but there are several different sensors being used to collate information on contacts, though shouldn't be a huge number of them - so I'm hoping this isn't a problem for Wecdis, but may be something to revisit for other formats that do have lots of platforms/sensors.

mew-nsc:
> Hello @mew-nsc - I've just run through an import of WECDIS test data to generate highlighted report.
> 
> I wasn't sure where the `TTM-18_a` and `TTM-18_b` sensor names came from: screencast
> 
> Could you explain what is happening there please?

Those sensor names would have come from the TTM lines in the sample file. This was the form being used by the MWC parser, I'll admit that I was surprised that these are numbers. I'm not really sure what the underlying source for the TTM data is, though the format strongly indicated some kind of fused contact information (it consistently refers to a/b in the parser and there are two pairs of bearings/ranges in the data)

importers/nmea_importer.py
```diff
@@ -45,7 +45,9 @@ def can_load_this_header(self, header):
         return "$POSL" in header
 
     def can_load_this_file(self, file_contents):
-        return True
+        # Need to check this isn't WECDIS (Which is derived from this format)
+        contents_string = " ".join(file_contents[0:100])
+        return "VER" not in contents_string and "CHART" not in contents_string
```
IanMayo, line 50: Minor observation, please @mew-nsc 

There is a chance that `VER` or `CHART` could appear in the first 100 lines, within some other message - particularly as a substring of another piece of text.

But, if tested for `,VER,` and `,CHART,` we could be much more confident that it's not WECDIS data.

Aah, I've just found a `spec` for NMEA.  It demonstrates that `,VER,` and `,CHART,` are not NMEA Messages:
https://gpsd.gitlab.io/gpsd/NMEA.html#_ttm_tracked_target_message

Surprisingly, the spec is maintained by Eric Raymond, the Open Source champion.

importers/wecdis_importer.py
```diff
@@ -0,0 +1,494 @@
...
+    def can_load_this_filename(self, filename):
+        return True
+
+    def can_load_this_header(self, header):
+        return "$POSL" in header
+
+    def can_load_this_file(self, file_contents):
+        # Need to differentiate from general NMEA - so check charts/version available
+        contents_string = " ".join(file_contents[0:100])
+        return "VER" in contents_string and "CHART" in contents_string
```
IanMayo, line 49: Aah, I suggest we make the same increased constraint as above here: `,VER,` and `,CHART,`.

importers/wecdis_importer.py
IanMayo: Onsite I only found two WECDIS files.  One had a `TZS` entry with `UTC`, the other didn't have a TZS entry.

I'll send a message to try to get more sample data.

IanMayo:
Hello @mew-nsc - I've been working with the importer tests, to verify we're failing for the correct reason, as explained here:
https://github.com/debrief/pepys-import/pull/1086/files/7f0219356ba716b60616fa0d8b440108da7b0af6#diff-0cbc34d1c6ef21660228e81c1388a5fb90caf28212605e4dffdbf85b9cab0aae

I've found an issue with the location parser.  There is an invalid longitude on this line (it has 4 digits in the whole part, rather than 5):
image
The parser creates a position of 3 degrees West.

I guess we just need to check that `lat` values have 4 digits in the whole number, and `long` values have 5 digits in the whole number.  It happens that previous tests failed with the wrong number of digits, but we're missing this one.

IanMayo (approved):
Approving, so we can include it in release & test on target network

PR #1088, opened 2021-11-17
Support Unknown Platforms for bulk data entry

### LINKED ISSUE #586: Referential integrity with bulk data
> Parsing hundreds of white shipping [*] units got too tedious - analysts gave up and just dragged their rep files directly into debrief. It's suggested to have default sensor, sensor types per Importer and the cli ability to bulk 'just import the names, however silly'.

Maybe the use of _RESOLVER:DEFAULT_ would have overcome the above issue.  

Hmm, but resolver:default values may be hard-coded in Pepys, and analysts may wish to use different defaults in different scenarios.

A defaults config file may be preferable.    So, for a folder of white-shipping data-files, they create a single defaults config file that provides nationality, sensor, privacy, etc.  Then use `-RESOLVER:white_config.ini`.   (Maybe not `.ini` files, I don't mind - but we use them already)

Aah, we can probably grow something quite useful here.  If the default value isn't found in the config file, then Pepys asks at the command line.  So, an analyst may create a defaults file with nationality and sensor privacy, platform privacy.  For all data loaded with that config file in the root folder, those values get auto-filled, but the analyst still provides platform name, platform type, sensor name, sensor type, etc.

[*] White shipping = general shipping that is unrelated to the exercise.  Ferries, fishing boats, cargo ships etc.

robintw:
## 🧰 Issue
Fixes #586 

## 🔨Work carried out:

- [x] Tests pass

## Confirmations


robintw:
Hi @IanMayo. I've done a 'first go' implementation of this tonight - it was actually a lot easier than I thought it would be.

I've tested it manually and it seems to be doing what I'd expect: I've checked importing various files and looked at the results in the Maintenance GUI.

Could you have a play with it and see if it behaves as you'd expect it to? I'll get on with writing the tests when I have chance.

Specifically, things to look at:
- You should see two new options when resolving a platform - to store as unknown, or store all remaining platforms as unknown. Can you check these behave in the way you'd expect. (The 'all remaining platforms' one only applies for _that specific file_.
- You can modify an importer (such as E-Trac, as it's a nice simple one) and add an argument `unknown=True` to a call to `get_cached_platform` - that will tell the resolver not to even ask the user, but to record it as unknown. You can try passing `None` as the platform name, and see that it generates its own UUID name.

pepys_import/core/store/data_store.py
```diff
@@ -1008,8 +1014,41 @@ def get_platform(
...
+        if unknown:
+            # The importer has specifically said it wants to get a 'unknown platform' (ie. one with an unknown
+            # nationality and platform type) back from this call.
+            if platform_name is not None:
+                # The importer has provided a name, which we assume means they think this name is unique
+                # (like an IMO ID), so we search for this name and return it if it exists
+                # We'll use the first three characters of the platform name as the identifier - as this will have been used
+                # when the unknown platform was automatically created
+                platform = self.find_platform(
+                    name=platform_name, identifier=platform_name[:3], nationality="Unknown"
```
IanMayo, line 1032: If we have an IMO number, I think this is a legitimate `identifier` - and we shouldn't trim it to 3 chars.  I think we should have the full IMO in name _and_ identifier.  At some point in the future the analyst takes a deeper interest in the platform, and finds the name for the platform with that IMO.  Then they can swap the name value for the real name ("Solent Express"), and we'll have valid values in `name` and `identifier`.

pepys_import/database/default_data/Nationalities.csv
```diff
@@ -254,4 +254,5 @@ Wallis and Futuna,
 Western Sahara,
 Yemen,
 Zambia,
-Zimbabwe,
\ No newline at end of file
+Zimbabwe,
+Unknown,
```
IanMayo, line 258: @robintw - we won't have these 3 `Unknown` values in the analyst's PostGis instance.  Does Pepys generate the value if it isn't already preasent?  Or do we need to run a migration?

(yes - this question will probably get answered as I get further down the review 🤣  )

pepys_import/resolvers/command_line_resolver.py
```diff
@@ -200,11 +213,40 @@ def is_valid_dynamic(option):  # pragma: no cover
...
+        elif 5 <= int(choice) <= len(choices):
             # One of the pre-existing platforms was chosen
-            platform_index = int(choice) - 3
+            platform_index = int(choice) - 5
             return platforms[platform_index]
 
+    def add_unknown_platform(self, data_store, platform_name):
+        if platform_name is None:
+            platform_name = str(uuid.uuid4())
+        identifier = platform_name[:3]
```
IanMayo, line 229: Aah, as before... I dont think we should produce the identifier by trimming the platform name.  If the IMO is being provided as `name` - then the full value is the best identifier we have, I think...

tests/gui_tests/test_gui_methods.py
```diff
@@ -583,6 +584,7 @@ def test_generating_column_data(pytestconfig, test_datastore):
                 "Yemen",
                 "Zambia",
                 "Zimbabwe",
+                "Unknown",
```
IanMayo, line 587: This is for a separate PR, but Isn't it worth defining this list of countries (plus the list of platform and sensor types) in just one location?  I guess we could have a `tests/consts.py` file where they are centrally defined.
Yes - the rate at which these lists change is forever dropping, and we may not require any future changes, but it seems like a (very small) piece of technical debt.

IanMayo:
Working as expected for SQLite.  Will try PG when I'm next working (prob tomorrow).

* I've loaded `NMEA_TRIAL.log` and `Store as Unknown` worked
* I've loaded `e_trac` data, and `Store remaining platforms as Unknown` worked.
* I've extended the help text for `RESOLVE_PLATFORM` (to save you the quite boring chore). I note this won't be reflected in the analyst's Postgis instance, since the values are just used at database creation.  A migration doesn't seem justified for this change - I'd be happy to do it by hand on my next visit. Is that reasonable @robintw ?

* [ ] I think we should explain that `Remaining platforms` is just for the current file. So, the option should probably be `Store remaining platforms in this datafile as Unknown`.

pepys_import/database/default_data/Nationalities.csv
robintw: It didn't, but it does now.

pepys_import/core/store/data_store.py
robintw: Changed now.

pepys_import/resolvers/command_line_resolver.py
robintw: Changed now.

robintw:
Great @IanMayo. I've updated the text of the 'store remaining' option, and yes updating the help texts manually would probably be easiest.

IanMayo:
@robintw - you did suggest merging this early into the WECDIS PR.  But, it would be _cleaner_ to merge this one first, and then the WECDIS PR - then each PR just covers one area of functionality.

We can't _really_ finish this issue until we have the tests.  But, I could merge a "Unknown Platforms Part-1" PR, then "WECIDS", then you could finish off the tests in "Unknown Platforms Part-2".  

Does that sound ok?

robintw:
That sounds fine @IanMayo.

I won't have chance to do the tests for a day or two, as I had the booster jab yesterday and am feeling rubbish today.

IanMayo (approved):
Tested & working.

PR #1092, opened 2021-11-25
Compare version with network version on launch, and add Update Pepys option

### LINKED ISSUE #1035: Delay associated with running Pepys across network
When `pepys-import` or `pepys-admin` are run on the client network there is a 20 second delay before the python code starts.

I copied Pepys to the `C:` drive, and the apps opened with a delay of only around a second.

So, we should work with onsite technical support to provide local installs of Pepys.

In discussion today, we came up with this strategy:
1. Create "Update Pepys" script, which overwrites the local `C:\Site Programs\Pepys` folder with one from `M"\tools\Pepys\current", then runs `bin\install.bat"
2. Create script which creates shortcut to above script, called "Install updater", and puts it into the start menu
3. Invite Pepys users to run `install updater` from network location, then `Update Pepys` from their start menu

Then, on new updates, Pepys users will be invited to run `Update Pepys` from their Start menu

robintw:
## 🧰 Issue
Fixes #1035

## 🚀 Overview: 
Adds functionality to compare the current Pepys version to the version of a 'master' network install on Pepys launch, and give an error if the network version is newer. Also add an Update Pepys Start Menu item which can be run to copy the network install to the local install directory.

## 🤔 Reason: 
Help the users keep Pepys versions up-to-date while running the code locally so that it is faster.

## 🔨Work carried out:

- [x] Add checks between local version and network version on launch
- [ ] Add Update Pepys Start Menu entry
- [x] Tests pass

## Confirmations


requirements.txt
```diff
@@ -21,4 +21,5 @@ Flask
 Flask-CacheBuster
 flask-cors
 waitress
-lxml
\ No newline at end of file
+lxml
+packaging
```
IanMayo, line 25: hey @robintw - `packaging` was a good find :-)

robintw: I was really confused why it wasn't in our requirements.txt already, as I swear I remember adding it - then I remembered that I'd used it in one of my other open-source projects recently, and had got confused!

robintw:
@IanMayo This all seems to be working now, based on tests on my Windows machine, but I'd appreciate you testing it too.

To test, you will need to do something like this on a Windows machine:

1. Copy your Pepys Install folder to another folder somewhere called something like `network-pepys`. This will be the fake network copy that we'll be comparing to, and copying from.
2. Modify the `__init__.py` file inside the `pepys_import` folder in this copy to change the version number to something higher than the current (eg. `1.0.0`)
3. Modify the Pepys config file so that there is a new section like this at the end

```
[network]
master_install_path = c:\network-pepys
```
(replace the path with the path to your fake network version)
4. Run Pepys Admin or Pepys Import, you should get an error saying that the network version is newer.
5. Run `Upgrade Pepys` from the Start Menu. It should say that it is copying from the fake network location, and then print out a list of files it is copying as it copies them.
6. Run Pepys Admin again, and you should see that it has been upgraded to the new version - with the new version number displayed in the welcome banner.

Can you have a go and let me know how you get on. Once you've confirmed it is working, we'll need to work with the clients to get it working on their system, with the config file set up correctly etc.

IanMayo:
@robintw - did you consider deleting some (or all) local files before copying the new ones over?

If we rename a source-code file in Pepys, the previous one will still be there.  This could be relevant for `importers` - since a previous (broken) one may still be present in the folder, and get used.

Clearly if an analyst puts their own importer in the folder, or modifies an official version they'll lose their changes. I've just looked, but didn't find any documentation on a 'user' importer folder - where locally developed importers would be stored. Do you remember that at all?

robintw:
@IanMayo Good point.

I've updated the PR so that it now deletes the contents of the folder before copying the new version across. I've tested it and it works. You'll see from the code that it deletes things in a couple of steps, as Powershell/cmd.exe seems to want to keep hold of the bin folder even when you change directory out of it. We get around that by not actually deleting the bin folder, just deleting all the other folders, and then deleting the _contents_ of the bin folder but not the folder itself. Seems to work fine.

IanMayo:
This just worked fine, thanks @robintw .  But, we have mixed language.  The CLI advises me to `run the Update Pepys program`, but the item in the Start Menu is called `Upgrade Pepys`.

Will modify the CLI message.

IanMayo (approved):
Tested & working

PR #1100, opened 2021-12-03
Extend length of two fields in database

### LINKED ISSUE #1099: Need to increase size of Track Number in PG
@mew-nsc 's PR for WECDIS Importer (#1086 ) is failing because of a long Track Number (well, name):
https://github.com/debrief/pepys-import/runs/4396329479?check_suite_focus=true#step:7:1250

For WECDIS data it appears users can use free text to name tracks, and can end up with quite long descriptors.

We should increase the size of the `Contacts`.`TrackNumber` field from 20 characters to 40 characters, please.

Note: the PR also appears to fail the performance test, though small failures in this don't normally bother us.

It would be good to resolve this before I try to test WECDIS against bulk _real_ datasets.

robintw:
## 🧰 Issue
Fixes #1099 

## 🚀 Overview: 
Extends the length of two fields in the database, as we were running into problems fitting data in them. The two fields are:

- Contacts.track_number (this was from long 'numbers' used in the WECDIS data)
- Platforms.identifier (this was from UUID identifiers used for unknown platforms)

## 🤔 Reason: 
Allow longer data.

## 🔨Work carried out:

- [x] Update database models
- [x] Add migrations
- [x] Tests pass

## Confirmations

## 📝 Developer Notes:
I already had a migration to extend the length of the Platform.identifier field in my jupyter branch, as I ran into problems with it when importing the AIS data. I thought it made sense to consolidate both of these changes in one PR.


IanMayo (approved):
Thanks @robintw

PR #1108, opened 2021-12-08
Prepend the Pepys path to the PATH variable, rather than appending it

### LINKED ISSUE #1107: Issue with installing Pepys on multiple occasions
## Version

0.0.34

## 🐞 Description

I've just been running the wrong version of Pepys.  I copied Pepys to the local drive and ran `install_pepys.bat`.  But, it was running the network installation, since that appeared earlier in the `PATH` variable.

1. I think we should certainly prepend (_rather than append_) the PEPYS_PATH to the `PATH` variable on installation
2. It would be good to try to remove existing PEPYS paths from the PATH. I guess we could try to spot entries with `\Pepys\` somewhere in the path.

robintw:
## 🧰 Issue
Fixes #1107.

## 🚀 Overview: 
Changes the installation script to _prepend_ the Pepys path to the `PATH` environment variable, so it takes priority over other entries on the path.

## 🤔 Reason: 
So the correct version of Pepys is used.

## 🔨Work carried out:

- [x] Modify install pepys script
- [x] Tests pass

## Confirmations

### DIFF
--- a/bin/install_pepys.ps1
+++ b/bin/install_pepys.ps1
@@ -111,7 +111,7 @@ try {
     if (!($env:Path -split ';' -contains $pepys_bin_path)) {
         [Environment]::SetEnvironmentVariable(
             "PATH",
-            [Environment]::GetEnvironmentVariable("PATH", [EnvironmentVariableTarget]::User) + ";" + $pepys_bin_path,
+            $pepys_bin_path + ";" + [Environment]::GetEnvironmentVariable("PATH", [EnvironmentVariableTarget]::User),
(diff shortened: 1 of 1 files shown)


robintw:
In this PR I've done a 'quick fix' that prepends the Pepys path rather than appends it. It _might_ be possible to look for other possible Pepys paths and remove those, but that's going to be a little tricky and take a while - and I'm not sure whether it's actually the best idea. We can discuss that going forward, but for now this should at least get things working.

IanMayo (approved):

PR #1111, opened 2021-12-13
1101 Wecdis fixes

### LINKED ISSUE #1101: Trip Plan 2021/12/08
- install new release
- run migration
- verify local install & checking network version
- test WECDIS importer
- inspect other WECDIS data. Does it always have `,VER,` AND `,CHART,` in the first 100 lines
- inspect FullShore data, to verify importer assumptions:
    - Lat/Lon values look to be decimal radians, but it would be good to confirm
    - Height appears to be in metres (there's an entry of about 11000 for a height)
    - Speed could be anything
    - Course appears to be radians, but I couldn't see anything > 1 and MWC's parser multiplies these values by 3.6
- Update Help Texts in database (see #1095)

mew-nsc:
## 🧰 Issue
Fixes #1101 - Wecdis issues

## 🚀 Overview: 
Addressed the various Wecdis issues raised by @IanMayo following visit to MWC

## 🤔 Reason: 
To allow the Wecdis importer to better import data flexibly

## 🔨Work carried out:
- [x] Create ownship sensor rather than prompting
- [x] Turn off highlighting
- [x] Give hints about the sensor for contacts
- [x] Added Heading/Velocity to POS ownship message 
- [x] Tests pass

## Confirmations


IanMayo:
Thanks @mew-nsc   I'm onsite again tomorrow, and know I'll have time to do more WECDIS investigations.  I'll be posing questions for the WECDIS SME, but it would also be useful to test this branch.  Do you think I should be merging it to test the release, or do a private beta test?

mew-nsc:
> Thanks @mew-nsc I'm onsite again tomorrow, and know I'll have time to do more WECDIS investigations. I'll be posing questions for the WECDIS SME, but it would also be useful to test this branch. Do you think I should be merging it to test the release, or do a private beta test?

Thanks Ian, that would be useful. I'll be testing on the data we've been given tomorrow too, so hopefully between us we should be able to iron out any issues. I'll also look into why the tests are failing on this PR - they don't locally - so it's probably worth you doing a private beta test.

mew-nsc:
> > Thanks @mew-nsc I'm onsite again tomorrow, and know I'll have time to do more WECDIS investigations. I'll be posing questions for the WECDIS SME, but it would also be useful to test this branch. Do you think I should be merging it to test the release, or do a private beta test?
> 
> Thanks @IanMayo, that would be useful. I'll be testing on the data we've been given tomorrow too, so hopefully between us we should be able to iron out any issues. I'll also look into why the tests are failing on this PR - they don't locally - so it's probably worth you doing a private beta test.

Actually, it looks like these tests are the same ones that are failing in the Full Shore PR - I'm guessing there's an issue in the development branch.

IanMayo:
Hello @mew-nsc - I'm running a build from around 0830 this morning against client data.

It's a dived platform.  For each positional sensor, I'm being invited to create a sensor with that name.  I'm not adding value in providing the name, classification, or sensor-type.  I think the WECDIS importer should just create them.

IanMayo:
Runtime error, possibly related to your new code.
```
Position index == 2
speed: {'1': <Quantity(0.0, 'knot')>}
hdg: {}
Trackbacl (most recent call last):
wecdis_importer.py, line 74 in load_this_file self.handle_position
wecdis_importer.py, line 294, in handle_position state.speed = self.platform_speed[key]
KeyError: '2'
```

mew-nsc:
> Runtime error, possibly related to your new code.
> 
> ```
> Position index == 2
> speed: {'1': <Quantity(0.0, 'knot')>}
> hdg: {}
> Trackbacl (most recent call last):
> wecdis_importer.py, line 74 in load_this_file self.handle_position
> wecdis_importer.py, line 294, in handle_position state.speed = self.platform_speed[key]
> KeyError: '2'
> ```
> 
> It _looks like_ we're trying to retrieve `platform_speed[2]`, but there is only one item in there.
> 
> Oh, and this is the first time those debug lines are shown - there aren't any previous occurrences.
> 
> In another file, I do see the debug logging for the above calls. When the `KEY` is `1`, it works ok. When the Position index is `2`, it fails.

I think this issue is fixed in the changes I made post-0830, a combination of assumptions I made about how similar Wecdis was to the existing NMEA data.

importers/wecdis_importer.py
```diff
@@ -209,11 +253,15 @@ def handle_position(self, data_store, line_number, tokens, datafile, change_id):
         combine_tokens(source_token, sensor_token).record(self.name, "sensor", sensor_name)
 
         platform = self.get_cached_platform(data_store, self.platform_name, change_id=change_id)
-        sensor = self.get_cached_sensor(
-            data_store=data_store,
-            sensor_name=sensor_name,
-            sensor_type=sensor_token.text,
-            platform_id=platform.platform_id,
+        sensor_type = data_store.add_to_sensor_types("Location-Satellite", change_id=change_id).name
```
IanMayo, line 256: @robintw - we're processing a line representing ownship position here.  There are around  6 different ownship tracks in this format, each from a different sensor. Since the sensors are predictable, we don't ask the analyst to add/create them - we just add them.

So, `data_store.add_to_sensors` may get called for every position line.  (Finally...) here's my worry:  do we cache the sensor data?  Or do we  have the cost of checking if the sensor already exists in the database?

IanMayo (approved):
I suspect we're not exploiting all of the available data, but based upon our limited system documentation I'm comfortable we're doing the best we can.
This PR significantly improves the user experience for importing WECDIS data.

importers/wecdis_importer.py
robintw: We don't cache anything in `add_to_sensors`, no. That will do a db lookup for the information, and either return the existing entry or insert a new one if it doesn't already exist. Is there a reason you stopped using `get_cached_sensor`?

IanMayo: >  Is there a reason you stopped using get_cached_sensor?

Yes, to remove the user having to answer/acknowledge the prompts. There are ~6 ownship tracks for each file of this data-type. The analyst isn't adding any value by viewing/acknowledging each new sensor. So, I suggested to @mew-nsc  that we just create the sensor for the source, and move on - without any user interaction.

This _looked-like_ the correct way to handle that logic.  Is there another way?

Note: I compared old & new WECDIS importer (#rows/sec). The old one (with the requirement to acknowledge each new sensor) had about the same bulk import performance as the new one - though the new one avoided the need to answer the ownship track sensor answers.

robintw: Interesting that the speed was roughly the same. That may slow down with a remote Postgres server over a slow-ish network though.

I was about to tell you to use the `unknown=True` argument for the `get_cached_sensor` call, but that doesn't actually exist at the moment. That would cache the sensor, but also treat it as unknown and not ask the user any information about it. Probably best to create a separate issue to add that functionality to `get_cached_sensor` and then we can upgrade the importer to use that later.

PR #1113, opened 2021-12-14
1109 new jchat format

### LINKED ISSUE #1109: ANOTHER JChat format encountered
The team have a load of JChat data.  It's failing to load.

The issue is because the parents of the DTG have been swapped.

Instead of:
```
<tt>
  <font>[DDHHMMSSZ]</font>
</tt>
```

The files are:
```
<font>
  <tt>[DDHHMMSSZ]</tt>
</font>
```

We use `div.find` to find the element.  I suggest that if it fails to find a matching item we try it with the `tt` and `font` items swapped around.

I've changed the importer, to swap the items around.  When I import again, it reports failing to find the platform.  These fields have also been swapped.

Now the platform is inside `<font><b><a>CO RICH</a></b></font>`

While the structures are changing - I believe there is some commonality.  The DTG is always the lowest item in the first child of the div, and the platform is in the lowest item of the second div.

Note: I've introduced a data-file in this format [here](https://github.com/debrief/pepys-import/blob/develop/tests/sample_data/jchat_files/jchat_modern_2_format.html)

Once the parser is implemented, we should add some sample lines to the `combined_format.html` file.  I haven't done it in this PR, in case it breaks the existing tests.

mew-nsc:
## 🧰 Issue
Fixes #1109 

## 🚀 Overview: 
Added support for new JChat format using a slightly more generic approach that assumes that all JChat messages are always ordered as: timestamp - platform - message (which seems to be more reliably true than the order of the tags...)

## 🤔 Reason: 
To add support for another JChat format

## 🔨Work carried out:

- [x] Support added for new JChat format
- [x] More generic approach to support other combinations of tags 
- [x] Tests pass

## Confirmations

### DIFF
--- a/importers/jchat_importer.py
+++ b/importers/jchat_importer.py
@@ -122,20 +122,23 @@ class JChatImporter(Importer):
             # Older format uses id=
             # Newer format uses msgid=
             msg_id = div.attrib["msgid"]  # Grabbing ID to help with error reporting
-            version = JCHAT_MODERN
         except KeyError:
             try:
                 msg_id = div.attrib["id"]
-                version = JCHAT_LEGACY
             except KeyError:
                 # Ignore any non-comment messages (e.g. connect/disconnect)
                 return
 
-        time_element = div.find("{*}tt/font")
-
         # Sample data included some "Marker" messages with the id="marker"
         if str.upper(msg_id) == "MARKER":
             return  # Ignore these messages
+
+        text_blocks = []
+        text_blocks.append([item for item in div.findall(".//*") if item.text])
+        if text_blocks[0]:
+            time_element = text_blocks[0][0]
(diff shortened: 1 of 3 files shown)


IanMayo (approved):
Tested import of combined  jchat file using CodeSpace.  Verified import by inspection of SQLite file. 7 comments found, of `JChat` comment-type, with correct source.

mew-nsc:
@IanMayo The drop in code coverage is because the error message for an empty/unicode is no longer hit when the message is empty - this is caught by the earlier check for missing timestamp/platform/message. I've never been able to hit the unicode error, but I'm not certain that it can't be hit, so I'd prefer to leave that check in, but I've not found a way to test this - hence the drop in coverage.

PR #1129, opened 2022-02-07
Fix bug causing Python to completely crash when merging databases

### LINKED ISSUE #1121: Problem moving snapshot between databases
## Version

0.0.35

## 🐞 Description

Analysts are experimenting with how they would run Pepys across two disconnected networks.  For the trial, they did this:
* create two MS-Windows VMs
* install postgres on each VM
* load a track into VM1 using pepys-admin
* export a snapshot of all data for the year when the track happened
* move the snapshot to VM2
* try to load the snapshot into the VM2

The analyst (MH) reported that the import got to 18% before Pepys crashed, citing a Python error.

We won't be able to get a log of the output until the New Year.

We should try to reproduce the above.  Note: I guess we could use `tracstor` as the donor database (VM1), and a locally install of postgres as VM2.

robintw:
## 🧰 Issue
Fixes #1121 

## 🚀 Overview: 
Fixes an issue where Python crashed (completely - to the extent that Windows reported a problem with Python, or it just exited completely) when merging databases. After a lot of debugging (see comments on #1121) it seems this was caused by calls to `session.add_all()` and `session.commit()` being called a number of times in a loop with no work to do. This was a bug in our code, but also something that I wouldn't have expected to have caused a full crash - instead I'd have expected the calls to basically be no-ops.

This PR fixes this by putting these calls _inside_ the if statement that checks whether there is anything that needs changing, so they only get called if there is actually something to do.

I suspect the reason that this occurred on Windows and not MacOS or Linux is that Windows can be more picky about resource limits and access to resources in general, and it may have reached some limit that caused a crash. That is just my guess though.

I will attempt to put together a minimal example and report this bug to the SQLAlchemy developers.

## 🤔 Reason: 
Fix crashing bug.

## 🔨Work carried out:

- [x] Debug issue
- [x] Fix bug
- [x] Add test
- [x] Tests pass

## Confirmations

### DIFF
--- a/pepys_admin/merge.py
+++ b/pepys_admin/merge.py
@@ -361,9 +361,9 @@ class MergeDatabases:
                     for result in results:
                         result.entity = to_id
 
-                # Commit changes
-                self.slave_store.session.add_all(results)
-                self.slave_store.session.commit()
+                    # Commit changes
+                    self.slave_store.session.add_all(results)
+                    self.slave_store.session.commit()
 
     def update_logs_table(self, modified_ids):
         """Updates the Logs table in the slave_store for entries which have had their GUID modified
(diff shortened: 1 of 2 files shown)


IanMayo (approved):
Great, thanks @robintw

